Issue #79, opened 2018-11-19
For all `fmt*()` fcns, have option to include/exclude processing of NA rows


rich-iannone:
There should be better `NA`-handling options for all the `fmt*()` fcns. Currently, this example:

```r
sza %>%
  filter(latitude == 20) %>%
  filter(!is.na(sza)) %>%
  spread(key = "tst", value = sza) %>%
  gt(rowname_col = "month") %>%
  fmt_number(columns = TRUE, decimals = 2) %>%
  fmt_missing(columns = TRUE, missing_text = "")
```

will not replace the `NA` values because the `fmt_number()` will include all rows in the stored function (regardless of `NA` or not) and so `fmt_missing()` will have no rows to format.

PR #85, opened 2018-11-28
[WIP] Changes to `render_formats()` and improved NA-handling for formatters

rich-iannone:
This is an important infrastructural change that @schloerke kindly initiated during a meeting. It simplifies the rendering pipeline for the internal `data_df`, and makes `data_df` a list column `data.frame` during formatting. All of the `fmt_*()` functions have been modified to operate on list columns and better handle `NA`s and negative values (for the optional negative value text transformations).

The `fmt_missing()` function has been renamed to `fmt_na()`. And all relevant `fmt_*()` functions now have an `na_value` argument where if a value is supplied, `fmt_na()` will be called internally on the `data` first. This allows for formatting and NA-handling with a single function call.

Also, this fixes https://github.com/rstudio/gt/issues/79. So even if we use a formatter function and then `fmt_na()` on the same cells (or the reverse), we get the preferred behavior of `NA` handling occurring for the `NA` cells and cell formatting on the other cells.

```r
exibble %>%
  gt(
    rowname_col = "row",
    groupname_col = "group"
  ) %>%
  fmt_na(
    columns = vars(num),
    value = "missing"
  ) %>%
  fmt_number(
    columns = vars(num),
    decimals = 4
  )
```

pr_fmt_change

In the `num` column, we can now successfully format the column to 4 decimal places with `fmt_number()` and process missing values with `fmt_na()`. Previously, one function would override the other (with a non-ideal workaround of having to specify the exact rows for each formatter function to obtain the result shown here).

All tests and documentation have been modified, as has vignettes, etc. Tests and R CMD Check pass on dev and Travis. All manual tests with the .R/.Rmd scripts work as well.


rich-iannone:
Closing this PR as we have a better branch in the works.

PR #94, opened 2018-12-06
Fix for using of `fmt_scientific()` on columns with NA values

### LINKED ISSUE #80: `fmt_scientific()` errors when processing an NA cell
This will result in an error since the column contains `NA` values:

```r
exibble %>%
  gt(
    rowname_col = "row",
    groupname_col = "group"
  ) %>%
  fmt_scientific(
    columns = vars(num),
    rows = c(5, 7, 8)
  )
```

rich-iannone:
Previously, this code...

```r
dplyr::tibble(x = c(-15.0, -1.5, -0.001, NA, 0, 0.001, 1.5, 22.5, 333.5)) %>%
  gt() %>%
  fmt_scientific(columns = vars(x))
```

...would result in this error:

```
 Error in FUN(X[[i]], ...) : subscript out of bounds 
```

The error occurred because there was no treatment of `NA`s as with the other `fmt*()` functions. With the changes here, we don't get an error and gt table output is correct:

after

A test was added to check the `fmt_scientific()` function for resilience to `NA` values.

This fixes #80.

### DIFF
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -143,8 +143,11 @@ fmt_number <- function(data,
           # Determine which of `x` are not NA
           non_na_x <- !is.na(x)
 
+          # Create `x_str` with same length as `x`
+          x_str <- rep(NA_character_, length(x))
+
           # Format all non-NA x values
-          x[non_na_x] <-
+          x_str[non_na_x] <-
             formatC(
               x = x[non_na_x] * scale_by,
               digits = decimals,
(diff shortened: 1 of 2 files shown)


jcheng5:
Don't mutate x, the previous code had the right idea of making a separate x_str. This is part of what I wanted to talk to you about with the fmt functions in general. x should always be of the raw data type. Make a separate variable for a char vector of the same length, and return it.

rich-iannone:
Totally clear on this now. Just refactored to not write anything back into `x`. Instead using `x_str`, which is returned at the end.

rich-iannone:
@jcheng5 I've refactored all `fmt_*()` functions to have the same pattern of using `x_str` to hold the formatted character data.

PR #95, opened 2018-12-06
Improvement to the handling of a `pattern` in most `fmt*()` functions

### LINKED ISSUE #61: Modify the transformation of formatted values using the `pattern` argument
The `pattern` argument in all of the `fmt_*()`, assumes that pattern only contains one copy of `{x}` (the formatted output for each cell undergoing formatting).

Because of this, we can't use a useful pattern like `<a href="{x}">{x}</a>`.

@jcheng5 's idea is to use `gsub("{x}", x, pattern, fixed = TRUE)` instead.

rich-iannone:
This adds a utility function to replace multiple instances of `{x}` in the vector of formatted character values. This requires https://github.com/rstudio/gt/pull/94 to be merged first, then the `fmt_scientific()` function will be modified with this change.

All current tests pass with the change. More tests should be added to verify that the additional functionality works.

This fixes #61.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -456,6 +456,60 @@ non_na_index <- function(values, index, default_value = NA) {
   positions[index]
 }
 
+non_na_index <- function(values, index, default_value = NA) {
+
+  if (is.logical(index)) {
+    index <- is.integer(index)
+  }
+
+  stopifnot(is.integer(index) || is.numeric(index))
+
+  # The algorithm requires `-Inf` not being present
+  stopifnot(!any(is.infinite(values) & values < 0))
+
+  # Get a vector of suffixes, which may include
+  # NA values
+  res <- values[index]
+
+  # If there are no NA values to reconcile, return
+  # the index
+  if (!any(is.na(res))) {
+    return(index)
+  }
+
+  # Create a vector of positions (`seq_along(values)`),
+  # but anywhere the `values` vector has an NA, use
+  # `-Inf`; (it's important that `values` not have `-Inf`
+  # as one of its elements)
+  positions <- ifelse(!is.na(values), seq_along(values), -Inf)
+
+  # Use rle (run-length encoding) to collapse multiple
+  # instances of `-Inf` into single instances. This
+  # makes it easy for us to replace them with their
+  # nearest (lower) neighbor in a single step, instead of
+  # having to iterate; for some reason, `rle()` doesn't
+  # know how to encode NAs, so that's why we use -Inf
+  # (seems like a bug)
+  encoded <- rle(positions)
+
+  # Replace each -Inf with its closest neighbor; basically,
+  # we do this by shifting a copy of the values to the
+  # right, and then using the original vector of (run-length
+  # encoded) values as a mask over it
+  encoded$values <-
+    ifelse(
+      encoded$values == -Inf,
+      c(default_value, head(encoded$values, -1)),
+      encoded$values
+    )
+
+  # Now convert back from run-length encoded
+  positions <- inverse.rle(encoded)
+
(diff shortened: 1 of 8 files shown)


jcheng5:
How'd this end up with so many files changed? Did you accidentally delete the man folder or something?

R/utils.R
```diff
@@ -372,16 +372,74 @@ markdown_to_text <- function(text) {
...
+#' @param values The values (as a character vector) that are formatting within
+#'   the \code{fmt_*()} function.
+#' @importFrom glue glue
+#' @noRd
+apply_pattern_fmt_x <- function(pattern,
+                                values) {
 
-  prefix <- strsplit(pattern, "\\{x\\}")[[1]][1]
-  suffix <- strsplit(pattern, "\\{x\\}")[[1]][2]
+  glue::glue(pattern, x = values) %>% as.character()
```
jcheng5: Do you really want glue here? It will mean that patterns can be used to execute arbitrary R code. It'd be more conservative to just look for, and replace, `{x}`.

jcheng5: Hmmm, WTF, glue? The presence of a newline can affect whitespace trimming behavior??

```
> glue::glue("   {'1'}   ") %>% as.character()
[1] "   1   "
> glue::glue("   {'1'}\n   ") %>% as.character()
[1] "1"
```

R/utils.R
```diff
@@ -372,16 +372,74 @@ markdown_to_text <- function(text) {
...
+}
 
-  prefix <- ifelse(is.na(prefix), "", prefix)
-  suffix <- ifelse(is.na(suffix), "", suffix)
+# This function normalizes the `suffixing` input to a
+# character vector which is later appended to scaled
+# numerical values; the input can either be a single
+# logical value or a character vector
+normalize_suffixing_inputs <- function(suffixing) {
```
jcheng5: What's happening here, did this just move from a different spot?

rich-iannone:
The hundreds of files changed is mostly due to this being an old branch and so the pkgdown site `docs` are out of sync with the current ones.

rich-iannone:
...the pkgdown HTML tables have a random `id`, meaning regenerations of the site will change most of the files. Perhaps there should be an option to disable the random `id` in the enclosing `<div>` as mentioned in https://github.com/rstudio/gt/issues/141.

R/utils.R
rich-iannone: I think this is a case (aside from the other point) for not using glue.

rich-iannone: This is something I wasn't entirely sure about and so I wanted your opinion on the safety aspect. Using glue seems very powerful because of allowing execution within `{ }` but that's also the disadvantage.

I'm thinking now that the original, non-glue plan might be better.

R/utils.R
rich-iannone: It's now moved to line 400 (compared to line 546 in master). I carefully checked to see if this is different from master and found it's identical.

R/utils.R
```diff
@@ -390,7 +390,12 @@ markdown_to_text <- function(text) {
 apply_pattern_fmt_x <- function(pattern,
                                 values) {
 
-  glue::glue(pattern, x = values) %>% as.character()
+  vapply(
+    values,
+    function(x) tidy_gsub(x = pattern, "\\{x\\}", x),
+    FUN.VALUE = character(1),
+    USE.NAMES = FALSE
+  )
```
jcheng5: gsub is vectorized, there's no need for vapply here.

jcheng5:
Other than removing that vapply, looks good. Can you make sure this squashes down to a nice commit or set of commits that doesn't have all the spurious changes to docs? (I can help you with that if necessary)

Issue #96, opened 2018-12-06
Support for PDF output


kvittingseerup:
I would like to requires the option of outputting tables in PDF format. That would be really useful for anything that needs to be put into print (where vector graphics is needed)

rich-iannone:
Excellent idea! Will try to get this going in the coming weeks. I'm also thinking that there should be a broad range of output graphic formats (both raster and vector-based).

baptiste:
FWIW gridExtra has a `grid.table()` function, which can be displayed using standard graphics devices. It's quite limited in terms of formatting; the only benefit that I can see would be in mixing content on a graphic, e.g. a table as an inset in a plot, etc. 
I haven't looked at the code for `gt` but in theory it would be possible to translate its table description into several sub-tables, each rendered as an individual tableGrob, and combine them together.

https://cran.r-project.org/web/packages/gridExtra/vignettes/tableGrob.html

PR #103, opened 2018-12-12
[WIP] Add PDF and raster image exporting functions

rich-iannone:
By interfacing with the `wkhtmltopdf` and `wkhtmltoimage` binaries (pre-compiled, all platforms, and easy-to-install) we can create PDF and raster image files that are replicas of what is visible in the RStudio Viewer.

Here is the example usage for a PNG file:

```r
# Use `countrypops` to create a gt table;
# export the table as a PNG file
countrypops %>%
  dplyr::select(-contains("code")) %>%
  dplyr::filter(country_name == "Mongolia") %>%
  tail(5) %>%
  gt() %>%
  cols_align(
    align = "left",
    columns = vars(population)
  ) %>%
  gt_png(file_name = "countrypops.png")
```

The output file looks like this:

countrypops

where the image whitespace on the left and right is trimmed by default (this is an option). 

Addresses #96 and partially solves the issue in https://github.com/rstudio/gt/issues/97.


rich-iannone:
@pasipasi123 Yes, this is a problem and I'm glad you pointed it out. A solution might be to add a standard page `width` that can be configurable.

Another thing we are mulling over is whether to use `wkhtmltopdf`/`wkhtmltoimage`, the **webshot** R package, or headless Chrome/Chromium to create these images. I'm currently testing these and seeing what produces the best results with as little dependency pain as possible.

rich-iannone:
As we are seeing that a better solution (headless Chrome/Chromium) is being developed for getting screenshots and PDFs from HTML, this PR will be closed.

Issue #113, opened 2018-12-17
table title doesn't span all columns when exported to html


yiqinfu:
Using the example from the documentation, we save the table to an html file:

```
start_date <- "2010-06-07"
end_date <- "2010-06-14"

plot = sp500 %>%
  dplyr::filter(date >= start_date & date <= end_date) %>%
  dplyr::select(-adj_close) %>%
  dplyr::mutate(date = as.character(date)) %>%
  gt() %>%
  tab_header(
    title = "S&P 500",
    subtitle = glue::glue("{start_date} to {end_date}")
  ) %>%
  tab_style(cells_styles(text_font = "Optima",
                         text_align = "center",
                         text_weight = "bolder"),
            locations = cells_title(groups = title)
  ) %>%
  fmt_date(
    columns = vars(date),
    date_style = 3
  ) %>%
  fmt_currency(
    columns = vars(open, high, low, close),
    currency = "USD"
  ) %>%
  fmt_number(
    columns = vars(volume),
    scale_by = 1 / 1E9,
    pattern = "{x}B"
  ) %>%
  as_raw_html(inline_css = T)
write_file(plot, "plot.html")
```

If we open the html file, we see this:

![screen shot 2018-12-16 at 9 43 55 pm](https://user-images.githubusercontent.com/23423360/50063735-bde4bb80-017b-11e9-9bf2-4b6a962b453a.png)

The title doesn't span all columns.

Edit: The title *does* span all columns when no style is added. If we add any style to it, it doesn't display properly.

rich-iannone:
I see the issue. Internally, the `gt:::inline_html_styles()` function will see CSS `class` and `style` attributes and, during inlining, merge generated style rules into the existing set. The problem is that the `colspan` attr (included in both `<th>`'s of the title and subtitle) is wiped out because of its positioning between `class` and `style` (which is added to the end of the end of the tag). The solution is to put the `colspan` attribute before `class`.

yiqinfu:
> I see the issue. Internally, the `gt:::inline_html_styles()` function will see CSS `class` and `style` attributes and, during inlining, merge generated style rules into the existing set. The problem is that the `colspan` attr (included in both `<th>`'s of the title and subtitle) is wiped out because of its positioning between `class` and `style` (which is added to the end of the end of the tag). The solution is to put the `colspan` attribute before `class`.

Thank you! Looking forward to the update.

rich-iannone:
@yiqinfu You're welcome! Thanks for filing the issue.

PR #117, opened 2018-12-20
Title/subtitle fix for inlining with existing custom styles

rich-iannone:
When a table has a custom style applied to either the title or the subtitle (through `tab_style()`), and, the table is exported to an HTML with inlined CSS styles, the title or subtitle won't span across the width of the table. The following code will produce a correct table in the Viewer:

```r
gt(mtcars[1:2, ]) %>%
  tab_header(
    title = "The title",
    subtitle = "The subtitle"
  ) %>%
  tab_style(
    cells_styles(
      text_align = "left"),
    locations = list(
      cells_title(groups = "title"),
      cells_title(groups = "subtitle")
    )
  )
```

table_in_viewer

When it is exported to HTML via `as_raw_html()` and saved to disk, the viewed HTML won't have the `colspan` attr in the title or in the subtitle.

table_in_viewer_before_fix

This PR fixes that erroneous result by preserving the `colspan` attr. A testthat test has also been added. 

Fixes https://github.com/rstudio/gt/issues/113

### DIFF
--- a/R/utils_render_html.R
+++ b/R/utils_render_html.R
@@ -294,9 +294,14 @@ create_heading_component <- function(heading,
     heading_component <-
       paste0(
         "<thead>\n<tr>\n",
-        "<th class='gt_heading gt_title gt_font_normal gt_center' colspan='",
-        n_cols, "'", create_style_attrs(title_style_attrs), ">",
-        heading$title, footnote_title_glyphs, "</th>\n</tr>\n")
+        "<th ",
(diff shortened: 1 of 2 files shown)


jcheng5:
After in-person discussion, I think `inline_html_styles` should at least be rewritten to use capturing groups so that any valid combination of class/style will be respected.

jcheng5:
Alternatively, we could rewrite with xml2.

PR #123, opened 2019-01-04
Add function to get LaTeX dependencies for external use

rich-iannone:
An exported function is added here (`get_latex_dependencies()`) to make it easier to declare which packages are required for **gt** when using Rnw. A minimal example:

````
%!sweave=knitr

\documentclass{article}

<<results='asis', echo=FALSE>>=
gt::get_latex_dependencies()
@

\begin{document}

Content...

\end{document}
````  

Using the `get_latex_dependencies()` function in this way places the LaTeX `\use_package{}` statements before the document start, which is convenient when using a **gt** table within the document. This change also centralizes the names of the LaTeX packages that are used in the LaTeX-rendering pipeline and in the aforementioned function.


jcheng5:
Let's rename `get_latex_dependencies` to `gt_latex_dependencies`

PR #130, opened 2019-01-09
Add the `gtsave()` function

rich-iannone:
The `gtsave()` function is analogous to `ggsave()` in **ggplot2** but can be added at the end of a pipeline (since it has the `data` argument for the gt table object). It will use the extension of the provided output filename to use the correct saving method (.html/.htm, .tex/.ltx, or .rtf). The function also accepts a user-defined function to allow for customized saving.


R/gtsave.R
```diff
@@ -0,0 +1,67 @@
...
+
+  stop("`gtsave.", gtsave_file_ext(filename),
+       "(data, filename, ...)` has not been implemented.",
+       call. = FALSE)
+}
+
+gtsave.html <- gtsave.htm <- function(data, filename, ..., inline_css = TRUE) {
+
+  data %>%
+    as_raw_html(inline_css = inline_css) %>%
```
jcheng5: Consider using `data %>% htmltools::as.tags() %>% htmltools::save_html(filename)`

rich-iannone: Thanks! Will implement that exactly.

tests/testthat/test-gtsave.R
```diff
@@ -0,0 +1,122 @@
...
+
+  # Expect that a file does not yet exist
+  # on that path
+  expect_false(file.exists(path_2))
+
+  # Save a gt table based on the exibble dataset
+  # as an HTML file to the `path_2`
+  exibble %>%
+    gt() %>%
+    gtsave(filename = path_2)
```
jcheng5: Add a test case for inlineCSS=...

PR #131, opened 2019-01-11
Fix rtf table generation example

elong0527:
1. Fix bug to Make the example work in as_rtf function.
2. Allow user to customize components in RTF table.


rich-iannone:
RTF table generation was rewritten for the current CRAN release so I’ll close this.

Issue #127, opened 2019-01-07
tab_spanner() does nothing if columns start with a number


sharlagelfand:
It looks like tab_spanner() is not able to handle columns that start with numbers! (not ideal in general, I know, but I have them)

the gt_tbl output is nonsense HTML so I've excluded it -- but this was created using reprex :)

``` r
library(dplyr)
#> 
#> Attaching package: 'dplyr'
#> The following objects are masked from 'package:stats':
#> 
#>     filter, lag
#> The following objects are masked from 'package:base':
#> 
#>     intersect, setdiff, setequal, union
library(gt)

df <- tibble(`1_x` = c(1, 2),
             `2_x` = c(1, 2), 
             x_1 = c(3, 4),
             x_2 = c(3, 4))

# Tab spanner works if variables start with a character
df %>%
  gt() %>%
  tab_spanner(label = "x",
              columns = vars(x_1, x_2))

# select_at() works if variables start with a number

df %>%
  select_at(vars(`1_x`, `2_x`))

#> # A tibble: 2 x 2
#>   `1_x` `2_x`
#>   <dbl> <dbl>
#> 1     1     1
#> 2     2     2

# Nothing happens when using tab spanner on variables starting with a number

df %>%
  gt() %>%
  tab_spanner(label = "x",
              columns = vars(`1_x`, `2_x`))
```

rich-iannone:
Thanks for reporting this! It's interesting that the `select_at()` example works fine with variables starting with a number (and that the gt analogue does not). The last `tab_spanner()` does work with some unintuitive changes:

```
# this works!
df %>%
  gt() %>%
  tab_spanner(
    label = "x",
    columns = vars("1_x", "2_x")
  )

# this too!
df %>%
  gt() %>%
  tab_spanner(
    label = "x",
    columns = c("1_x", "2_x")
  )
```

But it's pretty odd that it doesn't work like dplyr's vars with the back ticks which is the standard way to do it. I'll look into a fix for this. But, for now, at least there is a workaround (that should work even when the fix is implemented).

Issue #110, opened 2018-12-15
Add option to gather columns when using `tab_spanner()`


rich-iannone:
Currently, `tab_spanner()` will not modify the positions of the table columns when non-contiguous columns are included in the call. This is acceptable when you want a repeating spanner column label across the top of the column labels. However, it may be convenient to have the option to automatically gather the columns that are part of a single spanner column label.

Here's how you might move a column to join a group of columns with `cols_move()` and use `tab_spanner()`:

```r
gtcars %>%
  dplyr::select(
    -mfr, -trim, bdy_style, drivetrain,
    -drivetrain, -trsmn, -ctry_origin
  ) %>%
  dplyr::slice(1:8) %>%
  gt(rowname_col = "model") %>%
  tab_header(title = "gtcars") %>%
  tab_spanner(
    label = "performance",
    columns = vars(
      hp, hp_rpm, trq, trq_rpm,
      mpg_c, mpg_h)
  ) %>%
  cols_move(vars(msrp), after = vars(bdy_style)) %>%
  tab_spanner(
    label = "model info",
    columns = vars(
      year, bdy_style, msrp)
  )
```

The `cols_move()` call could be made unnecessary by implementing a `gather` argument. So the code would be simplified to this:

```r
gtcars %>%
  dplyr::select(
    -mfr, -trim, bdy_style, drivetrain,
    -drivetrain, -trsmn, -ctry_origin
  ) %>%
  dplyr::slice(1:8) %>%
  gt(rowname_col = "model") %>%
  tab_header(title = "gtcars") %>%
  tab_spanner(
    label = "performance",
    columns = vars(
      hp, hp_rpm, trq, trq_rpm,
      mpg_c, mpg_h)
  ) %>%
  tab_spanner(
    label = "model info",
    columns = vars(
      year, bdy_style, msrp),
    gather = TRUE
  )
```

Which yields this table:

<img width="835" alt="tab_spanner_issue" src="https://user-images.githubusercontent.com/5612024/50039207-02e1e400-fffc-11e8-8d91-dc27ef6347ce.png">

Issue originally discovered and suggested by @tareefk .

Issue #102, opened 2018-12-11
Implement `starts_with` column selectors in tab_spanner


MikeKSmith:
It would be good to allow the ability to use the `starts_with` column selectors in the tab_spanner `columns` definition:
   
``` r
library(tidyverse)
library(gt)
myTableData <- tribble(~Variable, ~Value, ~Female_n, ~Female_prop, ~Male_n, ~Male_prop,
"Compliance", "No",            2,      0.0270,      0,     0,    
"Compliance", "Yes",          13,      0.176,      59,     0.797)

# Explicit definition of table spanning columns
myTableData %>%
group_by(Variable) %>%
  gt() %>%
  tab_header(title = "Summary of categorical covariates") %>%
  tab_spanner(label="Female",
              columns=vars("Female_n", "Female_prop")) %>%
  tab_spanner(label="Male",
              columns=vars("Male_n", "Male_prop")) %>%
  cols_label(Female_n = "n",
             Male_n = "n",
             Female_prop = "Percentage",
             Male_prop = "Percentage")
```

<style>html {
  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Fira Sans', 'Droid Sans', 'Helvetica Neue', Arial, sans-serif;
}

#ngwpwvkoag .gt_table {
  display: table;
  border-collapse: collapse;
  margin-left: auto;
  margin-right: auto;
  color: #000000;
  font-size: 16px;
  background-color: #FFFFFF;
  /* table.background.color */
  width: auto;
  /* table.width */
  border-top-style: solid;
  /* table.border.top.style */
  border-top-width: 2px;
  /* table.border.top.width */
  border-top-color: #A8A8A8;
  /* table.border.top.color */
}

#ngwpwvkoag .gt_heading {
  background-color: #FFFFFF;
  /* heading.background.color */
  border-bottom-color: #FFFFFF;
}

#ngwpwvkoag .gt_title {
  color: #000000;
  font-size: 125%;
  /* heading.title.font.size */
  padding-top: 4px;
  /* heading.top.padding */
  padding-bottom: 1px;
  border-bottom-color: #FFFFFF;
  border-bottom-width: 0;
}

#ngwpwvkoag .gt_subtitle {
  color: #000000;
  font-size: 85%;
  /* heading.subtitle.font.size */
  padding-top: 1px;
  padding-bottom: 4px;
  /* heading.bottom.padding */
[...]

PR #136, opened 2019-01-16
Add option to gather columns in `tab_spanner()`

rich-iannone:
Currently, `tab_spanner()` will not move columns to create a contiguous block of columns under a spanner heading. While this is acceptable when you want a repeating spanner column label across the top of the column labels, it's likely not the most common use case. 

This change adds a `gather` argument that's set to `TRUE`. This makes it convenient for automatically gathering the columns that are part of a single spanner column label. Before the change, the user would have to carefully ensure the columns are contiguous before entry in `gt()`, or, would need to use some of the `cols_move*()` functions for manual movement of columns.

Here is an example where columns are automatically moved under the spanner column label

```r
gtcars %>%
  dplyr::select(
    -mfr, -trim, bdy_style, drivetrain,
    -drivetrain, -trsmn, -ctry_origin
  ) %>%
  dplyr::slice(1:8) %>%
  gt(rowname_col = "model") %>%
  tab_header(title = "gtcars") %>%
  tab_spanner(
    label = "performance",
    columns = vars(
      hp, hp_rpm, trq, trq_rpm,
      mpg_c, mpg_h)
  ) %>%
  tab_spanner(
    label = "model info",
    columns = vars(
      bdy_style, year, msrp)
  )
```

and this yields the following **gt** output table:

tab_spanner_arrange

The key thing to note here is that the `msrp` column in the second `tab_spanner()` call is separated from the `bdy_style` and `year` set of columns. The function now recognizes these gaps and moves columns together whilst preserving the ordering of columns in the group. 

Fixes https://github.com/rstudio/gt/issues/110.
Fixes https://github.com/rstudio/gt/issues/127.
Fixes https://github.com/rstudio/gt/issues/102.

### DIFF
--- a/R/table_parts.R
+++ b/R/table_parts.R
@@ -225,6 +225,9 @@ tab_row_group <- function(data,
 #' @inheritParams fmt_number
 #' @param label the text to use for the spanner column label.
 #' @param columns the columns to be components of the spanner heading.
+#' @param gather an option to move the specified \code{columns} such that they
+#'   are unified under the spanner column label. Ordering of the
+#'   moved-into-place columns will be preserved in all cases.
 #' @return an object of class \code{gt_tbl}.
 #' @examples
 #' # Use `gtcars` to create a gt table;
--- a/R/table_parts.R
+++ b/R/table_parts.R
@@ -280,6 +280,54 @@ tab_spanner <- function(data,
   # Set the `grp_labels` attr with the `grp_labels` object
   attr(data, "grp_labels") <- grp_labels
 
+  # Gather columns not part of the group of columns under
+  # the spanner heading
+  if (gather) {
+
+    # Extract the internal `boxh_df` table
+    boxh_df <- attr(data, "boxh_df", exact = TRUE)
+
+    # Get the sequence of columns available in `boxh_df`
+    all_columns <- colnames(boxh_df)
+
+    # Get the vector positions of the `columns` in
+    # `all_columns`
+    matching_vec <- match(columns, all_columns) %>% sort()
+
+    # Reassign `columns` as a sorted vector of columns
+    columns <- all_columns[matching_vec]
+
+    # Gather columns if there are gaps between index positions
+    if (length(which(diff(matching_vec) != 1)) != 0) {
+
+      # Determine the name of the column in `columns` that is
+      # in the rightmost position in the main group
+      last_column <-
+        columns[which(diff(matching_vec) != 1)][1]
+
+      # Determine which column in `columns` is the first to
+      # be separated from the main group
+      separated_start <- (which(diff(matching_vec) != 1) + 1)[1]
+
+      for (i in separated_start:length(seq(columns))) {
+
+        # Perform a single-column move (separated column
+        # moving to the right of the right-most column
+        # in the main group)
+        data <-
+          data %>%
+          cols_move(
+            columns = columns[i],
+            after = last_column
+          )
+
+        # Reassign the `last_column` as the column moved
+        # into the main group
+        last_column <- columns[i]
+      }
+    }
+  }
+
   data
 }
 
--- a/R/table_parts.R
+++ b/R/table_parts.R
@@ -253,7 +253,8 @@ tab_row_group <- function(data,
 #' @export
 tab_spanner <- function(data,
                         label,
-                        columns) {
+                        columns,
+                        gather = TRUE) {
   checkmate::assert_character(
     label, len = 1, any.missing = FALSE, null.ok = FALSE)
 
--- a/R/table_parts.R
+++ b/R/table_parts.R
@@ -301,7 +301,17 @@ tab_spanner <- function(data,
 
     # Get the vector positions of the `columns` in
     # `all_columns`
-    matching_vec <- match(resolved_columns, all_columns) %>% sort()
+    matching_vec <-
+      match(resolved_columns, all_columns) %>%
+      sort() %>%
+      unique()
+
+    # Issue a warning if duplicate values were given
+    # in resolved columns
+    if (any(duplicated(resolved_columns))) {
+      warning("Duplicate column names should not be provided in `columns`.",
+              call. = FALSE)
+    }
 
     # Get a vector of column names
     columns_sorted <- all_columns[matching_vec]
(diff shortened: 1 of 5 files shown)


R/table_parts.R
```diff
@@ -279,6 +283,54 @@ tab_spanner <- function(data,
...
+
+    # Get the vector positions of the `columns` in
+    # `all_columns`
+    matching_vec <- match(columns, all_columns) %>% sort()
+
+    # Reassign `columns` as a sorted vector of columns
+    columns <- all_columns[matching_vec]
+
+    # Gather columns if there are gaps between index positions
+    if (length(which(diff(matching_vec) != 1)) != 0) {
```
jcheng5: `any(diff(matching_vec) != 1)`

R/table_parts.R
```diff
@@ -279,6 +283,54 @@ tab_spanner <- function(data,
...
+
+      # Determine the name of the column in `columns` that is
+      # in the rightmost position in the main group
+      last_column <-
+        columns[which(diff(matching_vec) != 1)][1]
```
jcheng5: You could drop the `which`

R/table_parts.R
```diff
@@ -279,6 +283,54 @@ tab_spanner <- function(data,
...
+          data %>%
+          cols_move(
+            columns = columns[i],
+            after = last_column
+          )
+
+        # Reassign the `last_column` as the column moved
+        # into the main group
+        last_column <- columns[i]
+      }
```
jcheng5: 1. Please extract out the column rearranging logic into a separate function that is easily unit testable. Could operate just on column names, not the whole data frame, if you want.
2. I'd like to see this done without a for-loop.

R/table_parts.R
```diff
@@ -279,6 +283,54 @@ tab_spanner <- function(data,
...
+
+    # Extract the internal `boxh_df` table
+    boxh_df <- attr(data, "boxh_df", exact = TRUE)
+
+    # Get the sequence of columns available in `boxh_df`
+    all_columns <- colnames(boxh_df)
+
+    # Get the vector positions of the `columns` in
+    # `all_columns`
+    matching_vec <- match(columns, all_columns) %>% sort()
```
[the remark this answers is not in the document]
rich-iannone: This has now been added (both the `unique()` and the `warning()`).

R/table_parts.R
rich-iannone: This entire part has been removed (for a much simpler solution).

R/table_parts.R
rich-iannone: This entire part has been removed (for a much simpler solution).

R/table_parts.R
rich-iannone: I've taken the ~15-20 lines of code and simply replaced with a single `cols_move()` statement.

rich-iannone:
In addition to the recommended fixes, the `tab_spanner()` function can now use the **tidyselect** select helpers. This fixes https://github.com/rstudio/gt/issues/102.

R/table_parts.R
```diff
@@ -253,32 +256,75 @@ tab_row_group <- function(data,
...
+  resolved_columns <-
+    resolve_vars(var_expr = columns, var_names = colnames, data_df = data_df)
+
+  # Translate the column indices to column names
+  resolved_columns <- colnames[resolved_columns]
 
   # Filter the vector of column names by the
   # column names actually in `input_df`
-  columns <- columns[which(columns %in% colnames(data))]
+  resolved_columns <- resolved_columns[which(resolved_columns %in% colnames)]
```
jcheng5: I think you can drop this check now since resolved_columns errors on unknown columns.

R/table_parts.R
```diff
@@ -253,32 +256,75 @@ tab_row_group <- function(data,
...
+  if (length(resolved_columns) == 0) {
     return(data)
   }
 
   grp_labels <- attr(data, "grp_labels", exact = TRUE)
 
-  for (i in seq(columns)) {
-    grp_labels[[columns[i]]] <- label
+  for (i in seq(resolved_columns)) {
+    grp_labels[[resolved_columns[i]]] <- label
```
jcheng5: Please vectorize this for loop away (not apply)

R/table_parts.R
```diff
@@ -253,32 +256,75 @@ tab_row_group <- function(data,
...
+    # Get the vector positions of the `columns` in
+    # `all_columns`
+    matching_vec <-
+      match(resolved_columns, all_columns) %>%
+      sort() %>%
+      unique()
+
+    # Issue a warning if duplicate values were given
+    # in resolved columns
+    if (any(duplicated(resolved_columns))) {
```
jcheng5: Seems like this check should be outside the conditional?

R/table_parts.R
```diff
@@ -253,32 +256,75 @@ tab_row_group <- function(data,
...
+    # Get a vector of column names
+    columns_sorted <- all_columns[matching_vec]
+
+    # Move columns into place
+    data <-
+      data %>%
+      cols_move(
+        columns = columns_sorted[-1],
+        after = columns_sorted[1]
+      )
```
jcheng5: Much nicer!

R/table_parts.R
```diff
@@ -253,32 +256,75 @@ tab_row_group <- function(data,
...
-  if (inherits(columns, "quosures")) {
-    columns <- columns %>% lapply(`[[`, 2) %>% as.character()
-  }
+  data_df <- as.data.frame(data)
+  colnames <- colnames(data_df)
+
+  columns <- enquo(columns)
+
+  resolved_columns <-
+    resolve_vars(var_expr = columns, var_names = colnames, data_df = data_df)
```
jcheng5: I'm a little surprised this isn't `var_expr = !!columns`, are you pretty sure about that?

jcheng5: Scratch that sorry

R/table_parts.R
rich-iannone: Thanks :)

R/table_parts.R
rich-iannone: Done.

R/table_parts.R
rich-iannone: `FACE PALM` because: `grp_labels[resolved_columns] <- label` !

R/table_parts.R
rich-iannone: Yes. But, also, duplicates won't happen anymore because of the use of `resolve_vars()`. So I just took it out (after sanity-checking this thoroughly).

rich-iannone:
I put in a TODO about upcoming changes in the later PR that will simplify resolving. This will be the next PR submitted and the next that should be reviewed.

PR #137, opened 2019-01-17
Bugfix: fmt_scientific() didn't show exp value for small numbers.

gaospecial:
Here is an example:

```{r}
library("gt")
data <- data.frame(sci=c(0.000137036,3.23166e-94,0,273669))
data %>% gt() %>% fmt_scientific(columns=vars("sci"),decimals = 2)
```

Before fixing, it showed as 3.23 for 3.23e-94.

image

### DIFF
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -264,7 +264,7 @@ fmt_scientific <- function(data,
       small_pos <-
         ((x >= 1 & x < 10) |
            (x <= -1 & x > -10) |
-           is_equal_to(x, 0)) & !is.na(x)
+           is_equal_to(as.numeric(split_scientific_notn(formatC(x,format = "e"))$num),0)) & !is.na(x)
 
       # Create `x_str` with same length as `x`
       x_str <- rep(NA_character_, length(x))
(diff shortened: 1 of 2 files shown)


rich-iannone:
@gaospecial thanks for making a PR for this! I tested out your fix and found that while it worked for the small example, some testthat tests failed as well. The fix you proposed is meant to deal with very small numbers retaining their scientific notation. The reason that the notation disappears is because we check whether the incoming number is zero (we don't want to show `0 x 10^0`). However, the test for whether the value is zero used the following utility function (in `utils.R`):

```r
is_equal_to <- function(x, y) {
  vapply(x, function(a, b) isTRUE(all.equal(a, b)), logical(1), y)
}
```

The use of `all.equal()` is problematic since it's not exact enough for our purpose of comparing to zero. This should be changed to:

```r
is_equal_to <- function(x, y) {
  vapply(x, function(a, b) isTRUE(identical(a, b)), logical(1), y)
}
```

I'd like to give you credit for finding the problem and proposing a fix, so, could you make this exact change and push the commit? (Also, revert the change in `fmt_scientific()`.)

Finally, please sign the [individual](https://rstudioblog.files.wordpress.com/2017/05/rstudio_individual_contributor_agreement.pdf) or [corporate](https://rstudioblog.files.wordpress.com/2017/05/rstudio_corporate_contributor_agreement.pdf) contributor agreement as appropriate. Then, send the signed copy over to jj@rstudio.com. We need this for each person submitting a PR.

Thanks!

Issue #144, opened 2019-01-22
Feature request: "fmt_accounting()"


steveputman:
This was referred to in passing in #1, but it would be wonderful to be able to format negative numbers with parentheses. The `formattable` package had this (though I note that there are some annoyances with alignment, more so in pdf than in html).

Thanks again for the great package!

rich-iannone:
You’re in luck because we already have this functionality. I think it’s probably badly named though: the argument `negative_val` can be set to `parens` in `fmt_number()` and some other functions as well.

steveputman:
Ha! Awesome--sorry I didn't failed to track that down on my own (read *all* the code, Steve).

steveputman:
I made some changes on my fork to deal with the alignment issues when using parentheses (typically you want the last digit of the positive number to line up with the last digit of the negative number rather than the parenthesis). The modification appends `hphantom{(}` in LaTeX or a hidden `<span>` with a parenthesis in HTML to all non-negative numbers to make the alignment work in `fmt_number()`, `fmt_percent()`, and `fmt_currency()`. 

Not the prettiest solution, but it seems to work. Happy to prepare PR if interested.

rich-iannone:
@steveputman I'm looking for alignment solutions in general, so this is great! I'll reopen this issue (where you could show the API usage and outputs) and please do start a PR (and we can discuss the implementation more in there). Thanks!

steveputman:
Great. The idea was to honor whatever was passed in `cols_align`, so no API changes (unless for some reason, one needed the option of aligning the paren with the digit). 

With this code in an R Markdown file (varying the output parameter):
````
---
title: "parens_alignment"
output: html_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)

library(tidyverse)
library(gt)

df <- tribble(
  ~chars, ~dates, ~curr, ~num, ~pct,
  "rowA", "2018-01-01", 1000.25, 0, .353,
  "rowB", "2018-01-02", -24.37, 10000.823, 0,
  "rowC", "2018-01-03", 0, -.7209, 0.234,
  "rowD", "2018-01-04", -24385.1, 288.21, -1.399
  ) 

rightalign <- df %>%
  gt() %>%
  tab_header(
    title = "Default alignment",
    subtitle = "-"
  ) %>%
  fmt_currency(
    columns = vars(curr),
    negative_val = "parens"
  ) %>%
  fmt_number(
    columns = vars(num),
[...]

PR #149, opened 2019-01-23
Alignment when using parentheses for negative values (#144)

steveputman:
@rich-iannone Here's a first cut. Code changes in `format_data.R` (I conformed the order of the latex/html/default functions in what I dealt with--that might have screwed up the diff, so happy to try to put that back together). Basically just takes the nonneg vector from `x_str` and applies leading and trailing hidden spaces, using `hphantom{}` in LaTeX and `<span>` in HTML.  Changes applied for LaTeX and HTML in `fmt_number()`, `fmt_percent`, and `fmt_currency`.


jcheng5:
@steveputman Thanks for this contribution! It'd be great if this change didn't result in the default formatter code being copy/pasted for html and latex, but instead to use the "function factory" pattern--see `fmt_scientific` for an example of this pattern at work.

steveputman:
@jcheng5 — got it. It was bugging me when I was copying and pasting! Will have a look and revert.

steveputman:
Added factory functions for `fmt_number`, `fmt_percent`, and `fmt_currency`. In addition to resolving the issues with repetition of my code, doing this also cleaned up some legacy duplication (avoided the need for `currency_str` vs `currency_str_html` vs `markdown_to_latex(currency_str)`, for example). 

I suppose there could be a megafactory function for just about all the functions in `format_data` but already felt I was overstepping.

I assigned the HTML and LaTeX used for all the nonneg formatting to vars near the top of the file but wonder if the HTML and LaTeX should just be repeated in the calls to the factory functions to avoid a sort of Where's Waldo problem. 

Any further guidance would be appreciated.

rich-iannone:
This looks really good. I sort of agree with the idea of a larger/better function for all of the functions here. However, there is still some not-yet-merged PRs that need to be integrated here first. I think that this could be part of a separate refactoring PR for this file.

I would put this inside the `utils.R` file:

```r
# Strings for alignment of non-negative number in formatting functions
 paren_nonneg_start_html <- "<span style=\"visibility: hidden;\">(</span>"
 paren_nonneg_end_html <- "<span style=\"visibility: hidden;\">)</span>"
 paren_nonneg_start_latex <- "\\hphantom{(}"
 paren_nonneg_end_latex <- "\\hphantom{)}"
```

Thanks again for your continued work on this. This will get merged soon!

steveputman:
Thanks! Contributor agreement sent.

steveputman:
Closing this PR; will replace with new PR to reflect refactoring of formatters

Issue #53, opened 2018-11-01
Create util function to extract column name from quosured column


rich-iannone:
There are many instances of ``columns %>% lapply(`[[`, 2) %>% as.character()`` throughout. Replace with a utility function.

PR #171, opened 2019-02-08
Improvements to resolving column indices and names

rich-iannone:
Here we split out `resolve_vars()` -> `resolve_vars()` (returns column names) & `resolve_vars_idx()` (returns column indices).

The `resolve_vars_idx()` now takes the place of the earlier `resolve_vars()` and the new `resolve_vars()` wraps `resolve_vars_idx()` and uses `data` (a variable that's available in almost every API function) to translate the column indices to column names.

This standardizes the means to resolve to column names and indices when the user provides column names in various ways.

Fixes https://github.com/rstudio/gt/issues/53


jcheng5:
Let's discuss this in person, there are some things I don't understand about the existing code.

PR #172, opened 2019-02-10
Fix for `fmt_scientific()` when using very large or small numbers

rich-iannone:
This fix modifies the `is_equal_to()` util function, which now uses an `identical()` comparison (instead of `all.equal()`) between two numbers. This has important implications for the `fmt_scientific()` formatter function because we compare whether a number is equal to zero and apply scientific notation when a number isn't zero (among other conditions not mentioned here).

Very small numbers would erroneously be deemed equal to zero before this change, resulting in incorrect formatting of such numbers in scientific notation.

Now, if we apply scientific notation to these numbers

`-1.5E200`, `-1.5E100`, `-2.5E0`, `-3.5E-100`, `-3.5E-200`, `1.5E-200`, `1.5E-100`, `2.5E0`, `3.5E100`, `3.5E200`

we get (in the default context)

`-1.50000 x 10(200)`, `-1.50000 x 10(100)`, `-2.50000`, `-3.50000 x 10(-100)`, `-3.50000 x 10(-200)`, `1.50000 x 10(-200)`, `1.50000 x 10(-100)`, `2.50000`, `3.50000 x 10(100)`, `3.50000 x 10(200)`

Tests have been added to ensure that very large or small numbers are formatted correctly in scientific notation.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -751,7 +751,7 @@ inline_html_styles <- function(html, css_tbl) {
 }
 
 is_equal_to <- function(x, y) {
-  vapply(x, function(a, b) isTRUE(all.equal(a, b)), logical(1), y)
+  vapply(x, function(a, b) isTRUE(identical(a, b)), logical(1), y)
 }
 
 split_scientific_notn <- function(x_str) {
--- a/R/utils.R
+++ b/R/utils.R
@@ -750,10 +750,6 @@ inline_html_styles <- function(html, css_tbl) {
   html
 }
 
-is_equal_to <- function(x, y) {
-  vapply(x, function(a, b) isTRUE(identical(a, b)), logical(1), y)
-}
-
 split_scientific_notn <- function(x_str) {
 
   exp_parts <- strsplit(x_str, "e|E")
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -330,7 +330,7 @@ fmt_scientific <- function(data,
       small_pos <-
         ((x >= 1 & x < 10) |
            (x <= -1 & x > -10) |
-           is_equal_to(x, 0)) & !is.na(x)
+           x == 0) & !is.na(x)
 
       # Create `x_str` with same length as `x`
       x_str <- rep(NA_character_, length(x))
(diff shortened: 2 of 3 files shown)


R/utils.R
```diff
@@ -751,7 +751,7 @@ inline_html_styles <- function(html, css_tbl) {
 }
 
 is_equal_to <- function(x, y) {
```
jcheng5: 1. If I'm reading this correctly, `x` can be a vector with any length and `y` must be a vector of length exactly 1? If so, the name is a bit confusing.
2. `isTRUE` is entirely redundant here, isn't it?
3. If you remove the `isTRUE`, the `function(a, b)` also becomes redundant, i.e. `vapply(x, identical, logical(1), y)`.
4. Couldn't you just replace `is_equal_to(x, y)` with `x == y` (other than NA/NaN handling)?

rich-iannone: For this, both `x` and `y` can be of any size (in `is_equal_to()`). Comparisons for each value of `x` will be made with the entire vector of `y` (although in our actual use of `is_equal_to()` we bind `0` to `y`). 

I agree that the `isTRUE` and the anonymous function are not required. The `vapply(x, identical, logical(1), y)` is all that's required.

`x == y` would indeed work fine in most cases. In the cases of:

```r
# used currently:
small_pos <-
  ((x >= 1 & x < 10) |
     (x <= -1 & x > -10) |
     is_equal_to(x, 0)) & !is.na(x)
```

and 

```r
# could be used instead:
small_pos <-
  ((x >= 1 & x < 10) |
     (x <= -1 & x > -10) |
     x == 0) & !is.na(x)
```

Both work because any `NA`s produced before the `&` would still result in an overall `FALSE`. 

I believe that this util function was made just in case we wanted to use it elsewhere but that hasn't been the case so far. But I'm in favor of removing it entirely and using `x == 0` in place of `is_equal_to()` for the one time it's called.

rich-iannone:
I've removed the util function (only used in one case) and instead used `==` where it was once called.

Issue #142, opened 2019-01-22
`summary_rows()` fails when `sum()` is used with a character column present


JakeHurlbut:
Below example is taken from gt's `summary_rows()` [docs](https://gt.rstudio.com/reference/summary_rows.html), with two changes:

1) There is a new column (named `testString`) of type `character` added before `gt()` is called. This is a column that is wanted in the final output, but is not included in `summary_rows()`
2) `sum()`is added to `summary_rows()`.

The second example produces the desired output (with warnings) if the `sum()` function is not present in `summary_rows()`. It seems like `summary_rows()` is still trying to execute functions on `testString` even though it isn't specified in the `column` argument of `summary_rows()`

``` r
library(gt)
library(dplyr)

test_data = sp500 %>%
  filter(date >= "2015-01-05" & date <= "2015-01-16") %>%
  arrange(date) %>%
  mutate(week = paste0("W", strftime(date, format = "%V")),
         testString = 'lorem ipsum') %>% # Change 1: Add a character column to the data which is desired in the final output
  select(-adj_close, -volume)

# Case 1: Error on sum()
error_table = test_data %>% gt(
    rowname_col = "date",
    groupname_col = "week"
  ) %>%
  summary_rows(
    groups = T,
    columns = vars(open, high, low, close),
    fns = list(
      sum = ~sum(.), # Change 2: Add a sum function to summary_rows
      min = ~min(.),
      max = ~max(.),
      avg = ~mean(.)),
    formatter = fmt_number,
    use_seps = F
  )

error_table
#> Error in summarise_impl(.data, dots): Evaluation error: invalid 'type' (character) of argument.



# Case 2: Desired Output (With Warnings) 
working_table = test_data %>% gt(
  rowname_col = "date",
  groupname_col = "week"
) %>%
  summary_rows(
    groups = T,
    columns = vars(open, high, low, close),
    fns = list(
      # sum = ~sum(.), # Change 2: Add a sum function to summary_rows (disabled for working example)
      min = ~min(.),
      max = ~max(.),
      avg = ~mean(.)),
    formatter = fmt_number,
    use_seps = F
  )

working_table
#> Warning in mean.default(.): argument is not numeric or logical: returning
#> NA

#> Warning in mean.default(.): argument is not numeric or logical: returning
#> NA
```

<div id="cppofzncff" style="overflow-x:auto;">

<table class="gt_table">

<tr>

<th class="gt_col_heading gt_left" rowspan="1" colspan="1">

</th>

[...]

Issue #68, opened 2018-11-03
Several problems with `summary_rows()`


rich-iannone:
The `summary_rows()` function is problematic in many cases. It seems to only work at all when:

- there is a stub column (using any of the methods to create a stub column with `gt()`)
- there are row groups (using any of the methods to create groups with `gt()` or `group_by(data) %>% gt()`)

Also, there are usability problems with the arguments:

- the `groups` argument can't use group names enclosed in `vars()` but only `c()` (this is more of a consistency issue since a lot of label selections can use both) 
- the `columns` argument (even though the default is `NULL`) must contain either a `vars()` or `c()` with columns to be included in the aggregation for the selected `groups`: `NULL` or `TRUE` yield errors (``Error: `.vars` must be a character/numeric vector or a `vars()` object, not logical``)

For reference, this currently works:

```
gt(iris %>% tibble::rownames_to_column(), groupname_col = "Species") %>%
  summary_rows(
    columns = vars(Sepal.Length, Sepal.Width, Petal.Length, Petal.Width),
    fns = list(avg = "mean", `s.d.` = "sd"))
```

But the user may not want/need row captions and also wouldn't like to type out all of the column names.

There is also another issue (https://github.com/rstudio/gt/issues/38) that puts forward that `summary_rows()` should also create a 'grand summary', that is, include all rows (irrespective of any groups they may or may not be part of) and create summary rows at the bottom of the table.

Thus, to plan for this extra functionality, the `groups` argument might take in the following to perform the intended actions:

1. Grand Summary Rows (creates rows at the bottom, not concerned with individual groups at all) – becomes the default behavior
```
.... %>%
summary_rows(
  groups = NULL,
  columns = vars(....),
  fns = list(....))
```

2. Summary Rows for All Groups (creates rows at the bottom of each group, uses a sentinel function like `all_groups()`)
```
.... %>%
summary_rows(
  groups = all_groups(),
  columns = vars(....),
  fns = list(....))
```

3. Summary Rows for Selected Groups (creates rows at the bottom of each group specified)
```
.... %>%
summary_rows(
  groups = vars(....),
  columns = vars(....),
  fns = list(....))
```

Very open to discussion on this topic.

StephieLaPugh:
Please, please, please, I will sell my soul for an R package that will let me make pretty tables with grand totals. 99% of the time, I need to produce output as boring, conservative tables for boring, conservative lawyers. And they have to have totals.

shrektan:
[...]

Issue #38, opened 2018-10-19
Allow the `summary_rows()` function to work when there are no row groups at all


rich-iannone:
Currently, this:

```
gt(mtcars) %>%
  summary_rows(
    columns = vars(disp),
    fns = c(~mean(., na.rm = TRUE), "sum"))
```

doesn't provide a summary at all. Ideally it should do that and, also, it shouldn't create a single group (as a workaround) if none have been specified.

PR #175, opened 2019-02-12
Bugfixes and improvements to `summary_rows()`

rich-iannone:
This addresses a bug (https://github.com/rstudio/gt/issues/142) where aggregation function would operate on all columns and retain only the results for the requested columns. The prior behavior easily results in failures when an aggregation function operates on any column with incompatible data.

The PR also adds functionality for grand summaries (https://github.com/rstudio/gt/issues/38, https://github.com/rstudio/gt/issues/68). If `groups == FALSE` we obtain a collection of summary rows that operate on all of the data in the table. We can have both groupwise summaries and grand summaries in the same output table. Grand summaries are distinguishable from groupwise summaries by use of a large double border (and placement at the bottom of the table)

Finally, the PR applies several fixes for summary label generation and catches several edge cases. It's both more flexible, less prone to failure, and stops (with nice messages) if some of the stricter rules aren't followed. The documentation has been updated to reflect these changes.

Here's a table using a groupwise summary (across all groups):

```r
sp500 %>%
  dplyr::filter(
    date >= "2015-01-05" &
      date <="2015-01-16"
  ) %>%
  dplyr::arrange(date) %>%
  dplyr::mutate(
    week = paste0(
      "W", strftime(date, format = "%V"))
  ) %>%
  dplyr::select(-adj_close, -volume) %>%
  gt(
    rowname_col = "date",
    groupname_col = "week"
  ) %>%
  summary_rows(
    groups = TRUE,
    fns = list(
      min = ~min(.),
      max = ~max(.),
      avg = ~mean(.)),
    formatter = fmt_number,
    use_seps = FALSE
  )
```

groupwise_summary

Here's a table using a grand summary:

```r
sp500 %>%
  dplyr::filter(
    date >= "2015-01-05" &
      date <="2015-01-16"
  ) %>%
  dplyr::arrange(date) %>%
  dplyr::mutate(
    week = paste0(
      "W", strftime(date, format = "%V"))
  ) %>%
  dplyr::select(-adj_close, -volume) %>%
  gt(
    rowname_col = "date",
    groupname_col = "week"
  ) %>%
  summary_rows(
    groups = FALSE,
    fns = list(
      min = ~min(.),
      max = ~max(.),
      avg = ~mean(.)),
    formatter = fmt_number,
    use_seps = FALSE
  )
```

grand_summary

Fixes https://github.com/rstudio/gt/issues/38
Fixes https://github.com/rstudio/gt/issues/68
Fixes https://github.com/rstudio/gt/issues/142


rich-iannone:
Some design considerations:

- `columns` should default to `TRUE` (this has the same effect as `NULL`, including all columns)
- `groups` as `FALSE` should no-op (`NULL` means grand summary, `TRUE` is all groups, `c()` specifies individual groups)
- there should be an opportunity to include a grand summary row label above the grand summary rows
- a location helper function for grand summary rows is necessary as are CSS rules for the new content areas

@jcheng5 please feel free to add anything I missed or refine what's here.

rich-iannone:
What if there are no groups but we call `summary_rows()` (and don't set `groups = NULL`)? I think that instead of stopping we should just provide the grand summary.

rich-iannone:
> What if there are no groups but we call `summary_rows()` (and don't set `groups = NULL`)? I think that instead of stopping we should just provide the grand summary.

By setting `groups = NULL` as a default, we always get a grand summary (which is more general than groupwise summaries and should always work).

R/tab_footnote.R
```diff
@@ -219,6 +219,26 @@ set_footnote.cells_summary <- function(loc, data, footnote) {
   data
 }
 
+set_footnote.cells_grand_summary <- function(loc, data, footnote) {
+
+  rows <- (loc$rows %>% as.character())[-1] %>% as.integer()
```
jcheng5: What's going on with `loc$rows` here? I was expecting it to use a resolve of some kind. And isn't it a quosure? I'm confused...

R/helpers.R
```diff
@@ -364,6 +364,30 @@ cells_summary <- function(groups = NULL,
   cells
 }
 
+#' @rdname location_cells
+#' @import rlang
+#' @export
+cells_grand_summary <- function(columns = NULL,
```
jcheng5: I assume valid options for `columns` are the usual suspects (logical, numeric, character, `NULL`; all `!!` aware), but what about for `rows`? Summary names, I assume? Logical/numeric too?

R/tab_style.R
```diff
@@ -252,3 +252,23 @@ set_style.cells_summary <- function(loc, data, style) {
 
   data
 }
+
+set_style.cells_grand_summary <- function(loc, data, style) {
+
+  rows <- (loc$rows %>% as.character())[-1] %>% as.integer()
```
jcheng5: Same question as above

R/tab_footnote.R
rich-iannone: This does need some work. With the recent changes from before (with `resolver.R`) I'll now use the `resolve_*()` functions to properly handle all of these cases.

R/helpers.R
rich-iannone: This will take some doing... row names for summary rows (be it groupwise summaries or grand summaries) are available if the user supplies them. If they are not supplied, then during render time (in `create_summary_dfs()` <`utils_render_common.R`>) they are created from quoted function names (e.g., `"mean"`) or from RHS formulas (e.g., `~mean(., na.rm = TRUE)`).

A good option I think is to have the creation of names (those not explicitly supplied) performed right in `summary_rows()`. Then we could have those names stored in the `summary` attribute for both resolving positions and for rendering those labels.

R/tab_style.R
rich-iannone: No changed to the newer form.

R/tab_footnote.R
rich-iannone: Now uses `resolve_vars()`.

R/helpers.R
rich-iannone: The `rows` argument is now as powerful as `columns`.

R/utils_render_common.R
```diff
@@ -315,107 +315,183 @@ perform_col_merge <- function(col_merge,
...
-        stub_df[
-          seq(nrow(stub_df)),
-          c("groupname", "rowname")],
-        data_df)[, -2]
+    if (groups[1] != "::GRAND_SUMMARY") {
+
+      select_data_df <-
+        cbind(
+          stub_df[
+            seq(nrow(stub_df)),
```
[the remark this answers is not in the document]
rich-iannone: This has now been simplified.

Issue #190, opened 2019-02-21
feature request: format markdown paragraphs


genepi10:
First of all, thank you so much for this package. I've produced many nice tables from it so far and people have been very impressed with the output. It's now part of my standard workflow for reporting!

I have a feature request that I'm sure many will find incredibly useful. Some of the tables I need to make have a lot of text within cells. I think it would be great to have a formatter for markdown text (could be called fmt_markdown). 

Here is an example:

```r
first <- 
"Here are some *bullet* points:
 - one
 - two
 - three
"

second <- 
  "
  Here is line one

  This is line two
  "

tibble(a = c(first, second)) %>% gt() %>% fmt_markdown(columns = vars(a, b))
```

These should appear in the table as you expect them from being translated from markdown to HTML.

rich-iannone:
Thanks for posting this issue! We've been getting more than a few issues related to formatting of text-based columns. I think this is a great idea since there are many examples of tables that can have qualitative comparisons (e.g., small paragraphs, bullet points, etc.) and currently there is no way to make this look acceptable.

PR #191, opened 2019-02-21
Add the `fmt_markdown()` function

rich-iannone:
This adds the formatter function, `fmt_markdown()`, that transforms Markdown-formatted text into the correct output format.

Here is an example of its use:

```r
first <- 
"Here is a **list** of numbers:
 - one
 - two
 - three
"

second <- 
"Here is line *one*, and

 this is line *two*."

tibble(markdown_text = c(first, second)) %>% 
  gt() %>%
  fmt_markdown(columns = vars(markdown_text))
```

fmt_markdown_example

Fixes https://github.com/rstudio/gt/issues/190.


rich-iannone:
Note: the CSS rule should be fixed such that the first and last `<p>` has a 0 margin.

PR #193, opened 2019-02-23
Avoid formatting of cells with NA values

rich-iannone:
The reason for this change is to make is very easy to format values in one pass and handle missing values in another sweep—without explicitly specifying which cells have those missing values. To illustrate the problem and the benefits of the solution, let's use the `sza` dataset for its intended purpose: creating a table of solar zenith angles during sunrise for every month at a reference latitude. Here's the code:

```r
library(gt)
library(tidyverse)

sza %>%
  dplyr::filter(latitude == 20) %>%
  dplyr::select(-latitude) %>%
  gt(rowname_col = "month") %>%
  fmt_number(
    columns = TRUE,
    decimals = 1
  ) %>%
  fmt_missing(
    columns = TRUE,
    missing_text = ""
  ) %>%
  cols_align(
    align = "center",
    columns = TRUE
  ) %>%
  tab_stubhead_label(label = html("month<br>(20&deg;N)")) %>%
  tab_header(title = html("&#x2600; Solar Zenith Angles &#x2600;")) %>%
  tab_options(
    column_labels.font.size = "smaller",
    table.font.size = "smaller",
    row.padding = px(3)
  )
```

Here is the gt table before the PR changes:

sza_no_changes

And here is the better version with the changes of the PR in place:
 
sza_with_changes

Before the PR, the second formatter would *win* and nullify the numeric formatting (notice that some numbers have no decimal values). Changing the order of the formatters would have all numbers being formatted to a single decimal place, but `NA` values would be present.

The new way allows for any ordering of the two formatters because `fmt_missing()` won't interfere with cells operated on by the other formatters (which avoid processing `NA` cells).

Fixes https://github.com/rstudio/gt/issues/79.


rich-iannone:
Note: testthat tests are now required.

PR #195, opened 2019-02-24
dplyr::data_frame deprecated

martinjhnhadley:
Use tibble::tibble() instead


rich-iannone:
Thanks for the PR! This looks really good but the one commit that's not really necessary is https://github.com/rstudio/gt/pull/195/commits/168a649c1995f855a1c8f20f46feff200735a152. We can keep using `dplyr::tibble()` in the vignette (and in the package functions) because that function in **dplyr** will continue to be re-exported from **tibble**.

martinjhnhadley:
Ah, thanks for pointing out that! I've reverted that commit for you.

PR #222, opened 2019-03-13
Use jcheng5 GitHub PAT instead of rich-iannone

jcheng5:


rich-iannone:
Looks like it’s the right length. Hoping this will work (and there’s no way to know unless merged).

PR #244, opened 2019-04-03
Make several refactoring improvements to some `format_*()` functions

rich-iannone:
These changes provide a few more improvements to some `format_*()` functions and related utility functions.

### DIFF
--- a/R/utils_formatters.R
+++ b/R/utils_formatters.R
@@ -196,10 +196,10 @@ scale_x_values <- function(x,
 #' @param format The numeric format for `formatC()`.
 #' @noRd
 format_num_to_str <- function(x,
-                              decimals,
-                              sep_mark,
-                              dec_mark,
-                              drop_trailing_zeros,
+                              decimals = NULL,
+                              sep_mark = NULL,
+                              dec_mark = NULL,
+                              drop_trailing_zeros = NULL,
                               small_pos = NULL,
                               exp_marks = NULL,
                               minus_mark = NULL,
@@ -221,13 +221,13 @@ format_num_to_str <- function(x,
 #' @inheritParams format_num_to_str
 #' @noRd
 format_num_to_str_e <- function(x,
-                                decimals,
-                                sep_mark,
-                                dec_mark,
-                                drop_trailing_zeros,
-                                small_pos,
-                                exp_marks,
-                                minus_mark) {
+                                decimals = NULL,
+                                sep_mark = NULL,
+                                dec_mark = NULL,
+                                drop_trailing_zeros = NULL,
+                                small_pos = NULL,
+                                exp_marks = NULL,
+                                minus_mark = NULL) {
 
   format_num_to_str(
     x,
@@ -246,9 +246,9 @@ format_num_to_str_e <- function(x,
 #' @inheritParams format_num_to_str
 #' @noRd
 format_num_to_str_c <- function(x,
-                                decimals,
-                                sep_mark,
-                                dec_mark,
+                                decimals = NULL,
+                                sep_mark = NULL,
+                                dec_mark = NULL,
                                 drop_trailing_zeros = FALSE,
                                 small_pos = NULL,
                                 exp_marks = NULL,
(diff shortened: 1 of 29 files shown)


R/utils_formatters.R
```diff
@@ -307,14 +295,18 @@ context_percent_mark <- function(context) {
...
+                                 currency) {
 
-  switch(context,
-         latex = "\\\\$",
-         "\\$")
-}
+  # If we supply `NULL` as currency, then
+  # return an empty string
+  if (is.null(currency)) {
+    return("")
```
rich-iannone: Return `NULL` instead.

R/utils_formatters.R
```diff
@@ -361,9 +358,174 @@ paste_currency_str <- function(x,
...
 }
+
+#' @noRd
+format_in_accounting_style <- function(x,
+                                       x_vals,
+                                       accounting,
+                                       minus_mark,
+                                       parens_marks) {
+
+  # Return values unchanged if there are no negative values
```
rich-iannone: Store `x_vals < 0` as a variable for repeated use.

R/utils_formatters.R
```diff
@@ -361,9 +358,174 @@ paste_currency_str <- function(x,
...
+    )
+
+  # Handle replacement of the minus mark in number
+  # and exponent parts
+  x %>%
+    tidy_gsub("-", minus_mark, fixed = TRUE)
+}
+
+#' @noRd
+num_formatter_factory <- function(context,
```
rich-iannone: Remove `NULL` defaults.

R/utils_formatters.R
```diff
@@ -361,9 +358,174 @@ paste_currency_str <- function(x,
...
+                                  dec_mark = NULL,
+                                  currency = NULL,
+                                  drop_trailing_zeros = NULL,
+                                  accounting = NULL,
+                                  incl_space = NULL,
+                                  placement = NULL,
+                                  pattern = NULL,
+                                  format_fn = NULL) {
+
+  force(context)
```
rich-iannone: Explore possibility of `force()`ing vars in a list.

rich-iannone: `force(list(a, b, c, ...))`

R/utils_formatters.R
```diff
@@ -361,9 +358,174 @@ paste_currency_str <- function(x,
...
+    # for scientific foramtting since their order would be zero
+    small_pos <- has_order_zero(x_vals)
+
+    # Format all non-NA x values with a formatting function
+    x_str_vals <-
+      format_fn(x_vals, decimals, sep_mark, dec_mark, drop_trailing_zeros)
+
+    # Format scientific notation values in a way that
+    # is prettier than the form offered by `formatC()`
+    x_str_vals <-
```
rich-iannone: Intersperse comments and function calls in piped sequence.

R/utils_formatters.R
rich-iannone: @schloerke this seemed to work in my testing (no R error about syntax). However, do you know of any way to verify that the variables are indeed forced?

rich-iannone: I tested this to ensure that the forcing works with a list.

jcheng5:
Addendum to 5. in my comment above: rather than forcing all of the number-formatting-related arguments, just force the ones that aren't passed directly to `num_formatter_factory` or whatever; i.e. the ones you use yourself in `format_fn`. It'll make sense after you do step 2.

Is there a reason you didn't push the calls to `get_locale_sep_mark`, `get_locale_dec_mark`, and `normalize_suffixing_inputs` into `num_formatter_factory` as well? I'm not necessarily saying it's wrong the way it is, just curious if there was a specific reason.

rich-iannone:
Thanks for the quick turnaround with the review here! I'll go through these changes but I'll answer one or two of the questions above first.

The `format_num_to_str_e()` and `format_num_to_str_c()` are separated just so we can have targeted tests (which don't yet exist). 

There is no reason that the calls to `get_locale_sep_mark()`, `get_locale_dec_mark()`, and `normalize_suffixing_inputs()` aren't in the factory. I'll work toward putting those in.

Issue #247, opened 2019-04-05
execute html formatted text in spanner column label


shannonpileggi:
The `html()` helper function does not work with `label` in  `tab_spanner()`. 

---

``` r
library(tidyverse)
library(gt)

airquality_m <- 
  airquality %>%
  mutate(Year = 1973L) %>%
  slice(1:10) %>%
  select(Month, Day, Year)
  
# Create a display table using the `airquality`
# dataset; arrange columns into groups
gt_tbl <- 
  gt(data = airquality_m) %>%
  tab_header(
    title = "New York Air Quality Measurements",
    subtitle = "Daily measurements in New York City (May 1-10, 1973)"
  ) %>%
  tab_spanner(
    label = html("Time <br> n = 10"),
    columns = vars(Year, Month, Day)
  )

# Show the gt Table
gt_tbl
```

<style>html {
  font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen, Ubuntu, Cantarell, 'Helvetica Neue', 'Fira Sans', 'Droid Sans', Arial, sans-serif;
}

#yljoxpsmnf .gt_table {
  display: table;
  border-collapse: collapse;
  margin-left: auto;
  margin-right: auto;
  color: #000000;
  font-size: 16px;
  background-color: #FFFFFF;
  /* table.background.color */
  width: auto;
  /* table.width */
  border-top-style: solid;
  /* table.border.top.style */
  border-top-width: 2px;
  /* table.border.top.width */
  border-top-color: #A8A8A8;
  /* table.border.top.color */
}

#yljoxpsmnf .gt_heading {
  background-color: #FFFFFF;
  /* heading.background.color */
  border-bottom-color: #FFFFFF;
}

#yljoxpsmnf .gt_title {
  color: #000000;
  font-size: 125%;
  /* heading.title.font.size */
  padding-top: 4px;
  /* heading.top.padding */
  padding-bottom: 1px;
  border-bottom-color: #FFFFFF;
  border-bottom-width: 0;
}

#yljoxpsmnf .gt_subtitle {
  color: #000000;
  font-size: 85%;
  /* heading.subtitle.font.size */
  padding-top: 1px;
  padding-bottom: 4px;
  /* heading.bottom.padding */
  border-top-color: #FFFFFF;
  border-top-width: 0;
}

#yljoxpsmnf .gt_bottom_border {
  border-bottom-style: solid;
  /* heading.border.bottom.style */
[...]

Issue #178, opened 2019-02-13
`gt::latex` for regional modification for LaTex outputs?


sammo3182:
Given the `gt::md` and `gt::html`, could I suggest an additional input-output converter `gt::latex`? In version 0.1.0, plain latex codes are escaped, e.g., inputting `\\` becoming `\textbackslash` in the output. In some cases, though, it would be useful to call for the LaTex functions. For example, right now if the content is in double-space, so do the notes from the `tab_source_note`. It will be wonderful to have a `gt::latex` function allowing users to create a local single-space environment to make those notes more like notes rather than paragraphs after the table by something like`tab_source_note(latex("\\parbox{\\linewidth}{\\vspace{2pt} The single-space notes.}"))`.

This suggestion may relate to other suggestions to improve LaTex outputs, such as in #111, #119, and #176.

rich-iannone:
This is a great idea, thanks for submitting the issue. There will be some efforts toward having LaTeX functionality be more consistent with HTML and this will be part of that.

PR #251, opened 2019-04-09
Fix for Markdown/HTML-formatting of spanner column labels

rich-iannone:
Using the `md()` or `html()` helper function to format spanner column labels became broken after function used to detect the `html` class was created and used (because it would fail to detect the `html` class, it would `htmlEscape()` the HTML text).

This fixes the utility function (now renamed as `is_html()`) and adds several tests to ensure that the HTML is preserved and also that Markdown formatting for spanner column labels works.

Here is an example, with the changes from the PR:

```
library(tidyverse)
library(gt)

airquality_m <- 
  airquality %>%
  mutate(Year = 1973L) %>%
  slice(1:10) %>%
  select(Month, Day, Year)

# Create a display table using the `airquality`
# dataset; arrange columns into groups
gt_tbl <- 
  gt(data = airquality_m) %>%
  tab_header(
    title = "New York Air Quality Measurements",
    subtitle = "Daily measurements in New York City (May 1-10, 1973)"
  ) %>%
  tab_spanner(
    label = html("Time <br> n = 10"),
    columns = vars(Year, Month, Day)
  )

# Show the gt table
gt_tbl
```

column_spanner_label_html

Fixes https://github.com/rstudio/gt/issues/247


R/helpers.R
```diff
@@ -420,13 +420,11 @@ html <- function(text, ...) {
...
-  if (inherits(x, "html")) {
-    isTRUE(attr(x, "html"))
-  } else {
-    FALSE
-  }
+#' Does an object have the `html` class?
+#'
+#' @noRd
+is_html <- function(x) {
+  ifelse(inherits(x, "html"), TRUE, FALSE)
```
jcheng5: Is there a reason this isn't just `inherits(x, "html")`?

Also, I'd avoid `ifelse` unless you specifically are doing a _vectorized_ conditional; if it's just about saving some curly braces, I'd still use `if`/`else` since the semantics are more sensible.

R/table_parts.R
```diff
@@ -272,6 +272,11 @@ tab_spanner <- function(data,
   # of the `grp_labels` list
   grp_labels[column_names] <- label
 
+  # Ensure that the label classes are preserved
+  for (group_label_i in seq(grp_labels[column_names])) {
+    class(grp_labels[column_names][[group_label_i]]) <- class(label)
+  }
```
jcheng5: Hmmm, this is necessary because `grp_labels[columns_names] <- label` automatically drops the class? That's a bit scary--that's an easy mistake to make! Would it help to have a version of `rep_len` that preserves the class, and assign that to `grp_labels[column_names]`?

Also, what's the type of `grp_labels[column_names]`, anyway? Shouldn't it be a list of lists?

jcheng5:
Could you create a unit test that uses `md` or `html` for _every single_ parameter that takes such values?

Will there be an analogous `latex()` function in the future?

R/helpers.R
rich-iannone: This is now simplified (😬) and your advice on `ifelse()` is good (and I've remember being burned with improper `ifelse()`s before).

R/table_parts.R
rich-iannone: This is now rewritten to produce a `labels` list of lists of length `grp_labels[column_names]`. With that, the assignment of `grp_labels[column_names] <- labels` produces a list of lists (as before).

rich-iannone:
> Could you create a unit test that uses `md` or `html` for _every single_ parameter that takes such values?

These tests are now added in the `test-h_md_html.R` file.

> Will there be an analogous `latex()` function in the future?

Yes, this is https://github.com/rstudio/gt/issues/178. Once we get back to working on LaTeX support, we will add this helper function.

Issue #249, opened 2019-04-08
group column with `<=` sign dispaly wrong?


shizidushu:
It display "30<x<=60" as "30" (see the image below). How to make it work?

``` r
library(gt)
data.frame(age = c("0<=x<=30", "30<x<=60"), value = seq(1,10)) %>%
  gt(groupname_col = "age")
```
![result](https://user-images.githubusercontent.com/13694980/55703430-2a761580-5a0c-11e9-9fd1-dd67c00d69f6.png)

Edit:

Replace `<` with `&lt`, then it works.

But I think we need a better solution. The dataset column should be treat as it is, not as html. Something like `escape` argument in `DT::datatable` might be useful.

```r
library(gt)
data.frame(age = c("0<=x<=30", "30<x<=60"), value = seq(1,10)) %>%
  mutate(age = stringr::str_replace(age, "<", "&lt")) %>%
  gt(groupname_col = "age")
```

rich-iannone:
Thanks for reporting this! This is definitely a bug: the row group labels need escaping.

PR #252, opened 2019-04-10
Escaping for row group label

rich-iannone:
This fixes an HTML rendering bug, where row group labels aren't escaped at render time. For example, the following code

```r
library(gt)

data.frame(age = c("0<=x<=30", "30<x<=60"), value = seq(1,10)) %>%
  gt(groupname_col = "age")
```

gives us

unescaped_row_group_label

This is because the unescaped `<` results in the completion of the innerHTML.

With the fix provided here, this example now works:

escaped_row_group_label

Fixes https://github.com/rstudio/gt/issues/249.


shrektan:
Is there a way to **not** escape the row group label? I have a shiny app that use html to decorate the row group label, now it displays raw HTML text...

Issue #256, opened 2019-04-13
Saving of image files from gt tables


rich-iannone:
Add ability to create image files from gt tables (using the HTML output) through `gtsave()`.

PR #257, opened 2019-04-13
Use webshot to generate images of HTML tables

rich-iannone:
With this PR, we can use the webshot package's `webshot()` function to generate image files based on the HTML table. The image files are cropped to the `table` element and contain some padding as well. This is handled through the `gtsave()` function with the `gt_save_webshot()` function used in the switch statement (which calls different output functions based on the output file's extension).

We check if the webshot package is installed with `requireNamespace()` and suggest that it should be installed if it isn't available.

The following statement

```r

exibble %>%
  gt() %>%
  tab_header(
    title = md("The exibble dataset")
  ) %>%
  gtsave("ex.png")
```

Gives us this output:

ex

Fixes https://github.com/rstudio/gt/issues/256.
Fixes https://github.com/rstudio/gt/issues/96.


R/gtsave.R
```diff
@@ -78,22 +93,58 @@ gt_save_html <- function(data, filename, ..., inline_css = FALSE) {
...
-# Saving function for a LaTeX file
+#' Saving function for a LaTeX file
+#'
+#' @importFrom htmltools HTML save_html
+#' @noRd
+gt_save_webshot <- function(data, filename, ...) {
+
+  tempfile_ <- tempfile(fileext = ".html")
+
+  data %>%
```
jcheng5: Wouldn't this be better as `data %>% htmltools::as.tags() %>% htmltools::save_html(...)`? I think this would allow html dependencies that you might want to bring along, to be included.

R/gtsave.R
```diff
@@ -78,22 +93,58 @@ gt_save_html <- function(data, filename, ..., inline_css = FALSE) {
...
+    webshot::webshot(
+      url = paste0("file://", tempfile_),
+      file = filename,
+      selector = "table",
+      zoom = 2,
+      expand = 5
+    )
+
+  } else {
+    stop("The `webshot` package is required saving images of gt tables.",
```
jcheng5: Missing a word in this error message

R/gtsave.R
```diff
@@ -78,22 +93,58 @@ gt_save_html <- function(data, filename, ..., inline_css = FALSE) {
...
+  data %>%
+    as_raw_html(inline_css = TRUE) %>%
+    htmltools::HTML() %>%
+    htmltools::save_html(file = tempfile_)
+
+  if (requireNamespace("webshot", quietly = TRUE)) {
+
+    # Save the image in the working directory
+    webshot::webshot(
+      url = paste0("file://", tempfile_),
```
jcheng5: Can you make sure to try this on Windows? I'd want to make sure it works fine with Windows-style paths (i.e. starting with a drive letter) and also with paths that have spaces (which I think Windows temp paths do, by default).

R/gtsave.R
```diff
@@ -78,22 +93,58 @@ gt_save_html <- function(data, filename, ..., inline_css = FALSE) {
   }
 }
 
-# Saving function for a LaTeX file
+#' Saving function for a LaTeX file
+#'
+#' @importFrom htmltools HTML save_html
+#' @noRd
+gt_save_webshot <- function(data, filename, ...) {
```
jcheng5: What is the `...` for? Did you mean to pass it to webshot or something?

R/gtsave.R
rich-iannone: Thanks! Now fixed.

R/gtsave.R
rich-iannone: Yes, I agree. Changed this to use the `gt_save_html()` function (with the default option to *not* inline the HTML styles this time). This call performs the following:

```r
data %>%
  htmltools::as.tags() %>%
  htmltools::save_html(filename)
```

R/gtsave.R
rich-iannone: I will test this on Windows for sure.

R/gtsave.R
rich-iannone: It's currently unused, but there may be options for webshot (like `zoom`, `expand`) that we may expose after the `...` in the future. In this way, the `...` forces users to name the args after the `...` (the one current example is `gt_save_html()`).  Also, the `gtsave()` function has the signature `data, filename, ...`.

Let me know if any of this sounds off base. We can discuss.

Issue #258, opened 2019-04-15
Rename the `cells_styles()` function as `cell_text()`; create `cell_fill()` fcn


rich-iannone:
Rename the `cells_styles()` function as `cell_text()`. This helper function should only be concerned with cell text so all `text_*` args should be renamed to lose the `text_` part.

rich-iannone:
Additionally, the `cell_fill()` helper function needs to be created to handle cell fill parameters. Currently, the `bkgd_color` argument is present in `cells_styles()`. It needs to be in `cell_fill()` as `color`.

rich-iannone:
Fixed in #239.

Issue #245, opened 2019-04-04
Add helper function `cell_borders()` to provide options for formatting borders


rich-iannone:
It would be useful to add the `cell_borders()` function to target and style cell borders. This would work in conjunction with the `style` argument of `tab_style()`.

rich-iannone:
Fixed in #239.

PR #239, opened 2019-04-01
Refactoring of heading component

rich-iannone:
This builds upon the earlier refactoring effort, which made available several new utility functions for composing strings. Additional utility functions are included here to help with gathering footnote glyphs and custom styles for cells.

Templating with `glue` is used here to compose the HTML and LaTeX fragments (along with `paste_between()`). Some improvements to the CSS classes for the title and subtitle were also made.

PR #259, opened 2019-04-16
[WIP] Add the `cell_borders()` helper and update implementation of the stylizing helper fcns

rich-iannone:
This adds the `cell_borders()` helper function for selectively modifying cell borders in a `tab_style()` call. The existing `cells_styles()` helper was reworked as `cell_text()` and the single argument dealing with the cell background color was put in `cell_fill()`.

These three stylizing helper functions can be placed in a `list()` of calls for the `style` argument of `tab_style()`, allowing us to target cells and perform multiple stylizing operations on the targets at once.

Here is an example:

```r
library(gt)

exibble %>%
  gt() %>%
  tab_style(
    style = list(
      cell_borders(
        selection = c("left", "right"),
        color = "green",
        style = "solid",
        weight = px(3)
      ),
      cell_fill(color = "lightgray"),
      cell_text(color = "steelblue", font = "HelveticaNeue")
    ),
    locations = cells_data(columns = vars(row, char))
  )
```

exibble_styles

Fixes https://github.com/rstudio/gt/issues/258.
Fixes https://github.com/rstudio/gt/issues/245.

cc: https://github.com/ddsjoberg/gtsummary/issues/78


R/data_color.R
```diff
@@ -129,7 +129,7 @@ data_color <- function(data,
                        columns,
                        colors,
                        alpha = NULL,
-                       apply_to = "bkgd",
+                       apply_to = "fill",
```
jcheng5: Note: BREAKING change. At the very least, consider putting a prominent note in NEWS. Consider checking for `"bkgd"` in the body of the function, and respond appropriately (maybe warn, then act as if they asked for `"fill"`?).

jcheng5:
This is an incremental change on top of a whole pile of code I have not reviewed yet. Besides normal code quality stuff, I'd want to focus on whether you have a vision for making the styling features portable between HTML, LaTeX, and RTF.

R/data_color.R
jcheng5: Same with the move from `cells_styles` to `cell_text` and `cell_fill`, do you have a plan for how to notify?

rich-iannone:
Using #239 instead.

Issue #141, opened 2019-01-22
Disable random div id generation?


andrewheiss:
Right now, [print.R generates a random name](https://github.com/rstudio/gt/blob/0df27115d1453c5c100c98aaeb189c5a34d522a4/R/print.R#L47) for the wrapping div for gt tables (e.g. `<div id="joisdfjkdf">`). This is nice for uniqueness in HTML, but it causes issues with version control, since the table ID changes often (forcing new changes in git). 

Right now, my solution is to use `set.seed()` before any `gt()` stuff, and that's working, but it might be nice to have an argument to `gt()` or something that disables the random id generation and allows users to supply their own.

rich-iannone:
This sounds like a good idea. Thanks for suggesting it!

PR #286, opened 2019-05-08
Have the table ID be settable, random, or absent in `gt()`

rich-iannone:
This change allows us to more explicitly set the table ID though the `id` option in `gt()`. By default this is set to use the new helper function `id_random()`, which is equivalent to the current behavior of having a 10-letter random string being generated with each `gt()` call.

We can use a custom string here, or, use `NULL` to entirely disable ID generation. Here's a basic usage example where we set a custom `id` value:

```r
library(gt)

exibble %>% gt(id = "name")
```

The following screen capture shows that the `id` for the enclosing `<div>` is now set to `"name"`:

gt_id_set

Here's another showing that the `id` attr can be omitted entirely:

```r
library(gt)

exibble %>% gt(id = NULL)
```

gt_id_null

Fixes: https://github.com/rstudio/gt/issues/141.


NAMESPACE
```diff
@@ -50,6 +50,7 @@ export(gt_output)
 export(gt_preview)
 export(gtsave)
 export(html)
+export(id_random)
```
schloerke: Thoughts on `id_random` vs `random_id`?

R/gt_options_default.R
```diff
@@ -2,6 +2,7 @@ gt_options_default <- function() {
 
   dplyr::tribble(
     ~parameter,                           ~scss, ~category,             ~value,
+    "table_id",                          FALSE,  "table",               NA_character_,
```
schloerke: How does this connect to the rest of the PR?

NAMESPACE
rich-iannone: The **highcharter** package has an exported function called `random_id()`. Do you think we should just keep using `random_id()` and basically not care about the conflict? `random_id()` is a way better name so I'd probably like to keep it.

schloerke: `gt_it()`?

R/gt_options_default.R
rich-iannone: The table ID value is stored here. And it's later retrieved in the `as.tags.gt_tbl()` function.

schloerke: Ha! Correct. Missed it. Thanks

NAMESPACE
rich-iannone: That solves the namespacing issue but doesn't give the impression that a random ID is being created. Really though, if we use `random_id()` from **highcharter** on accident, it would still create a random ID (just a bit shorter in length).

Issue #128, opened 2019-01-08
Row Group Labels: Should be spanning the entire table (or have an option to enable it)


r-cheologist:
It's customary to have verbose row group labels (as one conceptually has the whole table width to work with in terms of horizontal space).

At least in the context of `mydataframe %>% gt::gt(groupname_col = '<SOMECOLUMN>')` and `bookdown`-mediated output to `LaTeX`/`PDF`, however, this currently does not work and the group labels only occupy the first column, potentially rendering it overly wide.

Consider:

    ---
    title: Bug Report
    subtitle: Non--Spanning Row Group Labels
    date: '`r format(Sys.time(), "%d %B, %Y")`'
    author: r-cheologist
    output:
      bookdown::pdf_document2
    ---
    
    ```{r prerequisites, message = FALSE}
    library(magrittr)
    requireNamespace('gt')
    ```
    
    ```{r table}
    data.frame(
      A      = 1:5,
      B      = rep('Longer than longish group label', times = 5),
      groups = c(
        rep('Longish group label', times = 2),
        rep('Label', times = 3))) %>%
      gt::gt(groupname_col = 'groups')
    ```
which gives rise to ![this](https://user-images.githubusercontent.com/3259333/50851420-107d5600-137d-11e9-9631-262580ad5796.png) ...

peterparkerspicklepatch:
Yup, this is a great feature addition and what I'm looking for! Having a group span the space of the relevant rows provides the same value as spanner column labels.

rich-iannone:
Now addressed in a recent commit. Will be in the CRAN v0.7.0 release of gt soon.

peterparkerspicklepatch:
Is this real life? H-Y-P-E

sanjmeh:
Rich you are doing a fantastic job. The gt() package is so intuitive. Waiting for this and the row spanners to be activated.

rich-iannone:
Thank you @sanjmeh , very nice of you to say!

PR #287, opened 2019-05-09
Make row group labels span the entire table (#128)

onesandzeroes:
I just tried out `gt` for the first time this week and found it very useful immediately, so thanks! I did run into issue #128 though, so I thought I would try a PR.

It produces the expected behaviour for the reprex in the original issue:

latex_row_group_labels

I also added a couple of basic tests for tables with + without rownames so the basic functionality should be tested.

Let me know if any this needs anything else.


rich-iannone:
Somewhere recently when working on LaTeX rendering this was addressed. Thank you for your contribution.

PR #306, opened 2019-06-13
Address column alignment issues when using accounting option

steveputman:
This PR attempts to address the alignment issues raised in #144 and initially contended with in #149. Since that PR was submitted, the formatters have been refactored, and this PR reflects the new formatters. I have also added accounting options to `fmt_number()` and `fmt_percent()`, because I use those all the time. Corresponding documentation and tests added.


jcheng5:
Thanks so much for this contribution, and sorry it's taken so long for us to get to it!

The only problem I see is that the column label (which defaults to right-aligned for numbers) will be aligned with the paren, not with the number. I think we can solve this by wrapping the right-paren with `\rlap{}` in LaTeX, and `<span style="position: absolute;">` for HTML (not sure yet how to deal with this in RTF). @rich-iannone is going to take a pass at this now.

steveputman:
Admittedly, I did not take column-label alignment into account, so thanks for the heads-up there. And I have not revisited this since RTF came into the picture, so let me know if I need to do anything on that front. Thanks again to you and Richard--just an awesome package.

koppsp (approved):
It appears from the code diff that you can now have parenthesis for negative values without dollar sign for "accounting", among other overall increased flexibility

steveputman:
Just keeping my branch up to date to reflect your addition of the accounting option to `fmt_percent` and `fmt_currency`. (and retitling PR to reflect that it's only contending with alignment). Also implemented @jcheng5's suggestions to deal with alignment relative to column label by wrapping the right-paren with `\rlap{}` in LaTeX, and `<span style="position: absolute;">` in HTML. Doing that does create a couple of problems that I could see:

- Overlapping text when `accounting` is used alongside `pattern` if `pattern` decorates the end of the formatted number;
- Some unwanted HTML output in RTF

Congratulations on 0.3!

steveputman:
Addressed by #1092

Issue #325, opened 2019-07-22
Installation instructions do not work because `sass` is not available on CRAN


vincentarelbundock:
The `sass` package is no longer available on CRAN. Since `gt` depends on `sass`, the installation instructions using `remotes::install_github('rstudio/gt')` no longer work.

I opened an issue on the `sass` github.

https://github.com/rstudio/sass/issues/22

In the meantime, it might be useful to change the installation instructions in the README to say:

```r
remotes::install_github('rstudio/sass')
remotes::install_github('rstudio/gt')
```

rich-iannone:
Thanks for posting these issues. We will modify a few things while **sass** gets fixed and resubmitted to CRAN.

PR #326, opened 2019-07-22
Add `rstudio/sass` to Remotes list

rich-iannone:
This change is necessary while **sass** is being fixed for resubmission to CRAN.

Fixes https://github.com/rstudio/gt/issues/325.


rich-iannone:
Agreed!

PR #331, opened 2019-07-30
Export `as.tags.gt_tbl()`

rich-iannone:
This change simply exports `as.tags.gt_tbl()` and provides internal documentation.


NAMESPACE
```diff
@@ -3,6 +3,7 @@
 S3method(print,gt_tbl)
 export("%>%")
 export(adjust_luminance)
+export(as.tags.gt_tbl)
```
jcheng5: This needs to be `S3method(as.tags,gt_tbl)`, there should be some way to convince roxygen to do that? @wch?

jcheng5: Oh, adding `@method as.tags gt_tbl` does the trick. Not sure why it's necessary.

rich-iannone: Thanks for checking the NAMESPACE for this. This is now updated.

rich-iannone: I also reexported `htmltools::as.tags` for this to work (`htmltools` is in Imports).

jcheng5:
Sorry for pushing onto your PR but it was taking too long to write up what I thought should be done. I think with the technique in my commit, there's no need to re-export `as.tags`.

Here are the tests I tried (each one in a totally fresh R session):

```r
library(gt); htmltools::div(gt::gt(cars))
```

```r
library(htmltools); htmltools::div(gt::gt(cars))
```

```r
htmltools::div(gt::gt(cars))
```

rich-iannone:
This is great, Joe. I did the same `register_s3_method()` but didn't take out the `@export` directive at `as.tags.gt_tbl()`!

PR #336, opened 2019-08-02
Allow use of integers in `pct()`/`px()`

rich-iannone:
Both the `pct()` and `px()` don't allow `integer`-class values as input (only `numeric`). This PR allows the use of integers, so, this will now work:

```r
sp500 %>%
  head() %>%
  gt() %>%
  tab_header(
    title = "S&P 500",
    subtitle = "5 rows"
  ) %>%
  tab_spanner(
    label = "prices",
    columns = vars(open, high, low, close)
  ) %>%
  fmt_date(
    columns = vars(date),
    date_style = 2
  ) %>%
  fmt_currency(
    currency = "USD"
  ) %>%
  fmt_number(
    columns = vars(volume),
    suffixing = TRUE
  ) %>%
  tab_options(
    table.font.size = px(10L)
  )
```

Fixes: https://github.com/rstudio/gt/issues/307.

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -1003,7 +1003,7 @@ cell_style_structure <- function(name, obj, subclass = name) {
 #' @export
 pct <- function(x) {
 
-  if (!inherits(x, "numeric") && !inherits(x, "integer")) {
+  if (mode(x) != "numeric") {
     stop("The supplied value must be numeric", call. = FALSE)
   }
 
@@ -1018,7 +1018,7 @@ pct <- function(x) {
 #' @export
 px <- function(x) {
 
-  if (!inherits(x, "numeric") && !inherits(x, "integer")) {
+  if (mode(x) != "numeric") {
     stop("The supplied value must be numeric", call. = FALSE)
   }
 
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -1003,7 +1003,7 @@ cell_style_structure <- function(name, obj, subclass = name) {
 #' @export
 pct <- function(x) {
 
-  if (!inherits(x, "numeric")) {
+  if (!inherits(x, "numeric") && !inherits(x, "integer")) {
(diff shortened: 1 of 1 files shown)


R/helpers.R
```diff
@@ -1003,7 +1003,7 @@ cell_style_structure <- function(name, obj, subclass = name) {
 #' @export
 pct <- function(x) {
 
-  if (!inherits(x, "numeric")) {
+  if (!inherits(x, "numeric") && !inherits(x, "integer")) {
```
schloerke: Maybe 

```r
if (mode(x) != "numeric") {
```

rich-iannone: Yeah, this is better. Made that change.

r2evans: Minor, perhaps, but isn't `!is.numeric(x)` faster (6-10x for this scale, likely not a real factor), just as declarative, and sufficiently-encompassing that it would work here, too? Asked another way, what is better about `inherits` or `mode` in this case?

rich-iannone: I think it might be faster and more correct. For example:

```r
# We wouldn't want a `Date` as input
Sys.Date() %>% mode()
#>  "numeric"

Sys.Date() %>%  typeof()
#> "double"

# This is better
Sys.Date() %>% is.numeric()
#> FALSE
```

@schloerke does this warrant a quick look? It does seem like `is.numeric()` allows `double` and `integer` storage types while also excluding classes like `Date` that also store as `double`.

r2evans: I had looked at everything (I thought) *except* `POSIXt` and `Date`, glad it still works. :-)

schloerke: I like `is.numeric()`. I think that's what I was looking for earlier.

Thanks @rich-iannone and @r2evans

rich-iannone: Great! I’ll open another PR with this small fix a bit later.

PR #338, opened 2019-08-02
Fixes to `groups` arg of `cells_column_labels()`

rich-iannone:
This PR incorporates resolver functions in the `cells_column_labels()` function as it pertains to `groups` (column spanner labels). The `resolve_cells_column_spanners()` util function allows inputs to be character vectors, spanner names in `vars()`, and select helpers like `starts_with()`.

This example now works:

```r
tibble(
  letter = rep(c("A", "B"), each = 3),
  id = rep(1:3, 2),
  x = runif(6),
  y = runif(6)
) %>% 
  gather(var, value, x:y) %>% 
  unite(column, letter, var) %>% 
  spread(column, value) %>%
  gt() %>%
  cols_split_delim("_", everything()) %>%
  tab_footnote(
    footnote = "Source: the definitive guide to the alphabet",
    location = cells_column_labels(groups = everything())
  )
```

Fixes: https://github.com/rstudio/gt/issues/228.


rich-iannone:
@schloerke This needs quite a bit more work. There is general confusion in the code between column names (i.e., the original column names) and the labels applied to them. The biggest problems are in `create_columns_component_h()`. I will debug this and add several tests.

schloerke:
Good catch! Thanks @rich-iannone

rich-iannone:
This PR will focus on modifying the `boxh_df` object. To do:

- transpose the df
- add columns `key`, `type`, `spanner_label`, `hidden_px`
  - `type` can contain `stub`, `row_group`, `hidden`, `hidden_at_px`, `default`
- top-down ordering is left-to-right for presentation

This PR will remove the `col_labels` data attr.

rich-iannone:
This is handled in https://github.com/rstudio/gt/pull/355. Closing.

Issue #250, opened 2019-04-09
Can't use alpha argument in data_color()?


mfherman:
I'm having trouble adjusting the alpha of background cell colors when using `data_color()` in conjunction with `scales::col_numeric()`. The documentation of `data_color()` says:

>  If using a colorizing helper function for colors then this option is ignored (each of the colorizing helper functions has its own alpha argument).

But when I try to add an alpha argument to either `data_color()` or `col_numeric()` I don't see any transparency change. Reprex below with example table from `data_color()` documentation. (Apologies for not including the output tables with the code, but I guess GH doesn't like the html formatting...)

``` r
library(gt)
library(tidyverse)

# no alpha
countrypops %>%
  filter(country_name == "Mongolia") %>%
  select(-contains("code")) %>%
  tail(10) %>%
  gt() %>%
  data_color(
    columns = vars(population),
    colors = scales::col_numeric(
      palette = c(
        "red", "orange", "green", "blue"),
      domain = c(0.2E7, 0.4E7))
  )

# set alpha in col_numeric()
countrypops %>%
  filter(country_name == "Mongolia") %>%
  select(-contains("code")) %>%
  tail(10) %>%
  gt() %>%
  data_color(
    columns = vars(population),
    colors = scales::col_numeric(
      palette = c(
        "red", "orange", "green", "blue"),
      domain = c(0.2E7, 0.4E7),
      alpha = 0.1)
  )

# set alpha in data_color()
countrypops %>%
  filter(country_name == "Mongolia") %>%
  select(-contains("code")) %>%
  tail(10) %>%
  gt() %>%
  data_color(
    columns = vars(population),
    colors = scales::col_numeric(
      palette = c(
        "red", "orange", "green", "blue"),
      domain = c(0.2E7, 0.4E7)),
    alpha = 0.1
  )
```

MadhulikaTanuboddi:
This does not work well on Edge and IE11 browsers. The population column where the color is applied does not get the color. Check the attached snapshot.
<img width="1704" alt="Issue_Edge_gt_app_alpha_color" src="https://user-images.githubusercontent.com/5369655/66544601-76099900-eaed-11e9-8d6e-3d2742f11130.png">

PR #340, opened 2019-08-03
Incorporate alpha option to `cell_fill()`

rich-iannone:
This PR adds the `alpha` argument to the `cell_fill()` stylizing helper function. This also fixes the `data_color()` function, which had an `alpha` argument that wasn't implemented. Here are two examples that now work:

```r
sza %>%
  dplyr::filter(
    latitude == 20 & tst <= "1000") %>%
  dplyr::select(-latitude) %>%
  gt(rowname_col = "month") %>%
  fmt_missing(
    columns = TRUE,
    missing_text = ""
  ) %>%
  tab_style(
    style = list(
      cell_fill(color = "darkblue", alpha = 0.4),
      cell_text(color = "white")
    ),
    locations = list(
      cells_data(columns = everything()),
      cells_column_labels(columns = everything())
    )
  )
```

cell_fill_alpha

```r
countrypops %>%
  dplyr::filter(country_name == "Mongolia") %>%
  dplyr::select(-contains("code")) %>%
  tail(10) %>%
  gt() %>%
  tab_options(row.striping.include_table_body = FALSE) %>%
  data_color(
    columns = vars(population),
    colors = c("red", "orange", "green", "blue"),
    alpha = 0.5
  )
```

data_color_alpha

Fixes: https://github.com/rstudio/gt/issues/250.


R/data_color.R
```diff
@@ -191,6 +195,16 @@ data_color <- function(data,
...
 
+      # Get alpha as hexadecimal value
+      if (color %>% is_hex_col_w_alpha()) {
+        alpha_hex <- color %>% get_alpha_vec()
+      } else {
+        alpha_hex <- alpha %>% frac_alpha_to_hex()
+      }
+
+      # Combine hexadecimal color with corresponding alpha
+      color <- color %>% paste_right(alpha_hex)
```
schloerke: this will paste_right when alpha values exist.  `color` should only be updated when `!is_hex_col_w_alpha(color)`

R/helpers.R
```diff
@@ -799,18 +799,36 @@ cell_style_to_html.cell_text <- function(style) {
...
 #' @param color The fill color. If nothing is provided, then `"#D3D3D3"` (light
 #'   gray) will be used as a default.
+#' @param alpha The alpha transparency value for the `color` as single value in
+#'   the range of `0` (fully transparent) to `1` (fully opaque). If not provided
+#'   the fill color will be fully opaque.
 #'
 #' @family helper functions
 #' @export
-cell_fill <- function(color = NULL) {
+cell_fill <- function(color = NULL,
```
schloerke: set defaults to `#D3D3D3` and `1` like we did for another helper.

R/helpers.R
```diff
@@ -799,18 +799,36 @@ cell_style_to_html.cell_text <- function(style) {
...
+  color_hex <- color %>% get_hexcolor_vec()
+
+  if (color %>% is_hex_col_w_alpha()) {
+    alpha_hex <- color %>% get_alpha_vec()
+  } else {
+    alpha_hex <- alpha %>% frac_alpha_to_hex()
+  }
+
+  # Combine hexadecimal color with corresponding alpha
+  color_val <- color_hex %>% paste_right(alpha_hex)
```
schloerke: same issue as above.  make helper function of `color_with_default_alpha(color, alpha)`

R/helpers.R
rich-iannone: Done. Thanks!

R/helpers.R
rich-iannone: Great idea. I made the `normalize_color()` util function.

R/data_color.R
rich-iannone: Now fixed with use of the `normalize_color()` util function. Thanks!

R/data_color.R
```diff
@@ -413,18 +428,71 @@ get_alpha_vec <- function(colors) {
...
+
+#' With any input color, transform to hexadecimal with alpha
+#'
+#' @noRd
+normalize_color <- function(color,
+                            alpha) {
+
+  # Get hexadecimal representation of `color`; this strips
+  # any alpha value that may be included
+  color_hex <- color %>% get_hexcolor_vec()
```
schloerke: This should be something like...

```r
  if (!is_hex_col_w_alpha(color)) {
    alpha_hex <- alpha %>% frac_alpha_to_hex()
    color <- paste0(get_hexcolor_vec(color), frac_alpha_to_hex(alpha))
  }

  color
```

This should only update color if necessary (my thought)

schloerke: this should work with a vector of color vals

Issue #358, opened 2019-08-16
Top lines in cell borders are overridden by row lines


rich-iannone:
In a gt table rendered as HTML, the top line of a cell border (created using `tab_style()` -> `cell_borders()`) is not visible unless it is `2px` or larger. The table body row lines currently have a width of `"thin"`, which is `1px`.

Here is a reproducible example with screenshot, where `cell_borders()` without any arguments draws a `1px` box at each location:

```r
library(gt)
library(tidyverse)

exibble %>%
  select(num, date, datetime) %>%
  gt() %>%
  tab_style(
    style = cell_borders(),
    locations = cells_data(
      columns = vars(num),
      rows = num > 100 & num < 10000)
  )
```

<img width="431" alt="top-border-missing" src="https://user-images.githubusercontent.com/5612024/63184076-90128800-c024-11e9-81b6-cd5e978390bb.png">

The solution to this shouldn't have anything to do with increased specificity when drawing border lines (these are already done as inline `style`s) but, likely, to change the drawing order of the row lines. Currently row lines are part of the `gt_row` CSS class:

```scss
  .gt_row {
    padding: $row_padding; /* row.padding */
    margin: 10px;
    border-bottom-style: solid;
    border-bottom-width: thin;
    border-bottom-color: #D3D3D3;
    vertical-align: middle;
    overflow-x: hidden;
  }
```

My thought is that we could substitute all `border-bottom-*` rules for `border-top-*` rules and  avoid the line-drawing conflict between row lines and border lines altogether.

PR #359, opened 2019-08-16
Fix for invisible top borders in rendered HTML tables

rich-iannone:
This PR addresses the problem whereby gt tables, when rendered as HTML, have invisible top lines of cell borders (not visible unless `2px` or larger).

The problem is isolated to the CSS rules of the `gt_row` class:

```scss
  .gt_row {
    padding: $row_padding; /* row.padding */
    margin: 10px;
    border-bottom-style: solid;
    border-bottom-width: thin;
    border-bottom-color: #D3D3D3;
    vertical-align: middle;
    overflow-x: hidden;
  }
```

Substituting all `border-bottom-*` rules for `border-top-*` avoids the line-drawing conflict between row lines and border lines. That is the only change in this PR and, through several visual tests, the changes made here fix the display issue.

Fixes: https://github.com/rstudio/gt/issues/358.

### DIFF
--- a/inst/css/gt_styles_default.scss
+++ b/inst/css/gt_styles_default.scss
@@ -126,9 +126,9 @@
   .gt_row {
     padding: $row_padding; /* row.padding */
     margin: 10px;
-    border-bottom-style: solid;
-    border-bottom-width: thin;
-    border-bottom-color: #D3D3D3;
+    border-top-style: solid;
+    border-top-width: 1px;
+    border-top-color: #D3D3D3;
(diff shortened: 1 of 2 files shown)


schloerke:
```r
exibble %>%
  select(num, date, datetime) %>%
  gt() %>%
  tab_style(
    style = cell_borders(),
    locations = cells_data(
      columns = vars(num),
      rows = num > 100 & num < 10000)
  )
```
fixed-top-border

PR #366, opened 2019-09-07
remove reliance on Github version of sass, which is now updated on CRAN

davidkane9:


davidkane9:
This is my first pull request on a real project. Apologies if I messed it up!

PR #376, opened 2019-10-01
Update README.md

william3031:
Removed reference to CRAN to avoid confusion.


william3031:
Just ignore this one.

Issue #238, opened 2019-04-01
Create table themes


rich-iannone:
Create preset table themes and include several additional styling options (e.g., control over horizontal and vertical borders).

rich-iannone:
Here are some ideas/schematics:

<img width="1055" alt="table_theme_schematics" src="https://user-images.githubusercontent.com/5612024/55635094-e07ffc00-578d-11e9-9fcb-8ba58b93b7cb.png">

The chosen colors use a standard set of six. And there is grayscale. The top row (`S`) is used exclusively for striping rows.

<img width="779" alt="standard_hues" src="https://user-images.githubusercontent.com/5612024/55635099-e4ac1980-578d-11e9-8095-51bb2fa0d87e.png">

One thing that is lacking so far is more grayscale themes. Those are forthcoming.

Aside from these mockups for the overall look, the different line segments (e.g., right of stub, summary border top/bottom, etc.) and cell types need to be a bit more rigorously defined, and, named.

genepi10:
These look fantastic! When can we have them?

rich-iannone:
The above preset schematics need to be updated to incorporate the header, row groups, and the grand summary.

charlesatkinson2000:
I fully support having these themes. I feel they would provide an easy way to style a table nicely (which could then be further modified/adjusted with tab_options).

charlesatkinson2000:
Also, nice color choices here. These all look very nice.

jennarowe99:
Any update on whether the themes will be added to the package? I'm very interested in this.

PR #394, opened 2019-11-12
Fix several issues related to styling of an HTML table

### LINKED ISSUE #337: The `table.font.size` option has no effect on column labels
The `table.font.size` option of `tab_options()` has no effect on the column labels (this includes the spanner labels). The following example demonstrates this:

```r
sp500 %>%
  head() %>%
  gt() %>%
  tab_header(
    title = "S&P 500",
    subtitle = "5 rows"
  ) %>%
  tab_spanner(
    label = "prices",
    columns = vars(open, high, low, close)
  ) %>%
  fmt_date(
    columns = vars(date),
    date_style = 2
  ) %>%
  fmt_currency(
    currency = "USD"
  ) %>%
  fmt_number(
    columns = vars(volume),
    suffixing = TRUE
  ) %>%
  tab_options(
    table.font.size = px(10L)
  )
```

table_font_size

We'd want this option to apply to all text elements in the table so that we could reliably scale table content without aberrations.

### LINKED ISSUE #361: [feature request] Add option to remove header/footer border
Please briefly describe your problem or your ideas for enhancement in the template below.

Have a question? Please don't use this form but instead ask on [RStudio community](https://community.rstudio.com), on [Stack Overflow](https://stackoverflow.com), or on Twitter (using the `#rstats` hashtag).

When describing a bug, please include a minimal reproducible example (otherwise known as a [reprex](http://reprex.tidyverse.org/)).

When posting, delete these explanatory lines and submit only the text derived from the template below.

---

Some academic journals require that borders of table header and table footer (see below for an example taken from the intro page) be removed before submission, but currently I don't see any option to do this in `gt`.

屏幕截图_1

I think it'd be nice to  have such feature.

rich-iannone:
This will fix a few issues mainly related to table options (accessed through the `tab_options()`): #337 and #361.

Issue #337 is concerned with font sizing of the column labels. The overall table font size is `16px` and textual elements in most table locations have font sizes defined as percentages relative to the table font size. This is not the case for the column labels and for the row group labels. The change here is to make relative sizing consistent as a default. In this way, changing the `table_font_size` value will change font-size values for all text in the table (we can always explicitly provide a font-size in `px` for any of the text elements as well through `tab_options()`).  All of this will make it possible to package up a set of options for predetermined table sizes (e.g., `small`, `medium`, `large`) in a future change.

Other changes:

- added new options in `tab_options()` for vertical lines and to access options for all extant horizontal lines
- made consistent options in `tab_options` to be able to set a background color for each formal table location
- added text manipulation options (e.g., font-size, text-transform) for certain table locations
- the naming of the options within `tab_options()` is much more internally consistent
- roxygen doc `@param`s have been reorganized/rewritten
- minor corrections throughout package to use the refined argument names
- corrected existing tests and added a plethora of new tests

Future changes:

With these changes, we finally have the ability to supply `preset_*()` functions that provide a distinct table look (and work on https://github.com/rstudio/gt/issues/238 can finally start). Here are a few example screenshots with the base table and some presets (one that removes all lines from the table, another that produces a fully lined table one, and, one that colors the locations differently):

01_base_table

02_lineless

03_fully_lined

04_color_locations

Given that there are a lot of options available in `tab_options()` (and I think this is justified), some thought should be given to helper functions to divide/define the options into those by location or parameter (two TODOs added to function body). There should also be an expansion of the `opt_*()` family of macro-type functions (e.g., `opt_footnote_marks()` makes it easy to define footnote marks without using `tab_options()`). Some ideas here are `opt_striping()`, `opt_table_outline()`, `opt_table_size()`, etc.

Fixes #337 
Fixes #361


R/dt_options.R
```diff
@@ -19,72 +19,96 @@ dt_options_init <- function(data) {
...
     "table_id",                          FALSE,  "table",            "value",   NA_character_,
     "table_width",                        TRUE,  "table",            "px",      "auto",
-    "table_font_size",                    TRUE,  "table",            "px",      "16px",
-    "table_background_color",             TRUE,  "table",            "value",   "#FFFFFF",
     "table_margin_left",                  TRUE,  "table",            "px",      "auto",
     "table_margin_right",                 TRUE,  "table",            "px",      "auto",
+    "table_background_color",             TRUE,  "table",            "value",   "#FFFFFF",
+    "table_font_color",                   TRUE,  "table",            "value",   "#333333",
+    "table_font_color_light",             TRUE,  "table",            "value",   "#FFFFFF",
+    "table_font_size",                    TRUE,  "table",            "px",      "16px",
```
jcheng5: *Not something to be fixed in this PR,* but it feels a little arbitrary which parts of the table get to change which font attributes. Also I might quibble (or bikeshed!) with which options can be set, e.g. you can't set different amounts of vertical and horizontal padding.

R/dt_options.R
```diff
@@ -19,72 +19,96 @@ dt_options_init <- function(data) {
...
+    "footnotes_border_width",             TRUE,  "footnotes",        "px",      "2px",
+    "footnotes_border_color",             TRUE,  "footnotes",        "value",   "#D3D3D3",
+    "footnotes_sep",                     FALSE,  "footnotes",        "value",   "<br />",
+    "footnotes_marks" ,                  FALSE,  "footnotes",        "values",  "numbers",
+    "source_notes_padding",               TRUE,  "source_notes",     "px",      "4px",
+    "source_notes_background_color",      TRUE,  "source_notes",     "value",   NA_character_,
+    "source_notes_font_size",             TRUE,  "source_notes",     "px",      "90%",
+    "source_notes_border_style",          TRUE,  "source_notes",     "value",   "solid",
+    "source_notes_border_width",          TRUE,  "source_notes",     "px",      "2px",
+    "source_notes_border_color",          TRUE,  "source_notes",     "value",   "#D3D3D3",
```
jcheng5: [No action needed] I suspect that in the future, when we get more serious about RTF and LaTeX, we'll need to stop populating the table so completely at startup, so we can tell the difference between what options the user has explicitly chosen and which should have the default values; this will let us warn if the user has explicitly chosen options that cannot be supported in RTF/LaTeX.

R/dt_options.R
rich-iannone: Let's talk more about this. I'd love to have all possible options for HTML table to be settable. I don't really think I have a great strategy to make this easy (and not overwhelming). I basically made some editorial decisions here but they are highly opinionated (and yes, arbitrary).

Issue #351, opened 2019-08-13
gtsave() does not export colors for R versions 3.5.3 or superior


Pietro-DAmelio:
When exporting, cell colors are visible only in html but not in any other format.
I tried in 4 different laptop and only the one with R version 3.5.1 the colors are visible.
Colors are correctly represented in the viewer window of RStudio but not exported.

A work-around I found is to open the html in Chrome, save as pdf ticking the background colors in the advanced option. 

Session info

```
R version 3.5.3 (2019-03-11)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17134)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] gt_0.1.0        reshape2_1.4.3  dplyr_0.8.3     tidybayes_1.1.0 Cairo_1.5-10   
 [6] ggplot2_3.2.1   sjPlot_2.7.0    loo_2.1.0       brms_2.9.0      Rcpp_1.0.2  
```
```r
<insert reprex here>
```

rich-iannone:
Cell colors are currently only implemented in HTML tables (not LaTeX or RTF currently), if that's what you mean. Although I suspect this is an issue entirely with the HTML format. Would you be able to share some labeled screenshots here?

Pietro-DAmelio:
Thank you for your interest. The html works fine. .png and .pdf do not show the colors. To save the html as pdf from chrome I had to enable background color, maybe when webshot captures the image from the html the background is not activated?

Pietro-DAmelio:
![table_tot_fit_nat18](https://user-images.githubusercontent.com/42181672/63031617-8d912080-beb4-11e9-80df-5b5bcad8a4ce.png)

rich-iannone:
Thanks for the quick response. One more thing though, could you provide the code that generated the table? I'd like to test this on my system as well.

rich-iannone:
Or, if the data are sensitive and you can't post that here, could you come up with a similar example table that exhibits the same problem?

Pietro-DAmelio:
data are not yet published :) but I could post them here I think.
But maybe not necessary....
[...]

PR #405, opened 2019-11-19
Image output fix when using `gtsave()`

rich-iannone:
Issue https://github.com/rstudio/gt/issues/351 correctly notes that webshot images of tables no longer provide any color (see the issue for further context). 

To solve this, it's proposed that the **webshot2** package instead be used. With this minimal change in place, the resulting PNG screenshot looks as it should. Before and after images are provided here.

A potential issue is that **webshot2** isn't on CRAN. This is okay so long as **gt** isn't submitted to CRAN before **webshot2** is. As with **webshot** before, **webshot2** is in Suggests.

webshot

webshot2

Fixes https://github.com/rstudio/gt/issues/351


rich-iannone:
Closing this PR in favor of a forthcoming PR that addresses the normalization of color values.

bacinf12:
I am searching for some support exporting tables with background color to pdf. Has this been addressed yet? Last year I had to export over a hundred html file reports then import them into pdf then export them. It was fine for last year since the reports were only a single page. But this year each report is 2 pages. PLEASE help me. Even a thread to subscribe to that has up to date details on this issue would be helpful. A solution would be amazing :)

Issue #339, opened 2019-08-03
The `cells_stub()` location helper isn't working


rich-iannone:
The `cells_stub()` location helper no longer applies styles to the stub location. This code produces a table with a stub (left-most column) that is unstyled:

```r
sza %>%
  dplyr::filter(
    latitude == 20 & tst <= "1000") %>%
  dplyr::select(-latitude) %>%
  dplyr::filter(!is.na(sza)) %>%
  tidyr::spread(key = "tst", value = sza) %>%
  gt(rowname_col = "month") %>%
  fmt_missing(
    columns = TRUE,
    missing_text = ""
  ) %>%
  tab_style(
    style = list(
      cell_fill(color = "darkblue"),
      cell_text(color = "white")
    ),
    locations = cells_stub(rows = TRUE)
  )
```

<img width="739" alt="unstyled_stub" src="https://user-images.githubusercontent.com/5612024/62415273-f21ac880-b5f4-11e9-95ae-ecf8ba365392.png">

MadhulikaTanuboddi:
This does not work well on Edge and IE11 (Windows 10). Left most column (month) data is missing on these browsers. Check the attached snapshot.
<img width="1677" alt="Issue_Edge_gt_app_cells_stub" src="https://user-images.githubusercontent.com/5369655/66544445-01cef580-eaed-11e9-97e4-7386abee271f.png">

PR #406, opened 2019-11-20
Rework color generation to favor rgba values when `alpha < 1`

rich-iannone:
This PR modifies some of the utility functions that normalize colors and choose foreground text colors (for maximal contrast w.r.t. the background color). The change was made to effectively remove the use of `rrggbbaa` hex color values (which aren't compatible with older browsers, including the webkit implementation used in **webshot**) and either use `rrggbb` hex color values or rgba values when alpha is anything other than 1.

Here are some working examples of `gtsave()` with this change (previously, most background colors and cell striping wouldn't appear).

many_styles

many_options_tbl

tbl_alpha1

tbl_alpha05

Fixes https://github.com/rstudio/gt/issues/351.
Fixes https://github.com/rstudio/gt/issues/339.
Fixes https://github.com/rstudio/gt/issues/250.


R/data_color.R
```diff
@@ -439,6 +443,70 @@ get_alpha_vec <- function(colors) {
...
+
+        build_rgba_string(color = color, alpha_val = 1)
+
+      } else if (color %>% is_hex_col_w_alpha()) {
+
+        alpha_dec <-
+          (
+            color %>%
+              substring(8, 9) %>%
+              paste_left("0x") %>%
```
jcheng5: Don't need `paste_left`, just `strtoi(base = 16L)`

R/data_color.R
```diff
@@ -439,6 +443,70 @@ get_alpha_vec <- function(colors) {
...
+              strtoi()
+          ) / 256
```
jcheng5: 255

R/data_color.R
```diff
@@ -439,6 +443,70 @@ get_alpha_vec <- function(colors) {
...
+
+        build_rgba_string(color = color, alpha_val = alpha_dec)
+
+      } else if (tolower(color) %in% grDevices::colors()) {
+
+        build_rgba_string(color = tolower(color), alpha_val = 1)
+
+      } else {
+        NA_character_
```
jcheng5: Should we throw?

R/data_color.R
```diff
@@ -439,6 +443,70 @@ get_alpha_vec <- function(colors) {
...
+              strtoi()
+          ) / 256
+
+        build_rgba_string(color = color, alpha_val = alpha_dec)
+
+      } else if (tolower(color) %in% grDevices::colors()) {
+
+        build_rgba_string(color = tolower(color), alpha_val = 1)
+
+      } else {
```
jcheng5: Should `""` and/or `NA` be allowed?

R/data_color.R
```diff
@@ -492,27 +561,51 @@ normalize_color <- function(colors,
...
+          }
+          color
+        }
+      )
+
+  if (alpha != 1) {
+
+    normalized_colors <-
+      normalized_colors %>%
+      vapply(
```
jcheng5: Avoid vapply here

R/data_color.R
```diff
@@ -439,6 +443,70 @@ get_alpha_vec <- function(colors) {
...
+    unlist() %>%
+    as.numeric()
+
+  hex_color <-
+    grDevices::rgb(
+      red = channels[1]/255,
+      green = channels[2]/255,
+      blue = channels[3]/255
+    )
+}
```
rich-iannone: This doesn't return anything. I think I have to fix this as a first commit before the internal documentation.

R/data_color.R
rich-iannone: Thanks for this advice (first time using that function). This whole function has been removed.

R/data_color.R
rich-iannone: Whole function removed.

R/data_color.R
rich-iannone: I don't think so.

R/data_color.R
rich-iannone: Yes. Although this function is removed, there should be a validation function for any user-supplied color strings.

R/data_color.R
rich-iannone: Fixed in the renamed `rgba_to_hex()` function.

R/data_color.R
rich-iannone: Function is now removed.

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
-  if (length(colors) == 0) {
-    stop("One or more colors must be provided", call. = FALSE)
+  col_matrix_to_rgba <- function(color_matrix) {
+
+    color_matrix %>%
+      as.data.frame() %>%
+      dplyr::mutate(
+        rgba = paste0("rgba(", r, ",", g, ",", b, ",", round(alpha, 2), ")")
+      ) %>%
+      dplyr::pull(rgba)
```
jcheng5: Please don't use dplyr (or a data frame, for that matter) for this operation, you can do it faster and more elegantly with base R. (Should be a one liner)

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
   }
 
-  if (!(length(colors) == length(alpha) || length(alpha) == 1)) {
-    stop("The length of `alpha` must be the same length as `colors` or `1`",
-         call. = FALSE)
+  color_matrix <-
+    vapply(
```
jcheng5: This should be vectorized

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
+      colors,
+      FUN.VALUE = character(1),
+      USE.NAMES = FALSE,
+      FUN = function(color) {
+        if (is_rgba_col(color)) color <- rgba_to_hex(color = color)
+        color
+      }) %>%
+    farver::decode_colour(alpha = TRUE)
```
jcheng5: Is `farver::decode_colour` doing anything for us here? Would `col2rgb` suffice?

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
+
+  if (!is.null(alpha)) {
+    color_matrix[, 4] <- alpha
```
jcheng5: Just to be clear, the `alpha` argument to `process_colors` is used to *override* the alpha information in the colors themselves (if any)?

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
-        color <- paste0(get_hexcolor_vec(color), frac_alpha_to_hex(alpha))
-      }
+  colors <- rep(NA_character_, nrow(color_matrix))
 
-      color
-    }
-  )
-}
+  colors_no_alpha <- unname(color_matrix[, 4] == 1 & !is.na(color_matrix[, 1]))
+  colors_w_alpha  <- unname(color_matrix[, 4] != 1 & !is.na(color_matrix[, 1]))
```
jcheng5: What would `is.na(color_matrix[,1])` mean? The whole color is `NA`?

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
 
-#' Determining the best `light` and `dark` colors for contrast
-#'
-#' @noRd
-ideal_fgnd_color <- function(bgnd_colors,
-                             light = "#FFFFFFFF",
-                             dark = "#000000FF") {
+  colors[colors_no_alpha] <-
+    color_matrix[colors_no_alpha, , drop = FALSE] %>%
+    farver::encode_colour()
```
jcheng5: This could just be a sprintf() call (`%02X` to convert each color channel to hex)

tests/testthat/test-table_parts.R
```diff
@@ -461,60 +461,60 @@ test_that("a gt table contains custom styles at the correct locations", {
...
-    rvest::html_nodes("[style='background-color: #90EE90FF;']") %>%
+    rvest::html_nodes("[style='background-color: #90EE90;']") %>%
     rvest::html_text() %>%
     expect_equal("gear_carb_cyl")
 
   # Expect that the `Mazdas` row group label
   # cell has a red background and white text
   tbl_html %>%
-    rvest::html_nodes("[style='background-color: #FF0000FF; color: white;']") %>%
+    rvest::html_nodes("[style='background-color: #FF0000; color: white;']") %>%
```
jcheng5: I would like to see some of these tests exercise the codepaths that result in `rgba` colors as well.

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
-        toupper(sub(".*(..)$", "\\1", color))
-      } else if (tolower(color) %in% grDevices::colors()) {
-        "FF"
-      } else {
-        NA_character_
-      }
-    }
-  )
+is_rgba_col <- function(color) {
+  color %>% tidy_grepl("^rgba\\([0-9]+?,[0-9]+?,[0-9]+?,[0-9\\.]+?\\)$")
```
jcheng5: There can be whitespace after `(`, before `)`, and before or after `,`.

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
-}
-
-frac_alpha_to_hex <- function(alpha) {
-  grDevices::rgb(0, 0, 0, alpha) %>% substr(8, 9)
-}
-
-#' Extract a vector of alpha values for a vector of colors
+#' Is a string a color value in rgba() format?
+#'
+#' The input for this is a character vector of length 1. The output is a single
```
jcheng5:
```suggestion
#' The input for this is a character vector of any length. The output is a vector of
```

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
+#' logical value.
```
jcheng5:
```suggestion
#' logical values.
```

R/data_color.R
```diff
@@ -369,18 +371,18 @@ adjust_luminance <- function(colors,
...
 
-  # Obtain the alpha values
-  alpha <- get_alpha_vec(colors)
+  # Retain the alpha values
+  alpha <- rgb_mat[, "alpha"]
 
-  # Get a matrix of values in the Luv color space
-  luv_matrix <- grDevices::convertColor(rgb_matrix, "sRGB", "Luv")
+  # Get a matrix of values in the HCL color space
+  hcl_mat <- farver::convert_colour(colour = rgb_mat, from = "rgb", to = "hcl")
```
jcheng5: Oh, is this the real reason you want to take the farver dependency?

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
+#' #RRGGBB or #RRGGBBAA), or color strings in the rgba() string format
+#' (e.g., "`rgba(255,170,0,0.5)`"). Output is the same length vector as the
+#' input but it will contain a mixture of either #RRGGBB colors (if the input
+#' alpha value for a color is 1) or rgba() string format colors (if the input
+#' alpha value for a color is not 1).
 #'
 #' @noRd
-normalize_color <- function(colors,
-                            alpha) {
+process_color <- function(colors, alpha = NULL) {
```
jcheng5: I'd love a clearer function name than `process_color`. I think you can think of this function as taking colors in various specifications, and returning the best representation for HTML--maybe `color_for_web` or `color_for_html` or `color_for_css` or `html_color`?

R/data_color.R
```diff
@@ -391,131 +393,118 @@ adjust_luminance <- function(colors,
...
-  if (!(length(colors) == length(alpha) || length(alpha) == 1)) {
-    stop("The length of `alpha` must be the same length as `colors` or `1`",
-         call. = FALSE)
+  color_matrix <-
+    vapply(
+      colors,
+      FUN.VALUE = character(1),
+      USE.NAMES = FALSE,
+      FUN = function(color) {
+        if (is_rgba_col(color)) color <- rgba_to_hex(color = color)
```
jcheng5: Just curious: is there any way we'd end up with `rgba()` colors here, besides the user manually typing in `rgba()` colors? I'm wondering if we even want to allow that syntax, or require people to provide gt with `#RRGGBB` or `#RRGGBBAA` colors (and `#RGB` if you're feeling generous). If there is a good reason to support `rgba()` (like if you might have to run this function on its own output??) then do we also need to support `rgb()`?

R/data_color.R
rich-iannone: This is a major reason and also because farver reliability converts input hexadecimal-string colors and color names to a rgba color matrix

R/data_color.R
rich-iannone: Yes, that's exactly right. I have to see whether this works in practice. Let me get back to you on that.

R/data_color.R
rich-iannone: I'm partial to `html_color` and made that change. I guess I prefer it because it matches the intended output context.

R/data_color.R
rich-iannone: The user should never provide rgba()-style colors. The reason this was included here is because a downstream function `ideal_fgnd_color()` will take the converted color (and it may be in rgba()) but requires a #RRGGBB color. I've since removed this `vapply()` and will instead use a revised `rgba_to_hex()` directly inside `ideal_fgnd_color()`.

I think this was a case of too much coupling.

tests/testthat/test-table_parts.R
rich-iannone: I will add *a lot* of new tests.

R/data_color.R
rich-iannone: I've changed the regex to: `"^rgba\\(\\s*(?:[0-9]+?\\s*,\\s*){3}[0-9\\.]+?\\s*\\)$"`. Thanks!

R/data_color.R
rich-iannone: I've completely rewritten the docs for this internal function to clearly explain what how it works and what it's used for. It is vectorized.

R/data_color.R
rich-iannone: Completely rewritten and, yes, returns logical values.

R/data_color.R
rich-iannone: So true, I ended up using `paste0()`.

R/data_color.R
rich-iannone: The `vapply()` has been removed and we just have `decode_colour()` (vectorized) to get the `color_matrix`.

R/data_color.R
rich-iannone: Using the farver analogue is a bit nicer in that the returned matrix has colors given as rowwise values and also because the alpha values are given in the range 0-1. The output is (for me) easier to work with because it's closer to the final representation.

R/data_color.R
rich-iannone: Yes. I removed those `!is.na(...)` stmts because we definitely *don't* want missing colors (i.e., `NA` inputs). I also added a stop statement to catch any incoming `NA`s.

R/data_color.R
rich-iannone: Update on this, user-defined hexadecimal colors with alpha *weren't working at all* if a vector of color values was passed directly to `data_color()`s `colors` arg. To fix this and get the promised outcome:
- the `alpha` arg is now defaulting to `NULL`, any provided value would force a fixed alpha value when not using one of the **scales** color functions
- just inside the first `for` loop in `data_colors()` we explicitly call **scales** functions with the user-defined palette; for those `alpha = TRUE` was added to use the colors' alpha component if available
- the documentation for the `alpha` param has been updated to better explain how the value works, especially in light of the default changing from `1`

With manual tests, this now works. Unit tests will be added for all combinations of color specifications.

R/helpers.R
```diff
@@ -822,13 +822,14 @@ cell_style_to_html.cell_text <- function(style) {
 cell_fill <- function(color = "#D3D3D3",
                       alpha = 1) {
```
jcheng5: Consider making `alpha = NULL` the default, which will respect the `color`'s alpha (if present).

rich-iannone: This is a great suggestion. Thanks, I made the change and modified the surround documentation, and added some tests.

tests/testthat/test-table_parts.R
rich-iannone: Tests were added.

R/data_color.R
rich-iannone: Thanks for the conversation earlier. The farver dependency was removed entirely.

R/data_color.R
rich-iannone: This now uses the `grDevices::rgb()` function.

R/data_color.R
rich-iannone: Extensive unit tests added.

R/data_color.R
rich-iannone: farver is now no more.

R/data_color.R
```diff
@@ -471,13 +470,15 @@ rgba_to_hex <- function(colors) {
...
   colors_to_hex <-
-    farver::encode_colour(
-      colour = color_matrix,
-      from = "rgb",
-      alpha = alpha_vals
+    grDevices::rgb(
+      red = color_matrix[, "r"] / 255,
+      green = color_matrix[, "g"] / 255,
+      blue = color_matrix[, "b"] / 255,
+      alpha = alpha
```
jcheng5: maxColorValue = 255, and stop dividing red/green/blue by 255, and multiply alpha by 255

bacinf12:
I'm new to R and am coding reports that spit out gt tables in pdfs. I am not able to get the background color of the row_groups (coded via tab_options) to export. This page is linked as a solution but I'm having a hard time teasing out what the solution is. Any help is much appreciated.

Issue #419, opened 2019-11-25
While mixture use UTF-8 characters, something goes wrong


gaospecial:
Suppose I have a data.frame of name like this:

`name <- data.frame(first_name=c("春","Fred"),last_name=c("高","Bäckhed"))`

```r
library(gt)
Sys.setlocale(locale = "Chinese")
gt(name)
```

![name1](https://user-images.githubusercontent.com/107597/69506277-dbfc9380-0f68-11ea-9349-178a1b6002cd.png)

```r
Sys.setlocale(locale = "English")
gt(name)
```

![name2](https://user-images.githubusercontent.com/107597/69506293-ecad0980-0f68-11ea-9f83-f0abff0bd951.png)

The unicode last name “Bäckhed" only show correctlly in "English" locale, while chinese is wrong in that situation.

My sessionInfo().

```
R version 3.6.1 (2019-07-05)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 17763)

Matrix products: default

locale:
[1] LC_COLLATE=English_United States.1252  LC_CTYPE=English_United States.1252   
[3] LC_MONETARY=English_United States.1252 LC_NUMERIC=C                          
[5] LC_TIME=English_United States.1252    

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
 [1] rmarkdown_1.17     gt_0.1.0           cowplot_1.0.0      export_0.2.2      
 [5] knitr_1.26         RColorBrewer_1.1-2 mrgut_0.1.0        rvest_0.3.5       
 [9] xml2_1.2.2         lubridate_1.7.4    forcats_0.4.0      stringr_1.4.0     
[13] dplyr_0.8.3        purrr_0.3.3        readr_1.3.1        tidyr_1.0.0       
[17] tibble_2.1.3       ggplot2_3.2.1      tidyverse_1.2.1   

loaded via a namespace (and not attached):
 [1] httr_1.4.1              sass_0.1.2.1            jsonlite_1.6           
 [4] modelr_0.1.5            stargazer_5.2.2         shiny_1.4.0            
 [7] assertthat_0.2.1        highr_0.8               cellranger_1.1.0       
[10] yaml_2.2.0              gdtools_0.2.1           pillar_1.4.2           
[13] backports_1.1.5         lattice_0.20-38         glue_1.3.1             
[16] uuid_0.1-2              digest_0.6.22           manipulateWidget_0.10.0
[...]

PR #436, opened 2019-12-11
Implement fix for UTF-8

rich-iannone:
This change ensures that UTF-8 encoded characters don't get flattened to text during the building of table data. Change is focused on only the `migrate_unformatted_to_output()` internal function, where character values were passed through `format()` (now, only `numeric` vectors are).

Fixes: https://github.com/rstudio/gt/issues/419

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -968,6 +968,7 @@ validate_length_one <- function(x, name) {
 }
 
 # print8 <- function(x) {
+#   force(x)
 #
 #   old_ctype <- Sys.getlocale("LC_CTYPE")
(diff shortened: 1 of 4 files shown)


R/utils.R
```diff
@@ -966,3 +966,12 @@ validate_length_one <- function(x, name) {
          call. = FALSE)
   }
 }
+
+# print8 <- function(x) {
+#
```
jcheng5: This should do `force(x)` before changing the locale

jcheng5:
```suggestion
#  force(x)
```

Issue #426, opened 2019-11-27
update html-script, html-rmd and html-email examples under gt-examples


MadhulikaTanuboddi:
https://github.com/rstudio/gt/tree/master/tests/gt-examples

1. **01-html-script examples**: review and double check these are working
2. **02-html-rmd examples:** review and double check these are working
3. **03-html-email examples** - With the new changes in blastula, following needs to be done
    https://github.com/rstudio/gt/blob/master/tests/gt-examples/05-html-email/emailing_table.R
     a. Update the existing html-email example to work with `smtp_send`
     b. Provide a new gt example (rmd) to test blastula with RStudio connect

MadhulikaTanuboddi:
Upon running all 01-html-script examples, I found the following cosmetic and functionality issues

1. Cosmetic - html-16-table-options_everywhere.R -> https://trello.com/c/cpX0ImOC/105-pkggt-html-16-tableoptionseverywhere-example-does-not-work
2. Cosmetic: https://github.com/rstudio/gt/blob/master/tests/gt-examples/01-html-script/html-05-airquality.R. Notice that `fmt_missing()` mentiond `Ozone` column 2 times.
3. Functionality: html-13 -> https://github.com/rstudio/gt/issues/437
4. Functionality: html-14 -> https://github.com/rstudio/gt/issues/438

MadhulikaTanuboddi:
@rich-iannone : 05-html-email-example -> https://github.com/rstudio/gt/blob/master/tests/gt-examples/05-html-email/emailing_table.R needs following changes/updates

1. Update this blastula email example with gt to use the latest smtp_send function
2. Update this example to get the pizza plot and gt table working
3. This example is a bit comprehensive that its not easy to understand for beginners using blastula/gt. So, please add couple of simple examples to quickly get a context.
4. Add RStudio connect rmd example that includes gt tables and emails to be sent via blastula/connect.

MadhulikaTanuboddi:
@rich-iannone Another minor update needed. Remove the usage of `emo`  from https://github.com/rstudio/gt/blob/master/tests/gt-examples/05-html-email/emailing_table.R until https://github.com/rich-iannone/blastula/issues/137 is fixed.

Issue #421, opened 2019-11-27
Update README, function reference, and Cloud examples


rich-iannone:
Update README and pkgdown reference section (`_pkdown.yml`). 

For the RStudio Cloud examples, create multiple .Rmd files in a single Cloud project (splitting by family). This will reuse examples from the function documentation.

MadhulikaTanuboddi:
@rich-iannone : https://gt.rstudio.com/index.html. Two requests

#1. Can you provide links to the functions in the svg?
#2. Update the svg file with the all the new functions and correct function names? for ex: tab_stubhead_label is now tab_stubhead.

MadhulikaTanuboddi:
Update needed in this context - https://gt.rstudio.com/reference/summary_rows.html
`grand_summary_rows()` function reference uses the same link above and the example in there does not show how to use this function. Figures also needs to be updated showing grand summary rows

Issue #322, opened 2019-07-16
Rename helper function `cells_group()` to `cells_rowgroup()`


rich-iannone:
As discussed in https://github.com/rstudio/gt/issues/319, the helper function `cells_group()` should be renamed to `cells_rowgroup()`. The function name is longer but it makes clearer that the target is a group of rows (rather than a group of columns like in a spanner `group`).

This also works better with the documentation which often uses the term 'row groups'.

Issue #321, opened 2019-07-16
Rename helper function `cells_data()` to `cells_body()`


rich-iannone:
As discussed in https://github.com/rstudio/gt/issues/319, this name change will fit better with the table model especially since the term 'data' is somewhat ambiguous (since we can create new data in terms of summary rows).

william3031:
Is this what you're after or is there more to this? I imagine that it would need to be changed in other places too. Where? https://github.com/rstudio/gt/pull/377

rich-iannone:
@william3031 It would need to change everywhere (tests, vignettes, `_pkgdown.yml`, etc.) and the best way to ensure that the name change is performed everywhere is to use a text editor that can do a search and replace globally. Then `devtools::document()` needs to be executed so that the .Rd files get modified.

PR #441, opened 2019-12-14
Renaming of API functions; documentation pass

rich-iannone:
The changes in this PR are focused on changing names of some helper functions, improving the internal help files, updating the pkgdown site, and updating the README.

Fixes: https://github.com/rstudio/gt/issues/426
Fixes: https://github.com/rstudio/gt/issues/421
Fixes: https://github.com/rstudio/gt/issues/322
Fixes: https://github.com/rstudio/gt/issues/321


R/text_transform.R
```diff
@@ -1,17 +1,14 @@
...
-#'   footnote text. These helper functions are: [cells_title()],
-#'   [cells_column_labels()], [cells_group()], [cells_stub()], [cells_data()],
-#'   and [cells_summary()]. Please see the help article \link{location_cells}
-#'   for more information on how these helper functions can be used.
-#'   Additionally, we can enclose several `cells_*()` calls within a `list()` if
-#'   we wish to link the footnote text to different types of locations (e.g.,
-#'   cell data values, stub group headings, the table title, etc.).
+#'   transformation. Only the [cells_body()] helper function can be used here.
+#'   We can enclose several [cells_body()] calls within a `list()` if we wish to
+#'   make the transformation happen at different body cell locations.
```
jcheng5: Actually other cell types work as well

README.Rmd
```diff
@@ -21,17 +21,17 @@ knitr::opts_chunk$set(
...
+With the **gt** package, anyone can make wonderful-looking tables using the **R** programming language. The **gt** philosophy: we can construct a wide variety of useful tables with a cohesive set of table parts. These include the *table header*, the *stub*, the *column labels* and *spanner column labels*, the *table body*, and the *table footer*.
 
 <p align="center"><img src="man/figures/gt_parts_of_a_table.svg" width="800px"></p>
 
-It all begins with preprocessed **table data** (be it a tibble or a data frame). You decide how to compose your **gt table** with the elements you need for the task at hand. Output can either be in the form of **HTML**, **LaTeX**, or **RTF**. All work beautifully inside **R Markdown** documents.
+It all begins with preprocessed **table data** (be it a tibble or a data frame). You decide how to compose your **gt table** with the elements you need for the task at hand.
```
jcheng5:
```suggestion
It all begins with preprocessed **table data** (be it a tibble or a data frame). You decide how to compose your **gt table** with the elements you need for the task at hand. Currently, **gt** supports **HTML** output, with **LaTeX** and **RTF** planned for the future.
```

R/text_transform.R
rich-iannone: Now changed, thanks!

README.Rmd
rich-iannone: I made the change, reads much better now.

R/helpers.R
```diff
@@ -532,6 +532,26 @@ cells_row_groups <- function(groups = TRUE) {
...
 
+#' @noRd
+#' @export
+cells_group <- function(groups = TRUE) {
+
+  warning("The `cells_group()` function is deprecated and will soon be removed\n",
+          " * Use the `cells_row_groups()` function instead",
+          call. = FALSE)
+
+  # Capture expression for the `groups` argument
```
jcheng5: Could the body of this function just be the deprecation warning, and then `cells_row_groups({{groups}})`?

R/helpers.R
```diff
@@ -708,6 +728,32 @@ cells_body <- function(columns = TRUE,
...
+#' @noRd
+#' @export
+cells_data <- function(columns = TRUE,
+                       rows = TRUE) {
+
+  warning("The `cells_data()` function is deprecated and will soon be removed\n",
+          " * Use the `cells_body()` function instead",
+          call. = FALSE)
+
+  # Capture expressions for the `columns` and `rows` arguments
```
jcheng5: Same here, could it just be the warning and then `cells_data({{columns}}, {{rows}})`?

R/helpers.R
rich-iannone: This was the first thing I tried but it resulted in an error (probably due to the unquoting). I did use `!!` on the arg value but it still didn’t work out.

R/helpers.R
rich-iannone: I tried this with `!!` but not with the double curly brace operator. Let me see if that makes a difference.

rich-iannone: Whoa. With the `{{ }}` we get an infinite loop!

rich-iannone: I think we may just have to keep the exact code in both places. The good thing is that I don't expect `cells_body()` or `cell_row_groups()` to change in ~6 months time (when I might change the `warning()` to a `stop()`).

Issue #449, opened 2019-12-20
Have each function that operates on a gt table verify that the input is a gt table


rich-iannone:
It's very easy to do this:

```r
exibble %>% 
  tab_header(title = "A title")
```

or this:

```r
gtcars %>%
  dplyr::select(
    -mfr, -trim, bdy_style, drivetrain,
    -drivetrain, -trsmn, -ctry_origin
  ) %>%
  dplyr::slice(1:8) %>%
  tab_spanner(
    label = "performance",
    columns = vars(
      hp, hp_rpm, trq, trq_rpm,
      mpg_c, mpg_h)
  )
```

The thing that's missing, and is easy to forget, is the `gt()` call. When it's missing you either get a strange looking table returned, or `Error: No tidyselect variables were registered`, or, worse!

Anyway, I think a good idea is to validate that the incoming data object is indeed a `gt_tbl` and provide a suitable message (with advice) for when we inevitably forget.

Issue #444, opened 2019-12-16
fmt_number reference article example 2 does not work


MadhulikaTanuboddi:
Run the `tab_2` format number example from https://gt.rstudio.com/reference/fmt_number.html

I see the console error as in the attached snapshot and `suffixing` is not applied to the columns

<brief description of the problem>

```r
library(tidyr)
library(gt)
# Use `countrypops` to create a gt
# table; format all columns to use
# large-number suffixing
tab_2 <-
  countrypops %>%
  dplyr::select(
    country_code_3, year, population) %>%
  dplyr::filter(
    country_code_3 %in% c(
      "CHN", "IND", "USA", "PAK", "IDN")
  ) %>%
  dplyr::filter(year > 1975 & year %% 5 == 0) %>%
  tidyr::spread(year, population) %>%
  dplyr::arrange(desc(`2015`)) %>%
  gt(rowname_col = "country_code_3") %>%
  fmt_number(
    columns = TRUE,
    decimals = 2,
    suffixing = TRUE
  )
tab_2
```

![fmt_number_reference_example_tab_2_console_error](https://user-images.githubusercontent.com/5369655/70883138-e10c9b80-1f86-11ea-996a-9f385665381d.png)

MadhulikaTanuboddi:
[https://github.com/rstudio/gt/issues/446](url)

Issue #385, opened 2019-10-29
Attempting to format cells of an incompatible type leads to an unhelpful error message


rich-iannone:
Using `fmt_number()` on a column that is `character` will result in an unhelpful error message. This should be changed so that the message identifies the problem and states which columns/cells are incompatible.

MadhulikaTanuboddi:
Verify example below when this issue is fixed

```
sza %>%
  filter(latitude == 20) %>%
  filter(!is.na(sza)) %>%
  spread(key = "tst", value = sza) %>%
  gt(rowname_col = "month") %>%
  fmt_number(columns = TRUE, decimals = 2) %>%
  fmt_missing(columns = TRUE, missing_text = "")
```

Issue #205, opened 2019-03-03
Formatters do not check column types, leading to confusing outputs/messages


rich-iannone:
Currently, the `fmt_*()` functions do not check the types of columns being operated on. This can variously lead to errors, incorrect outputs, or surprising outputs. Here are a few examples:

Formatting a character column as numeric leads to an error (as it should, but we should provide our own error message).

```r
exibble %>%
  gt() %>%
  fmt_number(columns = vars(char))
```

```
>> Error in round(x[non_na_x], decimals) : 
>>  non-numeric argument to mathematical function 
```

Formatting a numeric column as markdown results in no change (it should result in an error).

```r
exibble %>%
  gt() %>%
  fmt_markdown(columns = vars(num))
```

Formatting a time column as a date leads to all values being `Thursday, January 1, 1970`.

```r
exibble %>%
  gt() %>%
  fmt_date(columns = vars(time))
```

And there are several more.

We just need to validate `x` and stop when the class doesn't match what's required for each formatter function.

PR #450, opened 2019-12-20
Perform input validation and check for types when formatting

rich-iannone:
This PR checks that the incoming data is a `gt_tbl` object (i.e., created by the `gt()` function). Column types are also checked in `fmt_*()` functions and `stop()`s are put into place for column types that won't be evaluated correctly.

Fixes: https://github.com/rstudio/gt/issues/444
Fixes: https://github.com/rstudio/gt/issues/449
Fixes: https://github.com/rstudio/gt/issues/385
Fixes: https://github.com/rstudio/gt/issues/205

### DIFF
--- a/R/modify_columns.R
+++ b/R/modify_columns.R
@@ -51,6 +51,10 @@ cols_align <- function(data,
                        align = c("auto", "left", "center", "right"),
                        columns = TRUE) {
 
+  # Perform input object validation
+  stop_if_not_gt(data = data)
+
+  # Get the internal data table
   data_tbl <- dt_data_get(data = data)
 
   # Get the `align` value, this stops the function if there is no match
@@ -160,6 +164,9 @@ cols_align <- function(data,
 cols_width <- function(data,
                        ...,
                        .list = list2(...)) {
+
+  # Perform input object validation
+  stop_if_not_gt(data = data)
 
   # Collect a named list of column widths
   widths_list <- .list
@@ -312,6 +319,9 @@ cols_label <- function(data,
                        ...,
                        .list = list2(...)) {
 
+  # Perform input object validation
+  stop_if_not_gt(data = data)
+
   # Collect a named list of column labels
   labels_list <- .list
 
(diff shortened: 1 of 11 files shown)


R/modify_columns.R
```diff
@@ -714,6 +733,9 @@ cols_move <- function(data,
 cols_hide <- function(data,
                       columns) {
 
+  # Perform input object validation
+  # stop_if_not_gt(data = data)
```
jcheng5: ?

R/utils.R
```diff
@@ -967,6 +968,16 @@ validate_length_one <- function(x, name) {
   }
 }
 
+column_classes_are_valid <- function(data, columns, valid_classes) {
```
jcheng5: I have some issues with this implementation.

1. Each column of the data frame could have multiple classes, e.g. `data.frame(x = structure(class = c("num", "numeric"), 1:10))`
2. I think each of the columns should have at least one match with `valid_classes`, like `any(class(x) %in% valid_classes)` or something

rich-iannone: Thanks. I made the change to match your suggestion.

R/modify_columns.R
rich-iannone: With https://github.com/rstudio/gt/pull/447, this can be uncommented. (I realize now that this review should have waited until that PR is merged.)

This reason this check can't occur now is that `cols_hide()` is used once in the initialization of the object (before the object is given the `gt_tbl` class). With https://github.com/rstudio/gt/pull/447, that call disappears and the above can be uncommented and a test can be added for `cols_hide()`.

R/utils.R
```diff
@@ -967,6 +968,17 @@ validate_length_one <- function(x, name) {
...
+
+  dt_data_get(data = data) %>%
+    dplyr::select(resolve_vars(var_expr = !!columns, data = data)) %>%
```
jcheng5: Callers should use `!!`, and `!!columns` here should be `{{columns}}`

R/modify_columns.R
rich-iannone: Will uncomment this in https://github.com/rstudio/gt/pull/447 after merging.

Issue #446, opened 2019-12-18
Helper functions - Add examples for pct() and px()


MadhulikaTanuboddi:
Above 2 helper functions does not have examples in the reference articles.

They should be added.

1. https://gt.rstudio.com/reference/px.html
2. https://gt.rstudio.com/reference/pct.html

MadhulikaTanuboddi:
Related to https://github.com/rstudio/gt/issues/422

Issue #422, opened 2019-11-27
Include self contained examples for every function and helper functions


MadhulikaTanuboddi:
https://gt.rstudio.com/reference/cell_fill.html

for example: cell_fill helper function above does not have an example and expected output.

kmatosli:
When running any of the sample code, I get an error indicating that could not find function: cell_fill and cell_text.

rich-iannone:
@kmatosli try updating gt with `remotes::install_github("rstudio/gt")`.

PR #451, opened 2019-12-20
Expand documentation such that all functions get at least one self-contained example

rich-iannone:
This expands on the documentation already available. Includes examples, SVGs, `@return` values, and more, for those functions previously missing those documentation features.

Fixes: https://github.com/rstudio/gt/issues/446
Fixes: https://github.com/rstudio/gt/issues/422

### DIFF
--- a/R/helpers.R
+++ b/R/helpers.R
@@ -1,10 +1,15 @@
 #' Interpret input text as Markdown-formatted text
 #'
-#' @param text The text that is understood to contain Markdown formatting.
-#' @return A character object that is tagged for a Markdown-to-HTML
-#'   transformation.
+#' Markdown! It's a wonderful thing. We can use it in certain places (e.g.,
+#' footnotes, source notes, the table title, etc.) and expect it to render to
+#' HTML as Markdown does. There is the [html()] helper that allows you to ferry
+#' in HTML but this function `md()`... it's almost like a two-for-one deal (you
+#' get to use Markdown plus any HTML fragments *at the same time*).
 #'
-#' @return A character object of class `from_markdown`.
+#' @param text The text that is understood to contain Markdown formatting.
+#'
+#' @return A character object of class `from_markdown`. It's tagged as being
+#'   Markdown text and it will undergo conversion to HTML.
 #'
 #' @examples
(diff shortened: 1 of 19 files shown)


R/helpers.R
```diff
@@ -1,10 +1,15 @@
 #' Interpret input text as Markdown-formatted text
 #'
+#' Markdown! It's a wonderful thing. We can use it in certain places (e.g.,
+#' footnotes, source notes, the table title, etc.) and expect it to render to
+#' HTML as Markdown does. There is the [html()] helper that allows you to ferry
+#' in HTML but this function `md()`... it's almost like a two-for-one deal (you
+#' get to use Markdown plus any HTML fragments *at the same time*).
```
jcheng5: Probably should add a note that it's not safe to put untrusted user input in here (like database values that can be influenced by a potentially malicious user--even a username!--or literally anything from Shiny `input`) as it'd then be possible to inject JavaScript into the end result document.

Also, when you support RTF/LaTeX, these docs will need to change to talk about output formats generically instead of HTML.

R/helpers.R
```diff
@@ -38,12 +43,17 @@ md <- function(text) {
 
 #' Interpret input text as HTML-formatted text
 #'
+#' For certain pieces of text (like in column labels or table headings) we may
+#' want to express them as raw HTML. In fact, with HTML, anything goes so it can
+#' be much more than just text. The `html()` function will guard the input HTML
+#' against escaping, so, your HTML tags will come through as HTML when
+#' rendered... to HTML.
```
jcheng5: Probably should add a note that it's not safe to put untrusted user input in here (like database values that can be influenced by a potentially malicious user--even a username!--or literally anything from Shiny `input`) as it'd then be possible to inject JavaScript into the end result document.

R/helpers.R
rich-iannone: I going to make this separate advisory happen in a separate PR. I believe it's important and I want to get the language right to ensure there's no misunderstanding.

R/helpers.R
rich-iannone: Will file a separate issue for this change.

PR #453, opened 2019-12-21
Fix #397: error with suffixing=TRUE with abs(x) < 1

### LINKED ISSUE #397: Error in values[index] (only 0's may be mixed with negative subscripts) when one variable in df contains only zeroes
First of all, let me say congrats for this awseome package gt you are building! :-)

Issue detected: If you have a dataframe more than one variable, and one of them contains only empty values, you end up hitting this error: 

`Error in values[index] : only 0's may be mixed with negative subscripts` 
(using R 3.6.x under GNU/Linux 64 bits, in case it matters)

Reproducible example:

```r
  data.frame(x=rnorm(10), y=rep(0,10)) %>% 
  gt() %>%
  fmt_number(
    columns = vars(x,
                   y),
    decimals = 2,
    suffixing = TRUE
  )
```

FYI, If you attempt to make the similar reprex but only with one  column (y in the previous example), then no error is triggered and the expected result is produced.

```r
  data.frame(y=rep(0,10)) %>% 
  gt() %>%
  fmt_number(
    columns = vars(y),
    decimals = 2,
    suffixing = TRUE
  )
```

jcheng5:
The `i` variable is used to index into the `suffixes`
vector, and previously, it could go negative because
log(n) is negative when 0 < n < 1.

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -472,6 +472,7 @@ num_suffix <- function(x,
   # index here
   i <- floor(log(abs(x), base = base))
   i <- pmin(i, length(suffixes))
+  i <- pmax(i, 0)
 
   # Replace any -Inf, Inf, or zero values
   # with NA (required for the `non_na_index()`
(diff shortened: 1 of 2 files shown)


R/utils.R
```diff
@@ -472,6 +472,7 @@ num_suffix <- function(x,
   # index here
   i <- floor(log(abs(x), base = base))
   i <- pmin(i, length(suffixes))
+  i <- pmax(i, 0)
```
rich-iannone: Makes sense!

Issue #459, opened 2020-01-03
gt table styles not rendered in the blastula email when sent via Connect


MadhulikaTanuboddi:
When we publish and email a rmarkdown document (that has a gt table) via connect, styling is not rendered at all in the received email (as shown in the attached snapshot)

Note: Connect preview shows the table as expected. The issue occurs only when viewed in the email.

I tried on gmail client only.

Steps to repro:
1. Have https://github.com/rstudio/gt/blob/master/tests/gt-examples/05-html-email/connect-example-main.Rmd and https://github.com/rstudio/gt/blob/master/tests/gt-examples/05-html-email/connect-example-email.Rmd available in your IDE
2. Deploy and publish "main.Rmd" file to the connect server you use (I used beta.rstudioconnect.com)
3. Preview the main file in connect dashboard (gt table shown correctly with borders and styles) - my example https://beta.rstudioconnect.com/connect/#/apps/7890/access/1898
4. Email the rmarkdown to few users/colloborators
5. Check the gt table in the email received

You will notice that gt styles are not shown at all in the email.
<img width="1279" alt="gt_table_connect_email_no borders shown" src="https://user-images.githubusercontent.com/5369655/71753036-2d385680-2e36-11ea-94b1-52abaa6c476d.png">

MadhulikaTanuboddi:
@rich-iannone 's PR to fix this : https://github.com/rstudio/gt/pull/458

daranzolin:
FYI--I updated the package yesterday (to 0.1.0), and I now get the unstyled output above in all my regular Rmd reports. It renders correctly in the Viewer pane, but not knit in HTML.

rich-iannone:
@daranzolin Update to master once more. We just reverted a change which caused this issue but should be all fixed now.

daranzolin:
🙌 🙏 @rich-iannone works great! Thanks!

PR #462, opened 2020-01-09
Modify regex patterns to be less greedy during inline of CSS styles

rich-iannone:
This fix is isolated to the call of `as_raw_html(inline_css = TRUE)`, which is important for outputting HTML text of the table with inlined CSS styles for inclusion in HTML email message bodies.


MadhulikaTanuboddi:
Fixes https://github.com/rstudio/gt/issues/459

Issue #463, opened 2020-01-09
Bug: Detecting a subtitle when none present - fails to knit to latex. 


annacnev:
**Problem Description**
When using ```as_latex``` and knitr in RStudio, gt LaTeX table fails giving the following error:

>! LaTeX Error: There's no line here to end.

**Reproducible Example**
```
library(gt)
library(tidyverse)

summary_single_table <- tribble(~study, ~N, ~Observation,
                                1,      25,     0.555, 
                                2,      55,     0.777,
                                3,     109,     0.355)



gt_tab <- summary_single_table %>% 
          gt(rowname_col = "study") %>%
  tab_header(title = "Table S1: Summary of Clinical Study Data by Study")


gt_tab %>% as_latex()
```

The LaTeX is failing because gt is registering a subtitle that is not actually present. If I do:

```
g <- gt:::build_data(gt_tab, context='latex')
gt:::dt_heading_has_subtitle(data = g)

>>> [1] TRUE

gt:::dt_heading_get(data = g)

>>>$title
[1] "Table S1: Summary of Clinical Study Data by Study"

$subtitle
character(0)
```

The ```dt_heading_has_subtitle``` function is:

```
dt_heading_has_subtitle <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading) > 0 && !is.null(heading$subtitle)
}
```

If I change this to:

```
dt_heading_has_subtitle <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading$subtitle) > 0 && !is.null(heading$subtitle)
}
```

It accurately returns ```FALSE```  instead of ```TRUE``` and knitr does not fail to render the LaTeX table. 

Note the ```dt_heading_has_title``` should also probably be changed from:

```
dt_heading_has_title <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading) > 0 && !is.null(heading$title)
}
```
To:
```
dt_heading_has_title <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading$title) > 0 && !is.null(heading$title)
}
```

jonathan-g:
I'm running into this bug too. The pull request  #464 looks like a clean solution (a dozen lines of code in two .R files, plus a bunch of changes to generated .Rd files) and it looks like it passes all the tests, but it has been languishing since January.

Is there anything I can do to help move this forward?

annacnev:
[...]

PR #464, opened 2020-01-09
LaTeX Knitr Failure: `dt_heading_has_subtitle()` Detecting Subtitle Not Present- Issue #463

annacnev:
PR to resolve: [Issue 463](https://github.com/rstudio/gt/issues/463)

When using Knitr to render LaTeX table, ```dt_heading_has_subtitle()``` returns ```TRUE``` when subtitle is not present. If ```TRUE```  is returned when no subtitle is actually present, the ```create_heading_component()``` with ```context='latex'``` constructs a heading component that looks like this in LaTeX output:
```
\caption*{
\large TAB_HEADER \\ 
\small \\
} \\ 
```

The ```\small \\``` line gives the error:

> ! LaTeX Error: There's no line here to end.

Because small is a command that is being preformed on nothing so it returns nothing. The ```\\``` signifies newline when used in the ```longtable``` environment, but it can't be used if no line is actually present (i.e. there's nothing there).. so it is erroring

In the issue I posted, I proposed the solution of changing:

```
dt_heading_has_subtitle <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading) > 0 && !is.null(heading$subtitle)
}

```

To:

```
dt_heading_has_subtitle <- function(data) {

  heading <- dt_heading_get(data = data)

  length(heading$subtitle) > 0 && !is.null(heading$subtitle)
}
```

Because with a title and no subtitle present, ```dt_heading_get``` returns:
```
>>>$title
[1] "TAB_HEADER"

$subtitle
character(0)
```

With ``` length(heading) > 0 && !is.null(heading$subtitle)``` in the original function is returning ```True``` because it is counting the length of ```heading``` which is a list of length 2, not the subtitle component of the list- which is actually length zero. 

When I changed it to ```length(heading$subtitle) > 0 && !is.null(heading$subtitle)``` , two of the HTML tests failed (commit 7ac149c), so I  assumed the length of the list needed to be used for proper HTML rendering and the length of the subtitle component needed to be used for proper latex rendering.

In the most recent commit:

- I added a conditional and context variable to the ```dt_heading_has_subtitle```. 
    - With ```context=HTML```(default) it will preform with ``` length(heading) > 0 && !is.null(heading$subtitle)```. 
    - With ```context=latex``` it will preform with ```length(heading$subtitle) > 0 && !is.null(heading$subtitle)``` .

- I adjusted the ```create_heading_component()``` so that when it calls to ```dt_heading_has_subtitle``` under the latex context, it also passes ```context=latex```


rich-iannone:
We did a pretty major refactoring to the table header generation in https://github.com/rstudio/gt/pull/779 which covers what this PR has. Thanks for putting the work into this PR.

Issue #465, opened 2020-01-13
as_latex creates extra columns in summary_rows 


tbradley1013:
When a `gt` table with `summary_rows` is converted to LaTeX, additional columns are added to the summary rows section causing the alignment to be incorrect. Here is a minimal example (using the example from the `summary_rows` help page with only two columns getting the summary row):

````
---
title: "gt issue"
author: "Tyler Bradley"
date: "1/13/2020"
output:
  pdf_document: default
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)

library(gt)
library(tidyverse)
```

```{r}
sp500 %>%
    dplyr::filter(
        date >= "2015-01-05" &
            date <="2015-01-16"
    ) %>%
    dplyr::arrange(date) %>%
    dplyr::mutate(
        week = paste0(
            "W", strftime(date, format = "%V"))
    ) %>%
    dplyr::select(-adj_close, -volume) %>%
    gt(
        rowname_col = "date",
        groupname_col = "week"
    ) %>%
    summary_rows(
        groups = TRUE,
        columns = vars(open, high, low, close),
        fns = list(
            min = ~min(.),
            max = ~max(.),
            avg = ~mean(.)),
        formatter = fmt_number,
        use_seps = FALSE
    )

```
````

![image](https://user-images.githubusercontent.com/23705496/72260984-f276b100-35e1-11ea-89e0-00a4a565782c.png)

If you render to HTML or just run the code in the console than it is rendered correctly. If you use the `as_latex` function then it returns the LaTeX with the extra spaces. I have found some tables that add two extra columns rather than one as well. 

Any help with this is greatly appreciated!

**Comment by tbradley1013** [comment]
The problem appears to be [here in `create_summary_rows`](https://github.com/rstudio/gt/blob/c82a67d3d0217307a2abf8239ede09bc198513e4/R/utils_render_common.R#L411):
 ```  
 summary_df <-
      list_of_summaries$summary_df_display_list[[group]] %>%
      as.data.frame(stringsAsFactors = FALSE)
```

Running browser() on this lapply loop in that function the output for the group ending row of this chunk of code is:

```
# A tibble: 3 x 7
  rowname date  open      high      low       close     week 
  <chr>   <chr> <chr>     <chr>     <chr>     <chr>     <chr>
1 min     ---   $2005.55$ $2029.61$ $1992.44$ $2002.61$ ---  
[...]

PR #466, opened 2020-01-13
add a summary_vars argument to create_summary_rows

tbradley1013:
This PR would fix https://github.com/rstudio/gt/issues/465. 

Basically the underlying issue was the the `create_summary_rows` function was including the grouping and rowname columns in its creation of the latex summary row lines. This resulted in two extra columns being added somewhere in each summary row (one for group column and one for rowname column). 

The changes I made add an additional argument to `create_summary_rows` named `summary_vars` which is meant to be all of the column names in the table. To accomplish this in `create_body_component_l` I just used the `default_vars` variable before `::rowname` was added to it. I also renamed the `default_vars`-post `::rowname` addition to `body_vars` since it is being used for the body portion of the table. 

In the `create_summary_rows` function, I added a line that will only select for the column names that are passed into the `summary_vars` function. 

Using these changed functions, the example shown in the above example works as expected. 

Let me know if there is anything else I need to do to get this PR merged!

Thanks!

### DIFF
--- a/R/utils_render_common.R
+++ b/R/utils_render_common.R
@@ -413,6 +413,8 @@ create_summary_rows <- function(n_rows,
       list_of_summaries$summary_df_display_list[[group]] %>%
       as.data.frame(stringsAsFactors = FALSE)
 
+    if (!"rowname" %in% summary_vars) summary_vars <- c("rowname", summary_vars)
+
     summary_df <-
       summary_df[, summary_vars]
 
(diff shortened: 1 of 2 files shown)


jjennifer:
This problem has not been solved yet. Does anyone please have some news on this?

rich-iannone:
Thanks for the work on this PR. Recently this was fixed by another PR: https://github.com/rstudio/gt/pull/768.

Issue #474, opened 2020-01-21
Table does not render correct when rowname_col equals groupname_col


MadhulikaTanuboddi:
Looking at line 83 in https://github.com/rstudio/gt/blob/master/R/gt.R, the default rowname_col is "rowname"

For example, if you have a tibble where one of the column names itself is "rowname" and you used it for `groupname_col` argument, then you would notice that either table does not render correctly or you will see an error

Couple of examples below based on https://github.com/rstudio/gt/blob/master/tests/gt-examples/01-html-script/html-11-adding_a_stubhead_label.R and https://github.com/rstudio/gt/blob/master/tests/gt-examples/01-html-script/html-13-adding_footnotes.R

EXAMPLE 1
```r
library(gt)
library(tidyverse)

# Create a table that has a stubhead label
tbl <-
  dplyr::tribble(
    ~groups, ~rowname, ~value_1, ~value_2,
    "A",          1,      361.1,    260.1,
    "A",          2,      184.3,    84.4,
    "A",          3,      342.3,    126.3,
    "A",          4,      234.9,    37.1,
    "B",          1,      190.9,    832.5,
    "B",          2,      743.3,    281.2,
    "B",          3,      252.3,    732.5,
    "B",          4,      344.7,    281.2,
    "C",          1,      197.2,    818.0,
    "C",          2,      284.6,    394.4
  )

# Create a display table
sh_caption_tbl <-
  tbl %>%
  gt(groupname_col = "rowname") %>%
  tab_stubhead(label = "groups")

sh_caption_tbl

```

This would result in following warning message and table with wrong columns and values
<img width="1668" alt="html12-when_rowname_col equals groupname_col" src="https://user-images.githubusercontent.com/5369655/72836141-e8d3f580-3c40-11ea-93d6-9e1f3438b234.png">

EXAMPLE 2

```r
library(gt)
library(tidyverse)

# Create a table with footnotes in various cell types
tbl <-
  dplyr::tribble(
    ~date,        ~rowname,  ~value_1,  ~value_2,
    "2018-02-10", "1",       20.4,      361.1,
    "2018-02-10", "2",       10.9,      743.3,
    "2018-02-10", "3",       34.6,      344.7,
    "2018-02-10", "4",        8.3,      342.3,
    "2018-02-11", "5",       28.3,      234.9,
    "2018-02-11", "6",       75.5,      190.9,
    "2018-02-11", "7",       63.1,        2.3,
[...]

PR #481, opened 2020-02-04
Error with any common name supplied to `rowname_col` or `groupname_col`

rich-iannone:
This adds a `stop()` statement in the `gt()` function for the situation where any names supplied to `rowname_col` or `groupname_col` are shared amongst the two.

This will avoid any strange formatting of the table downstream. Furthermore, having columns specified in this way is probably not what the user wants anyhow (a clear error message provides the opportunity to correct the inputs for `gt()`).

Several **testthat** tests were added.

Fixes https://github.com/rstudio/gt/issues/474.

### DIFF
--- a/R/dt_stub_df.R
+++ b/R/dt_stub_df.R
@@ -28,7 +28,7 @@ dt_stub_df_init <- function(data,
   # If `rowname` is a column available in `data`,
   # place that column's data into `stub_df` and
   # remove it from `data`
-  if (rowname_col %in% colnames(data_tbl)) {
+  if (!is.null(rowname_col) && rowname_col %in% colnames(data_tbl)) {
 
     data <- data %>% dt_boxhead_set_stub(var = rowname_col)
 
--- a/R/gt.R
+++ b/R/gt.R
@@ -95,6 +95,16 @@ gt <- function(data,
     groupname_col <- NULL
   }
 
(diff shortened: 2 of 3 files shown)


R/gt.R
```diff
@@ -95,6 +95,16 @@ gt <- function(data,
     groupname_col <- NULL
   }
 
+  # Stop function if `rowname_col` and `groupname_col`
+  # have the same string values
+  if (!is.null(rowname_col) &&
+      !is.null(groupname_col) &&
+      any(rowname_col %in% groupname_col)) {
+
+    stop("Any values provided to `rowname_col` and `groupname_col` cannot be shared.",
```
jcheng5: "The value `{rowname_col}` appears in both `rowname_col` and `groupname_col`; these arguments cannot have values in common."

R/dt_stub_df.R
```diff
@@ -28,7 +28,7 @@ dt_stub_df_init <- function(data,
   # If `rowname` is a column available in `data`,
   # place that column's data into `stub_df` and
   # remove it from `data`
-  if (rowname_col %in% colnames(data_tbl)) {
+  if (!is.null(rowname_col) && rowname_col %in% colnames(data_tbl)) {
```
jcheng5: Needs a unit test?

rich-iannone: Just checked and this is tested.

Issue #476, opened 2020-01-22
R3.5.3: Windows: html-09-formatting_uncertainties.R example shows unicode character in the table


MadhulikaTanuboddi:
Run the https://github.com/rstudio/gt/blob/master/tests/gt-examples/01-html-script/html-09-formatting_uncertainties.R example on Windows with R3.5.3 version

The gt table shows a unicode character A in the table (check the attached snapshot)

Note: This example works fine on R3.6.2 version

```r
library(gt)
library(tidyverse)

# Create a table with values and uncertainties
tbl <-
  dplyr::tribble(
    ~value_1, ~uncertainty, ~value_2, ~uncertainty_2,
    0.352,    0.10,         0.32,     NA_real_,
    0.983,    0.13,         NA_real_, NA_real_,
    0.639,    NA_real_,     0.21,     0.10,
    NA_real_, 0.17,         0.74,     0.15
  )

# Create a display table with uncertainties
uncert_tbl <-
  gt(data = tbl) %>%
  cols_merge_uncert(
    col_val = vars(value_1),
    col_uncert = vars(uncertainty)
  ) %>%
  cols_merge_uncert(
    col_val = vars(value_2),
    col_uncert = vars(uncertainty_2)
  ) %>%
  fmt_number(
    columns = vars(value_1, value_2),
    decimals = 2
  ) %>%
  fmt_missing(columns = vars(value_1, value_2))

uncert_tbl
```

![Windows_output_formatting_uncertainities](https://user-images.githubusercontent.com/5369655/72923320-aa533f00-3d03-11ea-87de-e81dcb8d152e.png)

PR #483, opened 2020-02-04
Improve the `cols_merge_uncert()` function

rich-iannone:
This change adds the `sep` argument to the `cols_merge_uncert()` function, making its interface consistent with that of `cols_merge_range()`. This change also uses an existing underlying pattern to provide the correct mark for the output context. Finally, this fixes a bug that appears in rare circumstances (with table output to IE11, using **gt** with R 3.5.3).

Fixes: https://github.com/rstudio/gt/issues/476.


R/modify_columns.R
```diff
@@ -839,14 +840,13 @@ cols_hide <- function(data,
 cols_merge_uncert <- function(data,
                               col_val,
                               col_uncert,
+                              sep = " +/- ",
```
jcheng5: Allow `I(" +/- ")` to mean "use this literal value"

R/utils.R
```diff
@@ -810,12 +810,32 @@ split_scientific_notn <- function(x_str) {
 #' @noRd
 tidy_gsub <- function(x, pattern, replacement, fixed = FALSE) {
 
-  gsub(pattern, replacement, x, fixed = fixed)
+  if (!utf8_aware_sub) {
```
jcheng5: Please add unit tests

```r
identical(tidy_sub(".", ".", "\u00B1", fixed = TRUE), "\u00B1")
identical(tidy_gsub(".", ".", "\u00B1", fixed = TRUE), "\u00B1")
```

R/utils_formatters.R
```diff
@@ -306,7 +306,7 @@ context_missing_text <- function(missing_text,
                missing_text
```
jcheng5: It'd be nice if these patterns also respected `I()`

R/modify_columns.R
rich-iannone: Now allowed and explained in docs.

R/utils.R
rich-iannone: These tests were added (and tests for usage of `I()` were added as well).

Issue #494, opened 2020-02-26
A random table id shouldn't be stored in the `gt_tbl` object


rich-iannone:
The `id` value is used for isolating a gt table's CSS from the rest of the document. However, currently, this is stored as an option value inside the `_options` table. A problem arises when we are reusing a gt object multiple times in an RMarkdown doc for instance, and `tab_options()` are used. We effectively get the same 'random' id in multiple HTML tables, and thus styling is applied *across* tables. Here's an example .Rmd:

````
```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)
library(gt)
library(tidyverse)
```


```{r}
tab_1 <- 
  exibble %>%
  dplyr::select(
    -c(fctr, date, time, datetime)
  ) %>%
  gt(
    rowname_col = "row",
    groupname_col = "group"
  ) %>%
  tab_header(
    title = md("Data listing from **exibble**"),
    subtitle = md("`exibble` is an R dataset")
  ) %>%
  fmt_number(columns = vars(num)) %>%
  fmt_currency(columns = vars(currency)) %>%
  tab_footnote(
    footnote = "Using commas for separators.",
    locations = cells_body(
      columns = vars(num),
      rows = num > 1000)
  ) %>%
  tab_footnote(
    footnote = "Using commas for separators.",
    locations = cells_body(
      columns = vars(currency),
      rows = currency > 1000)
  ) %>%
  tab_footnote(
    footnote = "Alphabetical fruit.",
    locations = cells_column_labels(
      columns = vars(char))
  )
```

Modify the table's background color to be "lightcyan".

```{r}
tab_1 %>%
  tab_options(
    table.background.color = "lightcyan"
  )
```

Modify the table width to `100%` (which spans the entire content width area).

```{r}
tab_1 %>%
  tab_options(
    table.width = pct(100)
  )
```
````

<img width="796" alt="rmd_tables" src="https://user-images.githubusercontent.com/5612024/75367502-6d441480-588e-11ea-9aac-792ca00e8e2a.png">

We never see the table with a 'lightcyan' background, but if we reversed the order of the last two chunks then both tables would have that background color applied.

What should be done is to insert a `::random` keyword as a result of the default `random_id()` function. In the `compile_scss()` function, if `::random` is seen as the option value for `table_id`, then generate the random id. If anything else (because the user can manually enter an `id` value), preserve that id value.

PR #495, opened 2020-02-26
Produce a random ID at print time

rich-iannone:
This resolves the issue where a truly random ID isn't possible across different renders of a gt table (since the ID was stored and can be subsequently reused).

Here is a screenshot that shows the fixed Rmd output (see #494 for the problematic output):

rmd_table_fix

Fixes: #494


R/gt.R
```diff
@@ -84,7 +84,7 @@ gt <- function(data,
                groupname_col = dplyr::group_vars(data),
                rownames_to_stub = FALSE,
                auto_align = TRUE,
-               id = random_id(),
+               id = ":random:",
```
jcheng5: Change the default to `NULL`, and at render time, turn `NULL` into a random id.

jcheng5: And don't accept `NA` as an id.

rich-iannone: Now fixed.

Issue #486, opened 2020-02-10
Improve NA handling for cols_merge_range


jcheng5:
```r
tbl <-
  dplyr::tribble(
    ~value_1, ~uncertainty, ~value_2, ~uncertainty_2,
    0.352,    0.10,         0.32,     NA_real_,
    0.983,    0.13,         NA_real_, NA_real_,
    0.639,    NA_real_,     0.21,     0.10,
    NA_real_, 0.17,         0.74,     0.15
  )

tbl %>%
  gt() %>%
  cols_merge_range(
    col_begin = vars(value_1),
    col_end = vars(value_2)
  )
```

The current behavior is that if the `col_begin` value is `NA`, the merged output is `NA`; if the `col_end` is `NA`, the unmerged `col_begin` value is used.

A better behavior would be to always replace missing values with NA (or let user specify).

PR #497, opened 2020-03-03
Modify NA behavior for `cols_merge_range()`

rich-iannone:
This PR changes the way `cols_merge_range()` handles `NA` values in the two columns it uses as inputs. Before, it was using `cols_merge_uncert()`'s `NA` handling rules. The revision to `cols_merge_range()` now does this:

```
col_begin  col_end
------------------
value + value -> value -- value
NA + value -> NA -- value
value + NA -> value -- NA
NA + NA -> NA
```

The NA in case 2 can be replaced with
`tbl %>% fmt_missing(columns = vars(col_begin), missing_text = <replacement>)`

The NA in case 3 can be replaced with
`tbl %>% fmt_missing(columns = vars(col_end), missing_text = <replacement>)`

The NA in case 4 can be replaced with
`tbl %>% fmt_missing(columns = vars(col_begin), missing_text = <replacement>)`

In all cases with an `NA`, a single replacement with `fmt_missing()` is possible and encouraged since otherwise `"NA"` text will appear as part of the range (or by itself, as in case 4).

This can be experimented with using this code:

```r
library(tidyverse)
library(gt)

tbl <-
  dplyr::tribble(
    ~value_1, ~value_2,
    NA_real_, NA_real_,
    0.352,    0.32,    
    0.983,    NA_real_,
    0.639,    0.21,    
    NA_real_, 0.74,    
    NA_real_, NA_real_,
  )

tbl %>%
  gt() %>%
  cols_merge_range(
    col_begin = vars(value_1),
    col_end = vars(value_2)
  ) %>%
  fmt_missing(
    columns = vars(value_1, value_2),
    missing_text = "miss"
  )
```

Fixes: https://github.com/rstudio/gt/issues/486


R/utils_render_common.R
```diff
@@ -288,7 +288,9 @@ perform_col_merge <- function(data,
         )
     }
 
-    if (type == "merge_range") {
+    if (type %in% c("merge_range", "merge_uncert")) {
```
jcheng5: Rather than a conditional here, validate that `type` is a valid value at the beginning of the function

R/utils_render_common.R
```diff
@@ -288,7 +288,9 @@ perform_col_merge <- function(data,
...
+
+      data_tbl <- dt_data_get(data = data)
```
jcheng5: The repeated `tidy_gsub` below is a bit problematic, because one `tidy_gsub` operation may return substituted text that is matched on a following `tidy_gsub`; when intuitively it should all be done in "one pass". glue (as used above) doesn't have this problem, but does have the problem of arbitrary code execution.

Ideally we would have one utility function that does this kind of template expansion.

R/utils_render_common.R
```diff
@@ -319,24 +321,42 @@ perform_col_merge <- function(data,
...
 
-      body <-
-        body %>%
-        dplyr::mutate(
-          !!mutated_column_sym := dplyr::case_when(
-            dplyr::row_number() %in% no_na_rows ~ glue::glue(pattern) %>% as.character(),
-            TRUE ~ !!mutated_column_sym
+        not_all_na_rows <-
+          seq_along(body %>% dplyr::pull(!!mutated_column_sym)) %>%
+          base::setdiff(base::intersect(na_1_rows, na_2_rows))
```
jcheng5: `not_all_na_rows <- !na_1_rows | !na_2_rows`

R/utils_render_common.R
```diff
@@ -319,24 +321,42 @@ perform_col_merge <- function(data,
...
+        body <-
+          body %>%
+          dplyr::mutate(
+            !!mutated_column_sym := dplyr::case_when(
+              dplyr::row_number() %in% not_all_na_rows ~ glue::glue(pattern) %>% as.character(),
+              TRUE ~ !!mutated_column_sym
+            )
           )
-        )
+      }
```
jcheng5:
```r
body[not_all_na_rows, mutated_column] <- as.character(glue::glue_data(body[not_all_na_rows,], pattern, .envir = emptyenv()))
```

R/utils_render_common.R
```diff
@@ -319,24 +321,42 @@ perform_col_merge <- function(data,
...
+
+      if (type == "merge_uncert") {
+
+        no_na_rows <-
+          seq_along(body %>% dplyr::pull(!!mutated_column_sym)) %>%
+          base::setdiff(na_1_rows) %>%
+          base::setdiff(na_2_rows)
```
jcheng5:
```r
no_na_rows <- !na_1_rows & !na_2_rows
```

R/utils_render_common.R
```diff
@@ -319,24 +321,42 @@ perform_col_merge <- function(data,
...
+
+        body <-
+          body %>%
+          dplyr::mutate(
+            !!mutated_column_sym := dplyr::case_when(
+              dplyr::row_number() %in% no_na_rows ~ glue::glue(pattern) %>% as.character(),
+              TRUE ~ !!mutated_column_sym
+            )
+          )
```
jcheng5:
```r
body[no_na_rows, mutated_column] <- as.character(glue::glue_data(body[no_na_rows,], pattern, .envir = emptyenv()))
```

R/utils_render_common.R
rich-iannone: Now set to throw if `type` is invalid. An additional safeguard for checking `type` is now in the `dt_col_merge_entry()` function.

R/utils_render_common.R
rich-iannone: Now in a better place with use of our `glue_gt()` function:

```
glue_gt <- function(.x, ...) {
  glue::glue_data(.x, ..., .transformer = get, .envir = emptyenv())
}
```

There was a `glue_char()` function that just did `as.character()` after `glue::glue()`. It's no longer being used, so, it was removed.

R/utils_render_common.R
rich-iannone: Now fixed (thanks).

R/utils_render_common.R
rich-iannone: Now using `glue_gt()` which incorporates safe options.

R/utils_render_common.R
rich-iannone: This is fixed. Thanks again.

R/utils_render_common.R
rich-iannone: Now entirely reworked.

PR #500, opened 2020-03-05
Example image generation

rich-iannone:
This adds a `scripts` folder and an .R script to create super-optimized PNG files for all of the examples. It relies on having ImageMagick and `pngquant` available on the system but I think that's fair since this will be run by developers for the most part.

The other part of this is actually replacing the image files in `man/figures` and ensuring that the references to all figures are updated in the relevant function docs (tons of `.svg -> .png` changes).


scripts/generate_example_pngs.R
```diff
@@ -0,0 +1,191 @@
...
+
+  # Get the initial list of objects in the environment
+  obj_list_initial <- ls()
+
+  # Evaluate the examples in the examples for each topic/function; the
+  # objects we'd want to create images from all begin with `tab_`
+  utils::example(topic = fn_name, package = "gt", character.only = TRUE, give.lines = FALSE)
+
+  # Get a vector of new objects created
+  new_obj <- base::setdiff(ls(), obj_list_initial)
```
jcheng5: Use `local({ ... })` instead, to give yourself a clean sheet each time (rather than adding/removing stuff to the same environment for every iteration)

jcheng5: Or lapply

scripts/generate_example_pngs.R
```diff
@@ -0,0 +1,191 @@
...
+  for (i in seq_along(tab_obj)) {
+
+    # Create a systematic file name so that the help file
+    # can access multiple images in the correct order
+    filename <- file.path(temp_dir, paste0("man_", fn_name, "_", i, ".png"))
+
+    # Save the image with a high `zoom` value for clarity,
+    # but resized down to 50% of the original
+    save_with_webshot(
+      data = get(tab_obj[i]),
```
jcheng5: Maybe validate that `tab_obj[i]` is a gt object?

scripts/generate_example_pngs.R
```diff
@@ -0,0 +1,191 @@
...
+dir.create(temp_dir_final)
+png_list <- list.files(temp_dir, ".*\\.png$")
+file.copy(file.path(temp_dir, png_list), file.path(temp_dir_final, png_list))
+
+# Shrink PNG files with `webshot::shrink()`
+webshot::shrink(filename = file.path(temp_dir_final, png_list))
+
+# Use `pngquant` to losslessly compress PNG files
+# The project is at https://github.com/kornelski/pngquant
+system(glue::glue("cd {temp_dir_final}; pngquant --force --ext=.png --skip-if-larger --nofs --speed 1 *.png"))
```
jcheng5: This might not work if `temp_dir_final` has spaces in it; maybe do

```r
old_dir <- setwd(temp_dir_final)
on.exit(setwd(old_dir), add = TRUE)
system("pngquant ...")
```

scripts/generate_example_pngs.R
rich-iannone: Not possible it seems but the `for` loop approach is better now at least.

scripts/generate_example_pngs.R
rich-iannone: Now done before using **webshot**.

scripts/generate_example_pngs.R
rich-iannone: Used `shQuote()` for paths in all `system()` calls.

PR #518, opened 2020-03-31
Don't .Rbuildignore vignettes

rich-iannone:
While this was good for the CRAN submission, pkgdown won't build articles if the `vignettes` folder is in `.Rbuildignore` (it just hangs). This will allow for the pkgdown site to regenerate (and for the build to no longer be broken).

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -14,4 +14,3 @@
 ^CODE_OF_CONDUCT\.md$
 ^TODOS
 tests/gt-examples
-vignettes


R/export.R
```diff
@@ -41,7 +41,7 @@
 #'   function.
 #'
 #' @examples
-#' \donttest{
+#' \dontrun{
```
jcheng5, line 44:
```suggestion
#' \donttest{
```

PR #527, opened 2020-04-03
Add \dontrun{} around `gtsave()` examples

rich-iannone:
This PR wraps `\dontrun{}` around the examples in the `gtsave()` doc to avoid a R CMD check issue on Travis.

### DIFF
--- a/R/export.R
+++ b/R/export.R
@@ -41,6 +41,7 @@
 #'   function.
 #'
 #' @examples
+#' \dontrun{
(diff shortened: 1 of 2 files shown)


rich-iannone:
@jcheng5 using `wait = TRUE` causes `gtsave()` to fail when saving any image files (error message says argument doesn’t exist). Where’s the documentation on this argument? Couldn’t locate it in the webshot help docs.

PR #529, opened 2020-04-08
Add shadow DOM compatibility

jcheng5:
This rule sets the default font not only for the root html
element, but also for the host of the shadow DOM, if any.


jcheng5:
I think travis is failing because I branched off of a commit with the `wait=TRUE` problem

rich-iannone (dismissed):
This is great. The failed builds have entirely to do with a different issue (`gtsave()`) that we expect will be resolved soon.

rich-iannone (approved):
Approved once before and now up to date with `master`.

Issue #542, opened 2020-04-13
The `gtsave()` function does not work in the dev version of gt when saving an image


rich-iannone:
Using `wait = TRUE` inside of `gt_save_webshot()` causes `gtsave()` to fail when saving any image files (error message says argument doesn’t exist). This issue was caused by https://github.com/rstudio/gt/pull/527.

PR #544, opened 2020-04-14
Remove `wait = TRUE` from `webshot()` call

rich-iannone:
This removes `wait = TRUE` from an internal `webshot()` call.

Fixes: #542

### DIFF
--- a/R/export.R
+++ b/R/export.R
@@ -202,7 +202,6 @@ gt_save_webshot <- function(data,
       selector = "table",
       zoom = zoom,
       expand = expand,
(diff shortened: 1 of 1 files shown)


MadhulikaTanuboddi:
Ran a quick test below with by installing the dev version of `gt` (without this change) and noticed the error 
```
> many_styles_tbl %>% gtsave("many_styles_tbl.png")
Error in webshot::webshot(url = paste0("file:///", tempfile_), file = filename,  : 
  unused argument (wait = TRUE)
```

PR code change looks good.

Issue #535, opened 2020-04-09
add significant figures to number formatting


drolejoel:
Include significant figures as a built-in formatting style.

- The number of significant figures should be configurable as an argument
- The alignment should be left, right, centered or on the decimal point. The latter is very important for pharma reporting and is pretty much impossible to do currently. An example of decimal alignment is available in [Microsoft Word's decimal tabulator in tables](https://www.datanumen.com/blogs/line-numbers-decimal-points-word-tables/)

There is actually a good solution for significant figures on Stack Overflow  [Stack Overflow](https://stackoverflow.com/questions/3245862/format-numbers-to-significant-figures-nicely-in-r) which works nicely with both positive and negative numbers.

rich-iannone:
I previously thought the requirement for specifying significant digits would be useful, way back when writing the `fmt_number()`, but it never got implemented. That's okay, it can be added and I think it's necessary for `fmt_number()`,  `fmt_scientific()`, and `fmt_percent()` and I'll add the `sigfig` argument.

As for alignment on a decimal point, that is definitely needed and should be implemented (generalizing to any alignment on any character that's specified).

drolejoel:
Brilliant!  You guys are the best :-)

Joel
(908) 507-6321

> On Apr 13, 2020, at 11:49 AM, Richard Iannone <notifications@github.com> wrote:
> 

﻿
I previously thought the requirement for specifying significant digits would be useful, way back when writing the fmt_number(), but it never got implemented. That's okay, it can be added and I think it's necessary for fmt_number(), fmt_scientific(), and fmt_percent() and I'll add the sigfig argument.

As for alignment on a decimal point, that is definitely needed and should be implemented (generalizing to any alignment on any character that's specified).

PR #546, opened 2020-04-15
Add options for sig figs / inclusion of trailing dec marks

rich-iannone:
This adds the `n_sigfig` arg to the `fmt_number()` function and the `drop_trailing_dec_mark` arg to the `fmt_number()`, `fmt_percent()`, and `fmt_currency()` functions.

The significant figures method uses a combination of the `signif()` and `formatC()` functions to ensure compliance with modern significant digit display rules.

The option to retain a decimal mark (here, with `drop_trailing_dec_mark`) is often seen in scientific literature and in statistical summary tables. If `drop_trailing_dec_mark` is changed from the default value, it displays the otherwise implicit decimal mark (even if there are no decimal figures required for display). 

Fixes: https://github.com/rstudio/gt/issues/535


R/utils_formatters.R
```diff
@@ -225,23 +225,55 @@ scale_x_values <- function(x,
...
+  } else {
+    stop("The format provided isn't recognized.")
+  }
+
   x_str <-
     formatC(
       x = x,
+      format = format,
+      mode = mode,
       digits = decimals,
```
jcheng5, line 261: `digits = digits`

R/utils_formatters.R
```diff
@@ -225,23 +225,55 @@ scale_x_values <- function(x,
...
+    digits <- n_sigfig
+    flag <- "#"
+    drop0trailing <- FALSE
+  } else if (format == "f") {
+    mode <- "double"
+    digits <- decimals
+    flag <- ""
+    drop0trailing <- drop_trailing_zeros
+  } else if (format == "e") {
+    mode <- NULL
```
jcheng5, line 248: Should be `mode <- "double"`

R/utils_formatters.R
rich-iannone: Now fixed.

R/utils_formatters.R
rich-iannone: Changed now.

Issue #525, opened 2020-04-03
Using pct() in cols_width() causes columns to be dropped


gowerc:
Side note - This is an incredible package, thank you for developing it!

# Issue 

Using `pct()` in `cols_width()`  causes the column to be dropped and a warning message of :

> Warning message:
> In function_list[[i]](value) : NAs introduced by coercion

```r
library(gt)

iris %>% 
    gt() %>% 
    cols_width(
        matches("Sepal") ~ pct(150),
        vars(Species) ~ pct(150),
        everything() ~ px(100)
    ) 
```

The main use case is to be able to do some width-wise padding to space the data and the column labels out a bit. 

```
R version 3.6.3 (2020-02-29)
Platform: x86_64-apple-darwin15.6.0 (64-bit)
Running under: macOS Mojave 10.14.6

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/3.6/Resources/lib/libRlapack.dylib

locale:
[1] en_GB.UTF-8/en_GB.UTF-8/en_GB.UTF-8/C/en_GB.UTF-8/en_GB.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] gt_0.2.0.5   rJava_0.9-12

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.4        rstudioapi_0.11   magrittr_1.5      munsell_0.5.0     tidyselect_1.0.0  colorspace_1.4-1  R6_2.4.1         
 [8] rlang_0.4.5       fansi_0.4.1       stringr_1.4.0     dplyr_0.8.5       tools_3.6.3       grid_3.6.3        packrat_0.5.0    
[15] checkmate_2.0.0   gtable_0.3.0      cli_2.0.2         commonmark_1.7    htmltools_0.4.0   ellipsis_0.3.0    digest_0.6.25    
[22] assertthat_0.2.1  tibble_3.0.0      lifecycle_0.2.0   crayon_1.3.4      purrr_0.3.3       ggplot2_3.3.0     sass_0.2.0       
[29] vctrs_0.2.99.9006 glue_1.3.2        stringi_1.4.6     compiler_3.6.3    pillar_1.4.3      backports_1.1.5   scales_1.1.0     
[36] pkgconfig_2.0.3  
```

[...]

Issue #493, opened 2020-02-19
cols_width removes last column when table has rownames


csoderberg:
I really like the package, thanks for developing it! I'm trying to set the width of columns in a table with `col_widths`, but when the table has one column specificied as `rowname_col`, the last column in the table isn't shown. 

**works as expected**
```r
iris %>% 
  gt() %>% 
  cols_width(
    vars(Sepal.Width) ~ px(120),
    vars(Petal.Length) ~ px(120),
    vars(Petal.Width) ~ px(120),
    vars(Species) ~ px(120),
    everything() ~ px(250))
```
![Screenshot 2020-02-19 16 39 14](https://user-images.githubusercontent.com/7332030/74879057-61100280-5336-11ea-8a31-1281db6311bc.png)

**Species column is removed when `Sepal.Length` made into rownames**
```r
iris %>% 
  gt(rowname_col = 'Sepal.Length') %>% 
  cols_width(
    vars(Sepal.Width) ~ px(120),
    vars(Petal.Length) ~ px(120),
    vars(Petal.Width) ~ px(120),
    vars(Species) ~ px(120),
    everything() ~ px(250))
```
![Screenshot 2020-02-19 16 37 56](https://user-images.githubusercontent.com/7332030/74879018-4b024200-5336-11ea-98c9-2aa332bc033e.png)

rich-iannone:
Thanks so much for reporting this. This will get fixed soon!

Issue #460, opened 2020-01-07
Allow cols_width to be specified relative to each other


akgold:
Right now, widths in `col_width` must be specified as a number of pixels. I would love to be able to specify column widths relative to each other. For example, I'd love to be able to put in something like
```r
cols_width(gt, vars(double_wide_col) ~ 2, vars(half_wide_col) ~ 0.5, TRUE ~ 1)
``` 
or even having the `TRUE ~ 1` implied, so it'd be 
```r
cols_width(gt, vars(double_wide_col) ~ 2, vars(half_wide_col) ~ 0.5)
```
This would mainly be useful because I don't know the width of my table in px as I'm writing it, but I do know how relatively wide I want my columns.

rich-iannone:
This is a great idea and one that I would use quite a lot! I'd really like to spend some time enhancing the definition of column widths because I feel we are only 20% of the way there toward good usability.

PR #561, opened 2020-04-23
Fix issues with defining column widths in `cols_width()`

rich-iannone:
There are several issues with the `cols_width()` function. This PR fixes the bugs and allows for specification of widths in `px` and as relative values.

Fixes https://github.com/rstudio/gt/issues/493
Fixes https://github.com/rstudio/gt/issues/525
Fixes https://github.com/rstudio/gt/issues/460


R/utils_render_html.R
```diff
@@ -67,38 +67,51 @@ get_table_defs <- function(data) {
...
-        })
-      )
+  widths <-
+    boxh %>%
+    dplyr::filter(type %in% c("default", "stub")) %>%
+    dplyr::arrange(dplyr::desc(type)) %>% # This ensures that the `stub` is first
+    .$column_width %>%
+    unlist()
+
+  if (!all(grepl("^([0-9\\.]*?|[0-9\\.]*?px|[0-9\\.]*?%|\\*)$", widths))) {
```
jcheng5, line 87:
```suggestion
  if (!all(grepl("^([0-9\\.]*)(px|%)?$", widths))) {
```

R/utils_render_html.R
```diff
@@ -108,7 +112,7 @@ get_table_defs <- function(data) {
     table_style <- paste(table_style, paste0("width: ", table_width), sep = "; ")
   }
 
-  # Create the <colgroup> tag
+  # Create the `<colgroup>` tag
   table_colgroups <-
     htmltools::tags$colgroup(
       lapply(widths, function(width) {
```
jcheng5, line 118: Use `css()` so that width will be dropped if it's null or empty string

R/utils.R
```diff
@@ -1019,6 +1019,50 @@ validate_table_id <- function(id) {
   }
 }
 
+is_css_length <- function(x,
+                          must_have_unit = FALSE,
```
jcheng5, line 1023: Get rid of `must_have_unit`, by the time we get here we really need units

R/utils.R
```diff
@@ -1019,6 +1019,50 @@ validate_table_id <- function(id) {
   }
 }
 
+is_css_length <- function(x,
```
jcheng5, line 1022: Vectorize

R/utils_render_html.R
rich-iannone: Made this change. Thanks for the suggestion.

R/utils.R
rich-iannone: Created a single, new function called `validate_css_lengths()`.

tests/testthat/test-cols_width.R
```diff
@@ -40,6 +40,11 @@ selection_text <- function(html, selection) {
     rvest::html_text()
 }
 
+# Helper function to compare a contiguous set of HTML fragments with raw html
+html_fragment_within <- function(raw_html, ...) {
+  grepl(paste0(c(...), "[\\n\\s]*?", collapse = ""), raw_html, perl = TRUE)
```
jcheng5, line 45:
```suggestion
  grepl(paste0("\\Q", c(...), "\\E", "[\\n\\s]*?", collapse = ""), raw_html, perl = TRUE)
```

Issue #396, opened 2019-11-12
`cols_merge()` fails to preserve img src of a preceding `text_transform()` that uses `web_image()`


JakeHurlbut:
When using `cols_merge()` (the new re-worked version) to merge two columns **after** a `text_transform()` that uses `web_image()` has been applied to one of the two merging columns, the merged column from `cols_merge()` places the second column inside the src attribute of the img tag in the first column.

Here is the img tag generated by the reprex code below, note the end of the src attribute: `<img src="http://upload.wikimedia.org/wikipedia/commons/thumb/9/92/Audi-Logo_2016.svg/196px-Audi-Logo_2016.svg.png Audi" style="height:30px;">`

``` r
library(tidyverse)
library(gt)

audi = gtcars %>%
  filter(mfr == 'Audi') %>%
  mutate(mfr_logo = "http://upload.wikimedia.org/wikipedia/commons/thumb/9/92/Audi-Logo_2016.svg/196px-Audi-Logo_2016.svg.png") %>%
  select(mfr_logo, mfr, model)


gt_audi = audi %>%
  gt() %>%
  text_transform(
    locations = cells_data(vars(mfr_logo)),
    fn = web_image
  )

gt_audi
```

<div id="gvnoxxajdg" style="overflow-x:auto;overflow-y:auto;width:auto;height:auto;">

<table class="gt_table">

<tr>

<th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">

mfr\_logo

</th>

<th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">

mfr

</th>

<th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">

model

</th>

</tr>

<body class="gt_table_body">

<tr>

<td class="gt_row gt_left">

<img src="http://upload.wikimedia.org/wikipedia/commons/thumb/9/92/Audi-Logo_2016.svg/196px-Audi-Logo_2016.svg.png" style="height:30px;">

</td>

<td class="gt_row gt_left">

Audi

</td>

<td class="gt_row gt_left">

R8

</td>

</tr>

<tr>

<td class="gt_row gt_left gt_striped">

<img src="http://upload.wikimedia.org/wikipedia/commons/thumb/9/92/Audi-Logo_2016.svg/196px-Audi-Logo_2016.svg.png" style="height:30px;">

</td>

<td class="gt_row gt_left gt_striped">

Audi

[...]

PR #562, opened 2020-04-25
Reverse order of two build steps

rich-iannone:
In this, text transformation occur before performing column merging.

Fixes: https://github.com/rstudio/gt/issues/396


jcheng5:
I think `text_transform` makes equal sense before or after column merging is performed, it really depends on what the user is trying to do. That being the case, maybe let's not break backward compatibility unless we feel strongly that the current (post-merge) behavior is uncommon and have our hearts set on `text_transform` as the name for the pre-merge behavior.

Moving forward, we could either have a second function, or an argument to `text_transform` letting the user specify the build phase or whatever.

Issue #536, opened 2020-04-09
stubs are not striped even when `row.striping.include_stub = TRUE`


gergness:
The `row.striping.include.stub` option does not affect the table output. I expect the stubs on even number rows to be yellow below.

```r
library(gt)
exibble %>% 
  gt(rowname_col = "row") %>% 
  opt_row_striping() %>% 
  tab_options(row.striping.include_stub = TRUE, row.striping.background_color = "yellow")
```

<img width="485" alt="Screen Shot 2020-04-09 at 12 31 32 PM" src="https://user-images.githubusercontent.com/2104579/78923574-0eadb100-7a5e-11ea-9374-6ec5b473e345.png">

PR #537, opened 2020-04-09
stubs get striped behavior

gergness:
Fixes #536 

Code:
```r
library(gt)
exibble %>% 
  gt(rowname_col = "row") %>% 
  opt_row_striping() %>% 
  tab_options(row.striping.include_stub = TRUE, row.striping.background_color = "yellow")
```

Before:
Screen Shot 2020-04-09 at 12 34 58 PM

After:
Screen Shot 2020-04-09 at 12 35 58 PM

Thanks!

PR #564, opened 2020-04-27
Restore row striping option in stub cells (`row.striping.include_stub = TRUE`)

rich-iannone:
This simply moves the `.gt_striped` rule below others to increase precedence.

Thanks @gergness for creating the PR that prompted this one (https://github.com/rstudio/gt/pull/537). We ultimately decided to go with this change (which restores the earlier behavior) since it work better with HTML style inlining (using `as_raw_html()`).

Testable code from @gergness, where the yellow row stripes are also present in the stub cells (leftmost column).

```r
library(gt)

exibble %>% 
  gt(rowname_col = "row") %>% 
  opt_row_striping() %>% 
  tab_options(
    row.striping.include_stub = TRUE,
    row.striping.background_color = "yellow"
  )
```

striping_in_stub

Fixes: https://github.com/rstudio/gt/issues/536

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ BugReports: https://github.com/rstudio/gt/issues
 Encoding: UTF-8
 LazyData: true
 ByteCompile: true
-RoxygenNote: 7.1.0
+RoxygenNote: 7.1.0.9000
 Depends:
     R (>= 3.2.0)
 Imports: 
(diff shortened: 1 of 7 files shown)


markbneal:
If you need this working in a hurry (like I do), and its not yet in the released version, replace the `gt_styles_default.scss` file on your computer with the one on the branch that fixes this, from: https://github.com/rstudio/gt/blob/restore-row-striping-stub/inst/css/gt_styles_default.scss

PR #575, opened 2020-05-07
Remove test of scales behaviour

hadley:
You accidentally depended on a scales 1.1.0 bug (and in general,  it's a bad idea to test errors generated by other packages).

Is there any chance you could do a quick gt release with this fix so I can submit scales to CRAN in the near future?

### DIFF
--- a/tests/testthat/test-data_color.R
+++ b/tests/testthat/test-data_color.R
@@ -625,18 +625,6 @@ test_that("the correct color values are obtained when defining a palette", {
       )
   )
 
-  # Expect an error when providing an `NA` along
-  # with valid color names to `colors`
-  expect_error(
-    test_tbl %>%
-      gt() %>%
-      data_color(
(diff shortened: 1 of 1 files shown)


rich-iannone:
Thanks! Me and Joe will definitely review this during our next meeting. We have made a number of fixes for the next release of gt, so, this is great timing (and this fix will be included). However, the release probably won't happen until mid- to late-June because we have a very large backlog of PRs that have to go through our QA process (and that likely won't happen for a few weeks because there is Shiny testing happening first).

hadley:
I have to submit scales to CRAN with a deadline of May 12. If you're ok with gt failing on CRAN for a bit (it's only a broken unit test not a behaviour change), I can just go ahead and submit and say that you have a fix inflight.

jcheng5:
If CRAN seriously objects we can also do a patch release with just this change, QA has signed off on things like that before.

Issue #543, opened 2020-04-13
The `data_color()` function is slow


rich-iannone:
When using the `data_color()` function in an example (where a few dozen cells needed to be colored), I noticed sluggish performance (in the range of 6-8 seconds on my system). Here is a reproducible example demonstrating this:

```r
library(gt)
library(tidyverse)
library(glue)
library(omsvg)
library(eqn2svg)
library(paletteer)

# Define a formula for SZA (Solar Zenith Angle)
sza_formula <-
  "\\cos \\theta_s = \\sin \\alpha_s = \\sin \\Phi \\sin \\delta + \\cos \\Phi \\cos \\delta \\cos h" %>%
  eqn2svg::eqn2svg()

# Create a gt table based on the `sza` dataset
sza %>%
  dplyr::filter(latitude == 20) %>%
  dplyr::select(-latitude) %>%
  dplyr::filter(!is.na(sza)) %>%
  tidyr::spread(key = "tst", value = sza) %>%
  gt(rowname_col = "month") %>%
  fmt_missing(
    columns = TRUE,
    missing_text = ""
  ) %>%
  data_color(
    columns = TRUE,
    colors = scales::col_numeric(
      palette = paletteer::paletteer_d(
        palette = "ggsci::orange_material"
      ) %>% as.character(),
      domain = c(0, 90),
      na.color = "#FFFFFF"
    )
  ) %>%
  tab_stubhead(label = html("month<br>(20&deg;N)")) %>%
  tab_header(
    title = glue::glue("{omsvg::SVG_la('sun')} SZA {omsvg::SVG_la('moon')}") %>%
      md()
  ) %>%
  tab_source_note(
    glue::glue(
      "These calculations are made possible with the SZA formula: {sza_formula}.<br>",
      "More details in <a href=\"https://en.wikipedia.org/wiki/Solar_zenith_angle\">this Wikipedia article</a>.") %>%
      html()
  ) %>%
  tab_options(
    heading.title.font.size = px(24),
    heading.border.bottom.style = "none",
    column_labels.border.top.style = "none",
    column_labels.font.size = "smaller",
    table.font.size = "smaller",
    data_row.padding = px(4)
  ) %>%
  opt_align_table_header("left") %>%
  opt_all_caps()
``` 

The underlying implementation for `data_color()` needs to be refactored. We can use **profvis** for the profiling study.

vikram-rawat:
Please let me know if it is resolved.

PR #576, opened 2020-05-08
Refactor `data_color()` so that it executes faster

rich-iannone:
The `data_color()` function was seen to be extremely slow. Reason for this was a combination of missed opportunities for vectorization, overuse of loops, and longer than necessary routes to an otherwise simple addition of `_styles` rows.

Below is reproducible code that would otherwise take 5-8 s before the changes in this PR (it now takes about 0.15 seconds):

```r
library(gt)
library(tidyverse)
library(scales)
library(paletteer)

sza %>%
  dplyr::filter(latitude == 20) %>%
  dplyr::select(-latitude) %>%
  gt(rowname_col = "month") %>%
  fmt_missing(
    columns = TRUE,
    missing_text = ""
  ) %>%
  data_color(
    columns = TRUE,
    colors = scales::col_numeric(
      palette = paletteer::paletteer_d(
        palette = "ggsci::orange_material"
      ) %>% as.character(),
      domain = c(0, 90),
      na.color = "#FFFFFF"
    )
  )
```

Fixes: https://github.com/rstudio/gt/issues/543

### DIFF
--- a/R/data_color.R
+++ b/R/data_color.R
@@ -147,6 +147,15 @@ data_color <- function(data,
   columns <- rlang::enquo(columns)
   resolved_columns <- resolve_vars(var_expr = !!columns, data = data)
 
+  # Get the sequence of row indices for the table
+  rows <- seq_len(nrow(data_tbl))
+
+  data_color_styles_tbl <-
+    dplyr::tibble(
+      locname = character(0), grpname = character(0), colname = character(0),
+      locnum = numeric(0), rownum = integer(0), colnum = integer(0), styles = list()
+    )
+
   for (column in resolved_columns) {
 
     data_vals <- data_tbl[[column]]
@@ -155,12 +164,8 @@ data_color <- function(data,
 
       if (is.numeric(data_vals)) {
 
-        color_fn <-
-          scales::col_numeric(
-            palette = colors,
-            domain = data_vals,
-            alpha = TRUE
-          )
+        # Create a color function based on `scales::col_numeric()`
+        color_fn <- scales::col_numeric(palette = colors, domain = data_vals, alpha = TRUE)
 
       } else if (is.character(data_vals) || is.factor(data_vals)) {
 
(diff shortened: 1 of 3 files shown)


R/data_color.R
```diff
@@ -147,6 +147,15 @@ data_color <- function(data,
...
+
+  data_color_styles_tbl <-
+    dplyr::tibble(
+      locname = character(0), grpname = character(0), colname = character(0),
+      locnum = numeric(0), rownum = integer(0), colnum = integer(0), styles = list()
+    )
+
   for (column in resolved_columns) {
 
     data_vals <- data_tbl[[column]]
```
jcheng5, line 161: Should be `data_tbl[rows, column]`

R/data_color.R
```diff
@@ -147,6 +147,15 @@ data_color <- function(data,
   columns <- rlang::enquo(columns)
```
jcheng5, line 147: (Sorry I can't get the comment closer) Can we do a match.args style for apply_to?

jcheng5:
Let's take a dependency on htmltools >= 1.1.0 and remove the workaround for the scales bug.

R/data_color.R
rich-iannone: Because `data_tbl` is a tibble, the above would subset to a one-column tibble. Because we want of vector of values, and still want to incorporate `rows`, I went with `data_tbl[[column]][rows]`.

rich-iannone:
Turns out, we already have a min version requirement of `1.1.0` on `scales`, so, no change required in `DESCRIPTION`.

R/data_color.R
rich-iannone: Now added. Thanks.

rich-iannone:
Just confirmed that removing this code:

```
        if (length(colors) > 1) {
          nlvl <-
            if (is.factor(data_vals)) {
              nlevels(data_vals)
            } else {
              nlevels(factor(data_vals))
            }
          if (length(colors) > nlvl) {
            colors <- colors[seq_len(nlvl)]
          }
        }
```

does NOT result in non-interpolated colors with `scales 1.1.0` (unit test that checks the `colors` correction code fails).

Issue #589, opened 2020-05-19
Font weight not accepted when numeric


malcolmbarrett:
`cell_text()` seems not to accept numeric values for `weight` (although the docs say it can). I understand that this doesn't apply to every font, but it seems like it's not getting passed the validation of the argument to begin with.

``` r
library(gt)
cell_text(weight = 400)
#> Error: The provided `weight` value cannot be `400`; it can only be either of the following:
#>  * `"normal"`, `"bold"`, `"lighter"`, or `"bolder"`
```

<sup>Created on 2020-05-19 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

rich-iannone:
Thanks for reporting this. There used to be the option to include the `font-weight` numbers but it was removed. This **should be put back in**, especially since some work is being done to include web fonts, most of which have quite a few weights.

Issue #585, opened 2020-05-15
Set global font options (rather than specifying body, column, header, etc invdividually)


jthomasmock:
After a discussion with @rich-iannone I posted this issue.

`gt` has easy and rich support for custom fonts which I love, but to change fonts across the entire table I believe the only way so far is to do the following (and if you wanted to change things like the title you'd also need to specify further targets):

``` r
library(gt)
suppressPackageStartupMessages(library(dplyr))

ggplot2::mpg %>% 
  head() %>% 
  gt() %>% 
  gt::tab_style(
    style = list(
      cell_text(font = "Fira Mono")
    ),
    locations = list(
      cells_body(gt::everything()),
      cells_column_labels(gt::everything())
  
  ))
```
<sup>Created on 2020-05-15 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

![Screen Shot 2020-05-15 at 10 40 20 AM](https://user-images.githubusercontent.com/29187501/82071851-4979b900-969c-11ea-8876-7aa32566892a.png)

It'd be great if there were options to:
- Set a global/table level font
- Override sections optionally with the above formatting (cell body, columns, title, etc)
- Clearly understand WHERE fonts are allowed to be imported from IE:
  - System Installed fonts
  - Imported web fonts
  - Integration w/ font packages (`extrafont`, `systemfonts`, `gridtext`, etc)

Some ideas about supporting/loading fonts via:
- `systemfonts`
- `thematic`
- `gridtext`

Tagging @apreshill for her interest as well.

zrodriguez89:
On a related note, I have seen similar requests regarding setting other stylistic changes to a table. For example, if I want to easily change all of the borders to black instead of the default gray.

Issue #404, opened 2019-11-19
Allow selection of a font for the table (and maybe specific parts)


rmurrayed:
Right now, there is no choice in the font selection. It would be awesome if we could choose a webfont for the HTML table or at least monospace fonts in certain columns (for numbers).

rich-iannone:
This is partially addressed by using `tab_style()` and `cell_text()` but you are right in that you can’t just choose a font for the entire table. A webfont is not a bad idea for HTML output tables, and I think there could be other options available to make the table more self-contained. 

Thank you for this feature request.

kmatosli:
Rich, it might be end user error on my part but do you have a list of fonts that can be used with the table.  I have attempted to use tab_style and cell_text and neither actually change the font.

PR #591, opened 2020-05-20
Settable font options

rich-iannone:
This PR adds the ability to specify additional table fonts before the default set in the font family.  This is done globally with the `opt_font()` function, which either accepts a local font name, or, could use the `google_font()` function (for specifying a webfont hosted on the Google Fonts service).

There is also the new ability to specify a Google Font through `google_font()` in `cell_text()`'s `font` argument.

Here is some testable code:

```r
exibble %>%
  dplyr::select(char, time) %>%
  gt() %>%
  tab_style(
    style = cell_text(font = c("Comic Sans MS", "Menlo", default_fonts())),
    locations = cells_body(columns = vars(time))
  ) %>%
  opt_table_font(font = list(google_font(name = "Dancing Script"), "Menlo"))
```

Fixes: https://github.com/rstudio/gt/issues/595
Fixes: #585 
Fixes: #589 
Fixes: https://github.com/rstudio/gt/issues/598
Fixes: https://github.com/rstudio/gt/issues/404

cc @jthomasmock


R/compile_scss.R
```diff
@@ -9,6 +9,20 @@ compile_scss <- function(data, id = NULL) {
 
   has_id <- !is.null(id)
 
+  # Get the vector of fonts and transform to a `font-family` string
+  font_vec <- dt_options_get_value(data = data, option = "table_font_names")
+  font_family_attr <- as_css_font_family_attr(font_vec = font_vec)
+
+  # Get any font-related @import statements
+  font_imports <- dt_options_get_value(data = data, option = "table_font_imports")
```
jcheng5, line 17: Deduplicate font_imports

R/helpers.R
```diff
@@ -1758,3 +1785,41 @@ gt_latex_dependencies <- function() {
...
+      )
+    )
+
+  google_fonts <-
+    list(
+      name = name,
+      import_stmt = import_stmt
+    )
+
+  class(google_fonts) <- "google_fonts"
```
jcheng5, line 1822: Rather than `google_fonts` it'd be nice if this was something more generic, since it could be used for any combination of font name + CSS.

R/opts.R
```diff
@@ -497,6 +497,76 @@ opt_table_outline <- function(data,
...
+#'
+#' @return An object of class `gt_tbl`.
+#'
+#' @family Table Option Functions
+#' @section Function ID:
+#' 9-7
+#'
+#' @export
+opt_table_font <- function(data,
+                           font,
```
jcheng5, line 535: Should have a way to specify a list of fonts that includes a google font

R/compile_scss.R
rich-iannone: Now done. Thanks!

R/helpers.R
rich-iannone: Class renamed to `"font_css"`.

R/helpers.R
```diff
@@ -1758,3 +1767,89 @@ gt_latex_dependencies <- function() {
...
+#' @param name The complete name of a font available in Google Fonts.
+#'
+#' @return An object of class `google_fonts`.
+#'
+#' @family Helper Functions
+#' @section Function ID:
+#' 7-23
+#'
+#' @export
+google_fonts <- function(name) {
```
schloerke, line 1836: because this is a single name, I am leaning towards `google_font(name)` (with no `s`)

rich-iannone: Good idea, I'll make that change.

rich-iannone: Okay, change is made. Great suggestion.

jcheng5:
Add unexported function for appending to `initial_css`. This function would `paste(collapse="\n")` its CSS argument, and have an `allow_duplicates=TRUE/FALSE` parameter that determines whether it should be added if it is already `%in% initial_css`.

jcheng5:
Notes from in-person discussion.

1. Should this be generalized from an "include arbitrary CSS" feature to "prepend/append arbitrary htmltools HTML"?
2. Do we need a dedicated set of functions for composing CSS from R, and if so, should they live here or htmtools?
3. Do we support scss?
4. How can provided CSS use the id for table-instance-specific rules?

Issue #607, opened 2020-06-12
tab_spanner() and column misalignment


fullmetalomelette:
I noticed that when I using tab_spanner() before or after setting a custom alignment using cols_align(), there is a slight misalignment with one of the end columns:

In the code below for Example 1, I set leftmost column of Iris to left-align and the remaining columns to right-align. Looks good.

In Examples 2 and 3 however (I get the same result for both), I see that after introducing a tab spanner over the columns, the "Species" column label seems to now be center-align while the column data itself is right-align (as expected).

Am I doing something incorrectly here? How do I use tab_spanner() without messing up my column alignment.

(Example 1)
![image](https://user-images.githubusercontent.com/993804/84526551-62bd5780-ac92-11ea-92c1-5ffc6997f3fc.png)

(Examples 2 and 3)
![image](https://user-images.githubusercontent.com/993804/84526543-59cc8600-ac92-11ea-9ee1-5b266df76c0d.png)

```
library(tidyverse)
library(gt)

#Example 1
gt::gt(iris) %>%
  gt::cols_align(align = "left", columns = colnames(iris)[1]) %>%
  gt::cols_align(align = "right", columns = colnames(iris)[2:ncol(iris)])

#Example 2
gt::gt(iris) %>%
  gt::tab_spanner("tab spanner", 1:5, gather = TRUE) %>%
  gt::cols_align(align = "left", columns = colnames(iris)[1]) %>%
  gt::cols_align(align = "right", columns = colnames(iris)[2:ncol(iris)])

#Example 3
gt::gt(iris) %>%
  gt::cols_align(align = "left", columns = colnames(iris)[1]) %>%
  gt::cols_align(align = "right", columns = colnames(iris)[2:ncol(iris)]) %>%
  gt::tab_spanner("tab spanner", 1:5, gather = TRUE)
```

ddsjoberg:
I came to report the same 🕺 It seems all headers are centered after `tab_spanner()`.

rich-iannone:
Yeah. This is an idea I thought might work out well but really isn't great in practice. I'm going to remove that bit of code that forces the column labels to be aligned center. 

With that, `cols_align()` probably needs some options to set the alignment differently for column labels compared to cells below. (Maybe, we might include a `rows` option to select a portion of the body cells.)

ddsjoberg:
That sounds like a great solution! (Good to see you on the call earlier this week!)

rich-iannone:
[...]

PR #662, opened 2020-10-19
Spanner alignment correction

rich-iannone:
This PR fixes a longstanding issue where column labels subordinate to column spanner labels had their alignment forced to be "center". This resulted in a balanced spanner-label/column-label assembly but the misalignment between such column labels and their data cells resulted in a worse appearance (especially magnified as column widths were increased). Furthermore, the center alignment was at odds with what the community wanted and was contrary to advice from experts on the matter. Now there is no forced alignment of column labels.

Should you need the old behavior, `tab_style()` can be used along with `cell_text(align = "center")`.

This PR also changes to automatic alignment of integer values from "center" to "right". This was another oversight but is corrected now.

An example plot, using the `airquality` dataset is shown below:

``` r
gt(data = airquality) %>%
  cols_move_to_start(columns = vars(Month, Day)) %>%
  cols_label(Solar.R = html("Solar<br>Radiation")) %>%
  fmt_number(
    columns = vars(Wind),
    decimals = 2
  ) %>%
  tab_spanner(
    label = "Measurement Period",
    columns = vars(Month, Day)
  ) %>%
  fmt_missing(columns = vars(Ozone, Solar.R, Wind, Temp)) %>%
  tab_options(table.width = "100%")
```

airquality_gt

Fixes: https://github.com/rstudio/gt/issues/607


jcheng5 (changes requested):
Unit tests

jcheng5 (approved):

Issue #626, opened 2020-07-22
`tab_spanner_delim` with `gather = TRUE` not working


pdeffebach:
Here is an MWE

```
r$> df = tibble(A_X = c(1), B_X = c(2), A_Y = c(3), B_Y = c(4))                                          

r$> gt(df) %>% tab_spanner_delim(delim = "_", gather = TRUE) 
```
Gives me this:

![Screenshot from 2020-07-22 12-32-54](https://user-images.githubusercontent.com/23196228/88202952-938c9700-cc17-11ea-8366-9bb32733078a.png)

From the docstring, we `gather = TRUE` should re-arrange the columns. 

On version 0.2.1

rich-iannone:
Thank you for reporting this bug (which is pretty serious) and providing the MWE. This is highly prioritized so it'll be handled soon.

Issue #597, opened 2020-05-29
Unable to change font of the column spanner if it has footnote


kulbhushanchand:
Since column spanner with footnotes (or with markdown) cannot have IDs, they are overlooked by helper function and thus styling them is not possible.
```
tab_style(
    style = list(
      cell_text(font = "Stencil")
    ),
    locations = list(
      cells_column_spanners(spanners = "Commands")
 ))
```
In the code above the, column spanner has already a footnote attached. Thus the above styling does not works on that particular column. Even `everything()` selector also overlooks this column.

1. A journal asks for particular font in my paper and I am stuck at this point. Any ideas or workaround?  

2. Is there any way to change the default font of gt table to something else? This can be a good workaround for now

**Edit**: Just came to my notice, there is a pull request https://github.com/rstudio/gt/pull/591 for the same. Keeping this issue open to get any valuable feedback or workaround.

rich-iannone:
Sorry about this problem. That PR will partially address this but I do want to fix the underlying bug. Thanks for reporting!

luis-zapata-b:
Hi! I have the same problem. It was solved?

rich-iannone:
@luis-zapata-b This should be fixed once https://github.com/rstudio/gt/pull/665 is merged.

Issue #521, opened 2020-04-01
Styling spanner labels can be easier if spanners have id values


jenniferp96:
Hello, thank you very much for this package. I'm a fan of another one of your packages (pointblank) but I'm also using gt a lot these days. I have a feature request concerning spanners. Once they are made, they can be hard to style because cells_column_spanners doesn't use ids like cells_column_labels. Instead it uses labels, which can be unwieldy. 

Could you make it possible to set a spanner id along with the spanner label in tab_spanner? Then, we won't have this problem anymore (as long as the labels are made to be unique).

Thanks!

rich-iannone:
Sure, this sounds like something that should absolutely be done. Thanks for the detailed request, it's a good one!

kulbhushanchand:
Came here to file the same issue. 
I tried using footnote on spanner, but spanner cannot be referenced if they are written in markdown.

@rich-iannone @jenniferp96 Am I doing something wrong? or should I file a separate issue (if it is a bug)?

rich-iannone:
You are doing nothing wrong! This is gt's (my) fault because there is no easy key or id for column spanners that would make later access easy or even possible. What needs to be done (by me) is to allow the creation of a key along with the styled label. In that way, `tab_style()` would require the key and not the styled label. 

I believe that this is a critical requirement so this will be worked on quite soon.

kulbhushanchand:
Thanks for your assurance and for this awesome package. _Your commitment to this project is highlighting from your reply._ 👍

PR #665, opened 2020-10-23
Add spanner `id` as an argument to `tab_spanner()`

rich-iannone:
This PR allows for the setting of a spanner ID value, which makes it easier to access a spanner column label through `cells_column_spanners()` (when using `tab_style()` or `tab_footnote()]), it occurs through the `id` value (and not the `label`). If an `id` is not provided here, it will be generated from the label, stripping any tags and special characters, and limiting the generated string to 10 characters. 

Although a message will be provided in this case, it is advised that an explicit `id` value be set if planning to access this cell in a later function call. Also, when providing an `id`, it must be unique across all ID values set for column spanner labels.

Fixes: https://github.com/rstudio/gt/issues/521
Fixes: https://github.com/rstudio/gt/issues/626
Fixes: https://github.com/rstudio/gt/issues/597


jcheng5:
After discussing in person:

1. No restriction on id; just use the label as-is by default.
2. Don't attempt to automatically fix duplicate id's in `tab_spanner`; just stop with an error message informative enough that it's clear how they should proceed.
3. For `tab_spanner_delim`, automatically fix duplicate id's using TBD method (tibble::name_repair?) and if that was necessary then let user know what the resulting id's are. Maybe we should have a param that lets them turn the message off.
4. If we allow n-level spanner hierarchies I'd like to revisit this, I'd imagine uniqueness only mattering within a single scope (like filesystem hierarchies).

R/dt_spanners.R
```diff
@@ -19,21 +19,28 @@ dt_spanners_init <- function(data) {
...
 
-  data %>%
-    dt_spanners_get() %>%
+
+  dt_spanners_get(data = data) %>%
     dplyr::bind_rows(
       dplyr::tibble(
         vars = list(vars),
         spanner_label = list(spanner_label),
+        spanner_id = list(spanner_id),
```
jcheng5, line 43: Let's just do character() for this

R/tab_create_modify.R
```diff
@@ -54,6 +54,16 @@ tab_header <- function(data,
...
+#' @param id An optional, but strongly recommended, input for an ID. When
+#'   accessing a spanner column label through [cells_column_spanners()] (when
+#'   using [tab_style()] or [tab_footnote()]), it occurs through the `id` value
+#'   (and not the `label`). If an `id` is not provided here, it will be
+#'   generated from the label, stripping any tags and special characters, and
+#'   limiting the generated string to 10 characters. Although a message will be
+#'   provided in this case, it is advisable to set an explicit `id` value if you
+#'   plan to access this cell in a later function call. Also, when providing an
+#'   `id`, please ensure that it is unique across all ID values set for column
+#'   spanner labels.
```
jcheng5, line 66: Update for new implementation

R/utils.R
```diff
@@ -1230,6 +1229,18 @@ validate_style_in <- function(style_vals,
   }
 }
 
+check_spanner_id_unique <- function(data,
+                                    spanner_id) {
+
+  existing_ids <- dt_spanners_get_ids(data = data)
+
+  if (spanner_id %in% existing_ids) {
+    stop("The `id` provided (`\"", spanner_id, "\"`) is not unique:\n",
```
jcheng5, line 1238:
```suggestion
    stop("The spanner `id` provided (`\"", spanner_id, "\"`) is not unique:\n",
```

jcheng5 (changes requested):
Just the nits from the comments

R/dt_spanners.R
rich-iannone: This is now changed and many internal changes were made to ensure that the IDs were used instead of the labels.

R/tab_create_modify.R
```diff
@@ -91,25 +101,42 @@ tab_header <- function(data,
 #' @export
 tab_spanner <- function(data,
                         label,
-                        columns,
+                        id = label,
+                        columns = NULL,
```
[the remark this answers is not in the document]
rich-iannone: This has now been fixed.

R/utils.R
rich-iannone: Change is made, thanks!

R/tab_create_modify.R
rich-iannone: This is now updated.

jcheng5 (changes requested):
1. RTF needs to use spanner id's instead of labels.
2. rle() needs to use `spanner_id` instead of `spanner`.
3. Test case with adjacent spanners that use same label but different IDs; they should not be merged into one spanner.

rich-iannone:
> * RTF needs to use spanner id's instead of labels.
> * rle() needs to use `spanner_id` instead of `spanner`.
> * Test case with adjacent spanners that use same label but different IDs; they should not be merged into one spanner.

All of these items are addressed now and I took the time to fine tune the docs for `tab_spanner()` and `tab_spanner_delim()` as well.

R/utils_render_latex.R
```diff
@@ -143,7 +144,7 @@ create_columns_component_l <- function(data) {
             paste0(
               "\\multicolumn{", spanners_lengths$lengths[i],
               "}{c}{",
-              spanners_lengths$values[i],
+              spanners[i],
```
jcheng5, line 145: I think `spanners_lengths$values[i]` was correct

tests/testthat/test-table_parts.R
```diff
@@ -187,6 +187,90 @@ test_that("a gt table contains the expected spanner column labels", {
...
+    dplyr::tibble(A_X = c(1), B_X = c(2), A_Y = c(3), B_Y = c(4)) %>%
+    gt() %>%
+    tab_spanner(label = "A", id = "a", columns = ends_with("X"), gather = FALSE) %>%
+    tab_spanner(label = "A", id = "b", columns = ends_with("Y"), gather = FALSE) %>%
+    tab_style(
+      style = cell_fill("green"),
+      locations = cells_column_spanners(spanners = "a")
+    ) %>%
+    render_as_html() %>%
+    xml2::read_html()
```
jcheng5, line 202: Let's repeat this example (spanners with same label, different ids) for latex and rtf as well

jcheng5 (changes requested):
1. Test cases with spanner id != spanner label, for HTML, latex and rtf
2. More test cases for latex and rtf

tests/testthat/test-table_parts.R
rich-iannone: Now added as tests.

jcheng5 (approved):

Issue #210, opened 2019-03-06
Suggestion: warn user before rendering N rows


daattali:
Just an idea. I just tried running `gt::gt(gt::pizzaplace)` in the rstudio console and I thought my RStudio got frozen. I tried it several times, and each time I had to abort the operation. Eventually I realized it just took a full minute to complete the render.

Perhaps it'd be friendly to the user to show a message confirming that the user indeed wants to print the entire table if gt notices that there are above N number of rows? Of course only do this in interactive mode. Or maybe even just showing a message would be enough.

rich-iannone:
I shot myself in the foot with this probably five times. Seemingly each time I thought there was a system issue and didn't realize it was a huge render. 

Basically, this is a good idea. Need to think about the right implementation (but there are lots of case studies for this sort of thing).

sharlagelfand:
I personally like `visdat::vis_dat()`'s approach, which provides a warning if your data is above a certain size (ncol*nrow > 900000), but you can turn off this warning using the `warn_large_data` parameter (see [here](https://github.com/ropensci/visdat/blob/1c3cdbe19a49792c414bec9cf9d35b7c90653cf8/R/vis-dat.R#L20))

jcheng5:
Hi @daattali, I didn't realize how slow `gt::gt(gt::pizzaplace)` was, so thanks even for that feedback alone!

@rich-iannone and I talked about this in person. I'd like to investigate two other paths first:

1. It's been a while since we've done an optimization pass on gt, a cursory investigation last week showed that huge gains are easily attainable.
2. Showing a progress bar for larger tables/slower renders would convey all of the same information and more. This is assuming that your R session doesn't have any difficulty cancelling a slower render? (Worked fine for us in RStudio when we tried it)

jcheng5:
@rich-iannone recently merged a PR that makes this on the order of 10X faster, IIRC. We might still want to do a progress bar, but we're going to move that out of this CRAN release. Thanks!

PR #666, opened 2020-10-23
Add `row_limit` arg to `gt()`

rich-iannone:
This addition imposes a row limit on rendering a gt table in an interactive session (stopping if meeting or exceeding the limit).  This can be entirely circumvented in the interactive case when setting `row_limit` to `Inf`.

Fixes: https://github.com/rstudio/gt/issues/210


rich-iannone:
Since we've decided against this in the discussion of #210, I'm going to close this PR and delete the branch. A new issue should be made that introduces a progress bar for long-running table renders.

PR #667, opened 2020-10-24
add "hidden" in possible values of cell_borders

etiennebacher:
When specifying `style = NULL` in `cell_borders()`, I have the error:
```
 Error: The provided `style` value cannot be `hidden`; it can only be either of the following:
 * `"solid"`, `"dashed"`, or `"dotted"` 
```
This commit adds "hidden" (value given to `style` when `style` is NULL) to the possible values.


rich-iannone (approved):
Looks great, thank you!

Issue #651, opened 2020-10-08
`cols_merge_n_pct()` for N (%) format


thomas-neitmann:
In the pharmaceutical industry, one of the most commonly used formats in tables is `N (%)` to display the absolute number of patients and percentage of patients relative to the population of interest, e.g. `123 (85.6%)`. It would be great to have a utility `cols_merge_n_pct()` function for that.

lcougnaud:
It could also be interesting to have an option in this function to include only the counts in case of 'zero counts', a.k.a '0' instead of '0 (0%)'

rich-iannone:
This is a very reasonable feature request and really the usefulness of having `cols_merge_*()` variants is to handle exceptions and. I can work the zero counts option into it as well. Thanks!

thomas-neitmann:
The zero option would be great indeed. I'd be happy to try to take this on myself @rich-iannone. I guess looking at the other `cols_merge_*()` should get me enough inspiration to get started. Please let me know if that's fine with you 🙂

rich-iannone:
Thomas! We'd love your contribution.

thomas-neitmann:
Awesome, looking forward to contributing 😁👍

jcheng5:
(Notes from in-person discussion with @rich-iannone)

`cols_merge*` could be more like `fmt` in the sense of being extensible and based on functions being registered with the render pipeline; as opposed to their current implementation, which has the `cols_merge*` functions register *values* that are recognized by the render pipeline. The disadvantage of this is that the set of values recognized by the render pipeline are necessarily a closed set.

For this release I don't mind if we implement `cols_merge_n_pct()` as a one-off though, given its usefulness and our desire to go to CRAN soon.

PR #683, opened 2020-11-13
Add the `cols_merge_n_pct()` function

rich-iannone:
This PR adds the `cols_merge_n_pct()` function, which merges two columns together: a count column and a percentage column. This function will autohide the `col_pct` column by default (but that can be changed by setting `autohide = FALSE`). Different combinations of `NA` are given special consideration much like `col_vals_uncert()` function. Any zero values in the `col_n` column will be formatted solely as `0`.

Fixes: #651


R/modify_columns.R
```diff
@@ -1036,6 +1036,97 @@ cols_merge_resolver <- function(data, col_begin, col_end, sep) {
...
+#'   component.
+#' @param autohide An option to automatically hide the column specified as
+#'   `col_pct`. Any columns with their state changed to hidden will behave
+#'   the same as before, they just won't be displayed in the finalized table.
+#'
+#' @return An object of class `gt_tbl`.
+#'
+#' @family Modify Columns
+#' @section Function ID:
+#' 4-10
```
jcheng5, line 1091: Example

tests/testthat/test-cols_merge.R
```diff
@@ -434,3 +434,60 @@ test_that("the `cols_merge_range()` function works correctly", {
...
+
+  # Check that specific suggested packages are available
+  check_suggests()
+
+  tbl_n_pct <-
+    dplyr::tribble(
+      ~a, ~b,
+      1,  0.0714,
+      5,  0.3571,
+      0,  0.0,
```
jcheng5, line 449: Let's add a line for `0, NA` and `NA, 0`

R/modify_columns.R
rich-iannone: Example now added.

tests/testthat/test-cols_merge.R
rich-iannone: Now added.

jcheng5 (approved):

Issue #115, opened 2018-12-18
Add option to define table caption and enable cross-referencing


larnsce:
This is an extremely useful package. Thanks! In all my reporting with .Rmd files, I use the `bookdown` file formats `pdf_document2` and `html_document2`, which allow for cross-referencing of tables and figures in text. See here: https://bookdown.org/yihui/bookdown/tables.html

Is there an option available in the `gt` package that allows for tables to have named captions and numbering, which allows for cross-referencing in text using ` \@ref(tab:tab1)`?

rich-iannone:
There is no option currently for this, but it seems like this would be a worthwhile LaTeX integration feature, similar to #112. Thanks for filing this issue.

vincentarelbundock:
The new changes I pushed in this PR address this issue: https://github.com/rstudio/gt/pull/143

I added a `label` argument to `tab_header` because that seemed to make sense, but I'm not sure if that's the design you had in mind.

Currently, label is only inserted in the LaTeX caption, but it would be trivial to use the `bookdown` html tags.

kevinykuo:
+1 Not being able to cross reference blocks writing books and academic articles (and even technical blog posts in many cases probably).

davebraze:
+1 to this. 

I have been using the subtitle as "caption" in gt tables. Could a caption argument to tab_header() be setup as an alias for subtitle, in a way that would work with bookdown's html_document2 and pdf_document2? It'd be nice to have some formatting pixie dust too.

```
gt(head(mtcars)) %>%
    tab_header(title="", subtitle="Table 1: Mtcars with gt.") %>%
    tab_options(table.width=pct(90),
                    table.border.top.style=0,
                    row.padding=px(4)) %>%
        tab_style(style=cells_styles(
                      text_align='left',
                      text_color='grey'),
                  locations=cells_title(groups="subtitle"))
```

dmenne:
Check kable: Captions and Table numbering including cross-referencing for bookdown included. I don't know of any other package that supports it, but not having these two is a no-go for science.

ballengerj:
I agree. This needs to be a priority fix for the more widespread use of this package for the creation of reports.

dmenne:
Since cross-referencing is documented confusingly (with or without #?), I have created a cheat-gist sample bookdown file, with a function `mcaption`  that can substitute for the missing features in most cases I am interested in.

https://gist.github.com/dmenne/f8eb291c9e71a5de44764d442e8bdefd

[...]

PR #689, opened 2020-11-19
Add the `caption` arg to `gt()`; support bookdown crossrefs

rich-iannone:
This PR adds caption support for gt tables, making captions work in R Markdown and in bookdown.

Fixes: #115


jcheng5 (approved):

RegalPlatypus:
the caption argument works well for me and I thank you for adding it! Is there a way to format elements of the caption (e.g., size, font, color, etc.)?

PR #692, opened 2020-11-20
bslib theming integration

cpsievert:
This PR makes it so that, when printed as an HTML table, `{gt}` will derive CSS defaults from any currently active `{bslib}` theme (via `bs_current_theme()`)


olivroy:
Hi! Would there be interest in reviving this PR?

Issue #658, opened 2020-10-13
Incorporate `cols_width()` in RTF output


rich-iannone:
The `cols_width()` function works pretty well for HTML table output (mostly deferring to the browser for the sizing implementation). However, for RTF output, we currently divide all columns evenly over a standard 8.5x11in page width (taking into account 1-in margins).

We need to add robust support for `cols_width()` and find a way to incorporate both percentage- and pixel-based measures for widths. Also, it's worth finding out if there is a method for RTF/Word to fit columns to width (making that the new default, rather than equal column sizes).

PR #696, opened 2020-12-01
Incorporate `cols_width()` in RTF output

rich-iannone:
This PR makes it possible to specify column widths for RTF output tables in the same way as HTML tables. A number of test cases have been added and rendered (in `/tests/gt-examples/`). Also, the `table.width` option specified in `tab_options()` has an effect with these changes.

Fixes: https://github.com/rstudio/gt/issues/658
Fixes: https://github.com/rstudio/gt/issues/748


R/utils_render_rtf_redo.R
```diff
@@ -240,6 +240,134 @@ rtf_table <- function(rows) {
...
+
+twip_factors <-
+  c(
+    `in` = 1440, `pt` = 20, `px` = 15,
+    `cm` = 566.9291, `mm` = 56.69291, `tw` = 1
+  )
+
+parse_length_str <- function(lengths_vec) {
+
+  if (length(lengths_vec) == 1 && lengths_vec == "auto") {
```
jcheng5, line 255: Instead of having `"auto"` be understood by `parse_length_str()`, handle `"auto"` in the caller because it's impossible to interpret it here (not enough context).

R/utils_render_rtf_redo.R
```diff
@@ -240,6 +240,134 @@ rtf_table <- function(rows) {
...
+      NA_character_
+    }
+  }, character(1))
+
+  bad_values <- nzchar(lengths_vec) & is.na(vals)
+
+  if (any(bad_values)) {
+    stop(
+      "Some of the values supplied as absolute lengths cannot be interpreted:\n",
+      "* Use either of: `px`, `pt`, `in`, `cm`, `mm`, or `tw`.",
```
jcheng5, line 283: Let's include the bad values in the error message

R/utils_render_rtf_redo.R
```diff
@@ -240,6 +240,134 @@ rtf_table <- function(rows) {
...
+# margins in a default Word document
+standard_width_twips <- 9468L
+
+twip_factors <-
+  c(
+    `in` = 1440, `pt` = 20, `px` = 15,
+    `cm` = 566.9291, `mm` = 56.69291, `tw` = 1
+  )
+
+parse_length_str <- function(lengths_vec) {
```
jcheng5, line 253:
```suggestion
# Input: character vector of any length. Valid values are numbers with
# suffix of in, pt, px, cm, mm, tw; you can also include `""`.
# Output: data frame with columns `value` and `unit`, with NA for both
# if input element was `""`. Unparseable values throw errors.
parse_length_str <- function(lengths_vec) {
```

jcheng5, line 257: Add unit tests for `parse_length_str()`

R/utils_render_rtf_redo.R
```diff
@@ -240,6 +240,134 @@ rtf_table <- function(rows) {
...
+                                   n_cols) {
+
+  if (table_width == "auto" && is.null(col_widths)) {
+    return(rep(standard_width_twips / n_cols, n_cols))
+  }
+
+  table_width <- abs_len_to_twips(parse_length_str(table_width))
+
+  # Should stop if row count of `table_width` isn't 1
+  stopifnot(nrow(table_width) == 1)
```
jcheng5, line 327: This should happen first, and I'd just check `length(table_width)` before it's converted to a table.

R/utils_render_rtf_redo.R
```diff
@@ -240,6 +240,134 @@ rtf_table <- function(rows) {
   rtf_raw(paste(unlist(rows), collapse = ""))
 }
 
+# This is the assumed width of the area within page
+# margins in a default Word document
+standard_width_twips <- 9468L
```
jcheng5, line 245: It'd be great to have an `options()` option to set this value, e.g. for changing the default page size for a whole document.

R/utils_render_rtf_redo.R
rich-iannone: Now changed.

R/utils_render_rtf_redo.R
rich-iannone: Now implemented.

R/utils_render_rtf_redo.R
rich-iannone: Done (by you, actually).

R/utils_render_rtf_redo.R
rich-iannone: Now added as `gt.rtf_page_width` (equal to `9468L`).

R/utils_render_rtf_redo.R
rich-iannone: Now added.

R/utils_render_rtf_redo.R
```diff
@@ -240,20 +240,181 @@ rtf_table <- function(rows) {
...
+      m[[3]]
+    } else {
+      NA_character_
+    }
+  }, character(1))
+
+  # Check for negative values and stop of `allow_negative = FALSE`
+  non_na_vals <- vals[!is.na(vals)]
+  if (!allow_negative &&
+      length(non_na_vals > 0) &&
```
jcheng5, line 279:
```suggestion
      length(non_na_vals) > 0 &&
```

R/utils_render_rtf_redo.R
```diff
@@ -240,20 +240,181 @@ rtf_table <- function(rows) {
...
+
+  # Ensure that the `table_width` unit is `"tw"`
+  stopifnot(isTRUE(table_width$unit == "tw"))
+
+  # If given a table width (i.e., not `"auto"`) without any specification
+  # of column widths, generate evenly spaced column widths based on the
+  # table width
+  if (is.null(col_widths)) {
+    col_widths <- paste0(rep(table_width$value / n_cols, n_cols), "tw")
+  }
```
jcheng5, line 362: Remove this block--it's already handled above

R/utils_render_rtf_redo.R
```diff
@@ -240,20 +240,181 @@ rtf_table <- function(rows) {
...
+# The `col_width_resolver_rtf()` function returns a vector of
+# RTF column widths in units of twips (numeric)
+#
+# `table_width` and `col_widths` should be character vectors with
+# a numeric portion followed by the unit; the first should be a
+# length-1 vector and the latter should be a vector with as many
+# elements as there are visible columns in the gt table
+col_width_resolver_rtf <- function(table_width,
+                                   col_widths,
+                                   n_cols) {
```
jcheng5, line 330: I don't love that `n_cols` can potentially be different than `length(col_widths)`. The only reason we need `n_cols` is for the case where `col_widths` is NULL; we want to generate evenly-spaced columns.

jcheng5: Also, `col_widths = NULL` should be the default

jcheng5: At the very least, we should enforce that if `col_widths` is non-`NULL`, then `n_cols` must be the same as `length(col_widths)` (and we can even make `n_cols = length(col_widths)` the default).

R/utils_render_rtf_redo.R
```diff
@@ -240,20 +240,181 @@ rtf_table <- function(rows) {
...
+  stopifnot(isTRUE(table_width$unit == "tw"))
+
+  # If given a table width (i.e., not `"auto"`) without any specification
+  # of column widths, generate evenly spaced column widths based on the
+  # table width
+  if (is.null(col_widths)) {
+    col_widths <- paste0(rep(table_width$value / n_cols, n_cols), "tw")
+  }
+
+  col_widths <- abs_len_to_twips(parse_length_str(col_widths))
```
jcheng5, line 364: Left off here

R/utils_render_rtf_redo.R
rich-iannone: Now removed.

R/utils_render_rtf.R
```diff
@@ -1244,18 +1244,17 @@ create_body_component_rtf <- function(data) {
...
         i %in% groups_rows_df$row_end) {
 
-      group <-
-        groups_rows_df %>%
-        dplyr::filter(row_end == i) %>%
-        dplyr::pull(group)
+      group_id <-
+        groups_rows_df[
+          groups_rows_df$row_end == i &
+            !is.na(groups_rows_df$row_end), ][["group_id"]]
```
jcheng5, line 1250:
```r
groups_rows_df[na.omit(groups_rows_df$row_end == i), "group_id", drop = TRUE]
```

rich-iannone: Replaced statement with your suggestion. Thank you!

R/utils_render_rtf_redo.R
rich-iannone: The above suggestions were incorporated in `col_width_resolver_rtf()`.

jcheng5 (dismissed):

jcheng5 (approved):

Issue #675, opened 2020-11-05
`tab_row_group()` signals warning when used successively


kinto-b:
Annoying warning is signaled when adding two or more groupings by hand

```r
library(dplyr)
mtcars %>% 
gt::gt() %>% 
gt::tab_row_group(group = "One", rows = which(mtcars$am == 1)) %>% 
gt::tab_row_group(group = "Zero", rows = which(mtcars$am == 0))

# Warning messages:
# 1: In min(rows_matched) : no non-missing arguments to min; returning Inf
# 2: In max(rows_matched) : no non-missing arguments to max; returning -Inf
```

rich-iannone:
Thanks for reporting on this, seeing it as well. Currently working on fixing this (along with a bigger issue) in a branch that will be a PR soon.

Issue #671, opened 2020-10-30
Cant apply bold to row_group that has a footnote


luis-zapata-b:
I cant apply bold to a rop_group that has a footnote. Some help please?

exibble %>%
  dplyr::select(num, currency,group) %>%
  gt(rowname_col = "num",
     groupname_col = "group")%>%

tab_footnote(
  footnote = "testing footnotes",
  cells_row_groups(groups = "grp_a")
) %>% 
  tab_style(
    style = cell_text(weight = "bold"),
    locations = cells_row_groups()
  )

rich-iannone:
Try this:

```r
exibble %>%
  dplyr::select(num, currency,group) %>%
  gt(rowname_col = "num",
     groupname_col = "group",
     id = "table"
  ) %>%
  tab_footnote(
    footnote = "testing footnotes",
    cells_row_groups(groups = "grp_a")
  ) %>%
  tab_style(
    style = cell_text(weight = "bold"),
    locations = cells_row_groups()
  ) %>% opt_css(
    css = "
    #table .gt_group_heading {
      font-weight: bold;
    }
    "
  )
```

luis-zapata-b:
Hi. I tried it on my dataset, and it hasn't any effect. Also, in others dataset it doesn't work neither (i couldn't apply bold):

mtcars %>% 
  select(disp,hp,drat,wt,gear) %>% 
  mutate(gear = ifelse(gear==4,"test1",
                       ifelse(gear==3,
                              "test2","test3"))) %>% 
  gt(rowname_col = "hp",
     groupname_col = "gear") %>% 
  tab_footnote(
    footnote = "test1",
    cells_row_groups(groups = "test1")
  )%>% 
  tab_footnote(
    footnote = "test2",
    cells_row_groups(groups = "test2")
  ) %>% 
  tab_footnote(
    footnote = "test3",
    cells_row_groups(groups = "test3")
  )%>% 
  opt_css(
    css = "
    #table .gt_group_heading {
      font-weight: bold;
    }
    "
  )

rich-iannone:
That code is slightly incorrect, it should be:

``` r
mtcars %>%
  select(disp,hp,drat,wt,gear) %>%
  mutate(gear = ifelse(gear==4,"test1",
                       ifelse(gear==3,
                              "test2","test3"))) %>%
  gt(rowname_col = "hp",
     groupname_col = "gear",
     id = "table") %>%
  tab_footnote(
    footnote = "test1",
[...]

PR #699, opened 2020-12-10
Add row group `id` as an argument to `tab_row_group()`

rich-iannone:
This PR seeks to do what https://github.com/rstudio/gt/pull/665 did recently, which is to require and use IDs for different row groups. The function signature for `tab_row_group()` will be similar to that of the analogous `tab_spanner()`:

``` r
tab_row_group(
  data,
  label = NULL,
  rows = NULL,
  id = label,
  others = NULL
)
```

This is not the only way to specify row groups in a **gt** table. Here is a breakdown of the three different ways to include them:

1. with grouped data given to `gt()`; group labels are composed by the group column names being pasted together, each separated by `_`
2. using the `groupname_col` arg in `gt()`; we can only choose one column so flexibility in grouping is purposefully limited compared to method (1)
3. by use of the `tab_row_group` function. This is used past the initial `gt()` step.

Aside from introducing IDs here, this PR aims to fix a number of known bugs along the way.

Fixes: https://github.com/rstudio/gt/issues/671
Fixes: https://github.com/rstudio/gt/issues/675
Fixes: https://github.com/rstudio/gt/issues/717


R/tab_create_modify.R
```diff
@@ -351,9 +389,17 @@ tab_spanner_delim <- function(data,
 #' @import rlang
 #' @export
 tab_row_group <- function(data,
-                          group = NULL,
+                          label = NULL,
                           rows = NULL,
-                          others = NULL) {
+                          id = label,
+                          others_label = NULL) {
```
jcheng5, line 395: Can we change the `other_label` default to `missing_arg`, and allow users to unset `others_label` by explicitly passing `NULL`?

jcheng5: Or perhaps `others_label` goes into tab options as `default_row_group_label` or something

rich-iannone: The second option was implemented (i.e., using `tab_options()` to set the default row group label).

R/tab_create_modify.R
```diff
@@ -214,11 +218,14 @@ tab_spanner <- function(data,
 tab_spanner_delim <- function(data,
                               delim,
                               columns = everything(),
+                              split = c("last", "first"),
```
jcheng5, line 221: Introduce new params as the last param, unless there's a good reason not to--so it's not a breaking change

R/tab_create_modify.R
```diff
@@ -247,7 +254,30 @@ tab_spanner_delim <- function(data,
 
     split_colnames <- strsplit(colnames_with_delim, delim, fixed = TRUE)
 
-    spanners <- vapply(split_colnames, `[[`, character(1), 1)
+    if (split == "first") {
```
jcheng5, line 257: For next time: we can do this just by varying the pattern

R/tab_create_modify.R
rich-iannone: Changed now, thanks!

R/tab_create_modify.R
rich-iannone: Now done, thanks again.

R/tab_create_modify.R
```diff
@@ -405,24 +469,14 @@ tab_row_group <- function(data,
...
-  # Set a name for the `others` group if a
-  # name is provided
-  if (!is.null(others)) {
-    data <-
-      dt_stub_others_set(
-        data = data,
-        stub_others = process_text(others[1])
-      )
-  }
-
```
jcheng5, line 425: Instead of getting rid of this, put in a deprecation warning and use `tab_options()` to actually do it.

R/tab_create_modify.R
```diff
@@ -363,8 +399,28 @@ tab_row_group <- function(data,
...
+  if (rlang::as_label(row_expr) == "NULL")  {
+
+    stop(
+      "No value was provided for `rows`. Please use either of:\n",
+      "* A vector of row captions provided in `c()`\n",
+      "* A vector of row indices\n",
+      "* A helper function focused on selections (e.g., `starts_with()`)",
+      call. = FALSE
+    )
+  }
```
jcheng5, line 421: Let's add a unit test that does `expect_warning(tbl %>% tab_row_group(others = "blah"))`

rich-iannone: Now in.

R/tab_create_modify.R
rich-iannone: Done.

jcheng5:
Just to note a limitation on the column splitting logic, if `delim` contains `"\\E"` then the behavior will be wrong.

```r
split_at_first_delim <- function(delim, str) {
  delim <- gsub("\\E", "\\\\E", delim, fixed = TRUE)
  m <- regexec(perl = TRUE,
    paste0(
      "^(.*?)\\Q", delim, "\\E(.*)$"
    ),
    str
  )
  regmatches(str, m)
}

split_at_first_delim("\\E", "foEo\\E.bar")  # expected: foEo|.bar
split_at_first_delim("\\", "foo\\bar")      # expected: foo|bar
split_at_first_delim("\\\\E", "foo\\\\Ebar")      # expected: foo|bar
```

R/utils_render_html.R
```diff
@@ -752,11 +753,12 @@ create_body_component_h <- function(data) {
   all_default_vals <- unname(as.matrix(body[, default_vars]))
 
   alignment_classes <- paste0("gt_", col_alignment)
+
   if (stub_available) {
     n_cols <- n_data_cols + 1
     alignment_classes <- c("gt_left", alignment_classes)
     stub_var <- dt_boxhead_get_var_stub(data = data)
-    all_stub_vals <- as.matrix(body[, stub_var])
+    all_stub_vals <- as.matrix(stub_df[, "rowname"])
```
jcheng5, line 761: What's the reason for this change? Does it work when the data column for the stub isn't called "rowname"?

R/utils_render_latex.R
```diff
@@ -214,16 +212,8 @@ create_body_component_l <- function(data) {
...
-    default_vars <- c("::rowname", default_vars)
-
-    body <-
-      dt_stub_df_get(data = data) %>%
-      dplyr::select(rowname) %>%
-      dplyr::mutate(rowname = process_text(rowname, context = "latex")) %>%
-      dplyr::rename(`::rowname` = rowname) %>%
-      cbind(body)
+  if (stub_available && "__GT_ROWNAME_PRIVATE__" %in% names(body)) {
+    default_vars <- c("__GT_ROWNAME_PRIVATE__", default_vars)
```
jcheng5, line 216: Please test with LaTeX, where the rowname contains a LaTeX special character (that would need to be escaped by us).

jcheng5 (dismissed):

jcheng5 (approved):

jcheng5 (approved):

Issue #700, opened 2020-12-18
Feature Request: Unhide columns


ddsjoberg:
Hello Rich! In gtsummary, we auto-hide some columns. Sometimes users convert the gtsummary objects to gt and need to unhide these columns. Can we get a `cols_unhide()` companion to `cols_hide()`?

ヽ(ຈل͜ຈ)ﾉ︵ ┻━┻🍁

ddsjoberg:
Just realized this is related to https://github.com/rstudio/gt/issues/26

rich-iannone:
This will be prioritized for the upcoming release. Thanks!

ddsjoberg:
Thank you!! 😍

Issue #26, opened 2018-10-04
Add a `cols_unhide()` function and fix implementation for hiding/unhiding


rich-iannone:
This would work with `cols_hide()` and have a `column_display` row in `boxh_df` that can be switched between `show` (default) and `hide`. The render pipeline (near the end of it) would look at this and delete columns from `output_df` based on what's hidden.

PR #710, opened 2021-01-29
Add the `cols_unhide()` function

rich-iannone:
The `cols_unhide()` function is added here to make specific columns visible in the final render in a targeted manner. This is useful when provided the `gt_tbl` object without the source data and code to recreate it *AND* there are hidden columns that should be visible.

Fixes: #26 
Fixes: #700


R/modify_columns.R
```diff
@@ -794,6 +797,98 @@ cols_hide <- function(data,
...
+  # Stop function if no `columns` are provided
+  if (length(columns) == 0) {
+    stop("Columns must be provided.", call. = FALSE)
+  }
+
+  # Stop function if any of the `columns` don't exist in `vars`
+  if (!all(columns %in% vars)) {
+    stop("All `columns` must exist in the input `data` table.",
+         call. = FALSE)
+  }
```
jcheng5, line 883: Think this is done by `resolve_cols_c`

jcheng5 (approved):

PR #718, opened 2021-02-26
Convert columns to tidyselect semantics

jcheng5:
- [x] Revisit version number in `translate_legacy_resolve_expr` deprecation messages once we know what version we'll actually ship this
- [x] Implement `resolve_rows_l`/`resolve_rows_i`
- [x] Convert all existing code to use new resolution functions
- [x] ~~Fix `gt(exibble) %>% summary_rows(columns = num, fns = list(mean = ~mean(.)))`~~


R/resolver.R
```diff
@@ -314,3 +314,163 @@ resolve_vars <- function(var_expr,
   # Translate the column indices to column names
   column_names[columns_idx]
 }
+
+#' @param var_expr An unquoted expression that follows tidyselect semantics
+#' @param data A gt object or data frame or tibble
+#' @return Character vector
+#' @noRd
+resolve_cols_c <- function(var_expr, data, strict = TRUE) {
+  names(resolve_cols_i({{var_expr}}, data = data, strict = strict))
```
jcheng5, line 323:
```suggestion
#' @param expr An unquoted expression that follows tidyselect semantics
#' @param data A gt object or data frame or tibble
#' @return Character vector
#' @noRd
resolve_cols_c <- function(expr, data, strict = TRUE) {
  names(resolve_cols_i({{expr}}, data = data, strict = strict))
```

R/resolver.R
```diff
@@ -314,3 +314,163 @@ resolve_vars <- function(var_expr,
...
+
+  } else if (is.numeric(resolved)) {
+
+    if (length(setdiff(resolved, seq_len(nrow(data)))) != 0) {
+      stop(
+        "The following rowname(s) do not exist in the data: ",
+        paste0(setdiff(resolved, seq_len(nrow(data))), collapse = ", "),
+        call. = FALSE
+      )
+    }
```
jcheng5, line 445:
```suggestion
    unknown_indices <- setdiff(resolved, seq_len(nrow(data)))
    if (length(unknown_indices) != 0) {
      stop(
        "The following row number(s) do not exist in the data: ",
        paste0(unknown_indices, collapse = ", "),
        call. = FALSE
      )
    }
```

R/resolver.R
```diff
@@ -314,3 +314,163 @@ resolve_vars <- function(var_expr,
...
+
+  } else if (is.character(resolved)) {
+
+    if (length(setdiff(resolved, row_names)) != 0) {
+      stop(
+        "The following rowname(s) do not exist in the data: ",
+        paste0(setdiff(resolved, row_names), collapse = ", "),
+        call. = FALSE
+      )
+    }
```
jcheng5, line 458:
```suggestion
    unknown_rownames <- setdiff(resolved, row_names)
    if (length(unknown_rownames) != 0) {
      stop(
        "The following rowname(s) do not exist in the data: ",
        paste0(unknown_rownames, collapse = ", "),
        call. = FALSE
      )
    }
```

R/location_methods.R
```diff
@@ -70,26 +71,22 @@ add_summary_location_row <- function(loc,
...
 
-    col_idx <-
-      resolve_data_vals_idx(
-        var_expr = !!loc$columns,
-        data_tbl = NULL,
-        vals = dt_boxhead_get_vars_default(data = data)
+    columns <-
+      resolve_cols_c(
+        expr = !!loc$columns,
+        data = data
```
rich-iannone, line 77: Let's make sure we have a test case for adding a summary on a column that's hidden: both whether that produces an error or not, and if not, whether unhiding the column will result in the summary being shown.

rich-iannone: Now done. Left a TODO in the testing code to circle back and test the rendered table with `cols_unhide()` after the PR with that new function is merged.

R/resolver.R
```diff
@@ -352,64 +315,98 @@ resolve_vector_l <- function(expr, vector) {
...
+resolver_stop_unknown <- function(item, resolved) {
+
+  stop(
+    "Don't know how to select ", item , "s using an object of class ",
+    class(resolved)[1],
+    call. = FALSE
+  )
+}
+
+perform_resolution <- function(resolved, item, item_count, item_names, item_sequence) {
```
jcheng5, line 374: 1. Let's rename `perform_resolution` to `normalize_resolved`
2. Let's derive `item_count` and `item_sequence` from `item_names` as needed, instead of taking them all as arguments
3. Let's rename `item` to `item_label`

R/resolver.R
```diff
@@ -352,64 +315,98 @@ resolve_vector_l <- function(expr, vector) {
...
+
   if (is.null(resolved)) {
 
-    # Maintained for backcompat
-    resolved <- rep_len(TRUE, length(vector))
+    # Maintained for backcompatability
+    resolved <- rep_len(TRUE, item_count)
```
jcheng5, line 379: Warn that `NULL` should be `everything()`

R/resolver.R
```diff
@@ -352,64 +315,98 @@ resolve_vector_l <- function(expr, vector) {
       expr = rlang::eval_tidy(expr = quo, data = NULL)
     )
 
+  item_names <- vector
+  item_count <- length(vector)
+  item_sequence <- seq_along(vector)
+
```
jcheng5, line 321: Add a line that checks if `resolved` is `NULL`; if so,
1. warn that `NULL` is deprecated and they should use `TRUE` instead;
2. change the value of `resolved` to `TRUE`

R/resolver.R
rich-iannone: Now done.

R/resolver.R
rich-iannone: All changed now.

R/resolver.R
rich-iannone: Now done!

R/resolver.R
```diff
@@ -160,157 +154,272 @@ resolve_cells_column_spanners <- function(data,
...
 
-  # With the `resolved` output, check types and
-  # process inputs to reliably output as a vector
-  # of column indices based on `vals`
   if (is.null(resolved)) {
 
-    resolved <- seq_along(vals)
+    warning(
+      "The use of `NULL` for rows is deprecated since gt 0.2.3:\n",
+      "* please use `everything()` instead",
```
jcheng5, line 290:
```suggestion
      "* please use `TRUE` instead",
```

R/resolver.R
```diff
@@ -160,157 +154,272 @@ resolve_cells_column_spanners <- function(data,
...
+
+    # Maintained for backcompatability
+    resolved <- rep_len(TRUE, item_count)
+
+    # TODO: this may not apply to all types of resolution so we may
+    # want to either make this warning conditional (after investigating which
+    # resolving contexts still allow `NULL`)
+    warning(
+      "The use of `NULL` for ", item_label , "s is deprecated since gt 0.2.3:\n",
+      "* please use `everything()` instead",
```
jcheng5, line 354:
```suggestion
      "* please use `TRUE` instead",
```

R/utils.R
```diff
@@ -1318,8 +1318,14 @@ validate_css_lengths <- function(x) {
...
 
+  resolved <-
+    resolve_cols_c(
+      expr = {{ columns }},
+      data = data
+    )
+
   dt_data_get(data = data) %>%
-    dplyr::select(resolve_vars(var_expr = {{columns}}, data = data)) %>%
+    dplyr::select(resolved) %>%
```
jcheng5, line 1328:
```suggestion
    dplyr::select(dplyr::all_of(resolved)) %>%
```

rich-iannone (approved):
Everything looks great now.

PR #724, opened 2021-03-15
Fix for bookdown table reference support

### LINKED ISSUE #719: Cross references don't work in HTML
Using `caption` doesn't quite work for `bookdown::html_document2()`

Reprex: https://rstudio.cloud/project/2259811

Screenshot:

Screen Shot 2021-03-02 at 3 32 21 PM

cderv:
This will fix #719 

The escaping of # in `(\\# ... )` is not needed because the syntax won't be processed by Pandoc. This does not need to be escape. 

**bookdown** reference mechanism works by matching `(#tab:label)` 

This will fix the auto numbering of the reference.

But there is also an issue because `(#tab:label)` is not removed from the caption. I think this needs to be patched in **bookdown**.

Just sharing this to easily show the fix and test. I let you decide if you want to fix differently.

### DIFF
--- a/R/knitr-utils.R
+++ b/R/knitr-utils.R
@@ -12,7 +12,8 @@ kable_caption = function(label, caption, format) {
 # create \label{x} or (\#x); the latter is current an internal hack for bookdown
 create_label = function(..., latex = FALSE) {
   if (isTRUE(knitr::opts_knit$get('bookdown.internal.label'))) {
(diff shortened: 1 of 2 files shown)


cderv:
Companion PR is https://github.com/rstudio/bookdown/pull/1099/

jcheng5 (approved):

Issue #510, opened 2020-03-23
Troubleshooting local_image()


Worville:
Hi,

I'm trying to use the local_image() helper to enter a .png file into a table.

If I try and create the path to the local image in my dataframe, then point local_image() to that path, I get an error of 

```"Error in file(con, "rb") : invalid 'description' argument"```

Here's a minimum working (well, failing) example:

```
data <- data.frame(team_id = c(1, 22, 1),
                   stat = c(3, 4, 5)) %>% 
  mutate(logo_loc = str_glue('/path/to/image/{team_id}_logo.png'))

data %>% 
  gt() %>% 
  text_transform(
    locations = cells_body(columns = "logo_loc"), 
    fn = function(x) {local_image(filename =  as.character(x), height = 120)})
```

Any thoughts or advice?

Thanks a lot in advance.

primaj:
I posted a pull request which should fix this! 

Here's a workaround:

`tibble(img_path = rep(test_image(), 5)) %>% 
  gt() %>%
  text_transform(
    locations = cells_body(columns = vars(img_path)),
    fn = function(x) map_chr(x, ~local_image(filename = .x, height=50))
  )`

rich-iannone:
Now fixed in https://github.com/rstudio/gt/pull/803. Sorry @primaj for not accepting your PR, it turns out we had to do a lot of extra work on top of the fix.

PR #738, opened 2021-04-09
Bug fix in local_image where it errors on adding multiple images

primaj:
On trying to use the `local_image` function with more than 1 image, you get this error:

`"Error in file(con, "rb") : invalid 'description' argument"`

Had a look at the `local_image` function and noticed it was because the `get_image_uri` function can't handle arrays of paths, hence added an lapply for `get_image_uri` and it seems to have fixed the issue!

Fixes: #510


rich-iannone:
Thanks so much for contributing this PR. Unfortunately we determined that a lot more effort had to be done to fix related image issues (resulting in https://github.com/rstudio/gt/pull/803).

Issue #652, opened 2020-10-08
Page number for multi-page outputs


thomas-neitmann:
For tables that span multiple pages it would be great to have `Page x of X` being displayed in the document footer when outputting tables as `.rtf`.

rich-iannone:
I'll see if this could be done (though, it seems like it should be possible since another package was mentioned as providing this feature). Thanks Thomas. This should have been more obvious to me before :)

PR #755, opened 2021-05-05
Add RTF page numbering

rich-iannone:
This enhances the `gtsave()` and `as_rtf()` functions such that page numbering in RTF documents is possible. The page numbering text can either be in the document header or footer and is inactive by default (i.e., page numbering needs to be explicitly turned on).

Fixes: #652


R/export.R
```diff
@@ -231,12 +237,19 @@ gt_save_latex <- function(data,
 gt_save_rtf <- function(data,
                         filename,
                         path = NULL,
-                        ...) {
+                        ...,
+                        page_numbering.active = FALSE,
+                        page_numbering.location = c("footer", "header")) {
```
jcheng5, line 242: Let's use one option: `page_numbering = c("none", "footer", "header")`

rich-iannone: Now done. Thanks!

jcheng5 (approved):

Issue #715, opened 2021-02-17
column name leak/conflict?


espinielli:
`cols_label` fails when you have a column named `d`.
Without a call to `cols_label` it works.
Using another column name it works too.

``` r
library(tibble)
library(gt)

# this WORKS
# tribble(
#   ~a , ~e ,
#   1, 4,
#   5, 8
# ) %>% 
#   gt() %>% 
#   cols_label(
#     a = "label a",
#     e = "label e"
#   )

tribble(
  ~a , ~d ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    d = "label d"
  )
#> Error: The object to `data` is not a `gt_tbl` object.
```

<sup>Created on 2021-02-17 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

KaiAragaki:
This appears to be some kind of partial matching with a 'data' argument. 
``` r
library(tibble)
library(gt)
# Fails
tribble(
  ~a , ~dat ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    dat = "label d"
  )
#> Error: The object to `data` is not a `gt_tbl` object.

# Fails
tribble(
  ~a , ~data ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    dat = "label d"
  )
#> Error: The object to `data` is not a `gt_tbl` object.


# Works
tribble(
  ~a , ~datr ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    dat = "label d"
  )
#> Error: The object to `data` is not a `gt_tbl` object.

# Works
tribble(
  ~a , ~datas ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    datas = "label d"
  )
```


<sup>Created on 2021-04-02 by the [reprex package](https://reprex.tidyverse.org) (v1.0.0)</sup>

rich-iannone:
Thanks for posting this issue @espinielli and thanks @KaiAragaki for finding the root cause. What needs to happen is to change the `data` arg to `.data` and avoid most cases of this. There are likely other things that could be done in addition to this.

PR #772, opened 2021-05-26
Change the `data` arg to `.data` to avoid partial matching issues

rich-iannone:
In functions like `cols_label()`, having a column name close to `data` may result in a partial matching error. Here's a reproducible example of the error:

```r

tribble(
  ~a , ~d ,
  1, 4,
  5, 8
) %>% 
  gt() %>% 
  cols_label(
    a = "label a",
    d = "label d"
  )
```

```
#> Error: The object to `data` is not a `gt_tbl` object.
```

This can be solved by changing the argument name to `.data`. For consistency, all instances of `data` as argument names in exported functions were changed to `.data`.

On the user side, there should hopefully be little disruption as all examples and ways of using the gt API don't use the argument name `data` (either via piping or through implicit use of `data` when just using `gt()`).

Fixes: https://github.com/rstudio/gt/issues/715


jcheng5:
Let's check `lapply(as.list(as.environment("package:gt")), args)` and see what other functions have unbounded possibilities for `...`

rich-iannone:
> Let's check `lapply(as.list(as.environment("package:gt")), args)` and see what other functions have unbounded possibilities for `...`

The `cols_label()` and `cols_width()` functions both have `...` but the only other non-`data` arg for both is `.list`. So no other issues there.

The `summary_rows()` and `grand_summary_rows()` functions also have `...` but after looking at the function signatures of all the `fmt_*()` functions, none except for `fmt_missing()` (which you wouldn't use in `summary_rows()` or `grand_summary_rows()`) have arguments that collide with the summary functions. Quite a bit of testing was done to check that partial matching doesn't lead you to a strange error in the summary functions.

rich-iannone:
Changes have now been made to limit the `data` -> `.data` arg renaming to only the `cols_label()` and `cols_width()` functions.

jcheng5 (approved):

PR #768, opened 2021-05-20
LaTeX bugfix for summary rows, and, implementation of grand summary rows

rich-iannone:
This PR enables the proper rendering of summary rows in LaTeX table outputs. On top of that, there's an implementation of grand summary rows for the first time (previously, grand summary rows would not be rendered).

Two .Rmds have also been added to the `scripts` folder to aid in LaTeX testing (they serve as templates).

Fixes: #625

PR #777, opened 2021-05-27
FIX: Fix LaTeX rendering of summary rows

elmstedt:
Add `boxh` formal argument to `create_summary_rows()`.
`boxh` in `create_summary_rows()` allows the creation of `default_vars` in a manner consistent with the method in `summary_row_tags()`.
`default_vars` is used in the selection for `summary_df`.

Wrapped `visible_vars` with a call to `unique()`. Line 218 could add `"rowname"` to `visible_vars` even if it already exists in `default_vars` (from line 184). This keeps the table body from becoming misaligned.

### DIFF
--- a/R/utils_render_common.R
+++ b/R/utils_render_common.R
@@ -454,9 +454,13 @@ create_summary_rows <- function(n_rows,
                                 n_cols,
                                 list_of_summaries,
                                 groups_rows_df,
+                                boxh,
                                 stub_available,
                                 summaries_present,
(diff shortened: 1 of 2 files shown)


rich-iannone:
Thanks for your contribution. Turns out we don't need the changes made in this PR because they're covered in https://github.com/rstudio/gt/pull/768 and https://github.com/rstudio/gt/pull/792. Will close this now.

Issue #424, opened 2019-11-27
cells_title() can accept any input without stopping with an error message


MadhulikaTanuboddi:
In the console, give the following function `cells_title("test")`

Expected: Should show a error message saying only `title` and `subtitle` are allowed
Actual: Snapshot below

![cells_text_invalid_argument_works](https://user-images.githubusercontent.com/5369655/69755952-3f73f600-110e-11ea-9628-d776501871b1.png)

PR #783, opened 2021-06-02
Ensure that `cells_title()` errors on disallowed output

rich-iannone:
This PR provides more rigorously checks the inputs of `cells_title()`, erroring and providing actionable messaging when stopping does occur.

Fixes: https://github.com/rstudio/gt/issues/424


rich-iannone:
@jcheng5 I'm starting to think that the `stop()` in `cols_hide()` when no columns are resolved might be the wrong choice altogether. The formatters don't behave like this. For example, using this:

```r
library(gt)

mtcars %>%
  gt() %>%
  fmt_number(columns = starts_with("z"), decimals = 5)
```

doesn't result in an error but instead has no effect on the rendered output (no column in `mtcars` begins with `"z"`). The formatting behavior is pretty good, so, maybe the thing to do in `cols_hide()` is:

```r
  # Do nothing if no `columns` are resolved
  if (length(columns) == 0) {
    return(data)
  }
```

instead of:

```r
  # Stop function if no `columns` are provided
  if (length(columns) == 0) {
    stop("Columns must be provided.", call. = FALSE)
  }
```

R/tab_create_modify.R
```diff
@@ -632,7 +632,25 @@ set_footnote <- function(loc, data, footnote) {
...
+        data = data,
+        locname = c("title", "subtitle"),
+        grpname = NA_character_,
+        colname = NA_character_,
+        locnum = c(1, 2),
+        rownum = NA_integer_,
+        styles = footnote
+      )
+
+  } else if (title_components == "title") {
```
jcheng5, line 653: Instead of if/elseif/elseif, I would make it two completely separate conditionals (one for title, one for subtitle) that are not mutually exclusive.

R/tab_create_modify.R
```diff
@@ -1158,7 +1176,25 @@ set_style <- function(loc, data, style) {
...
+        data = data,
+        locname = c("title", "subtitle"),
+        grpname = NA_character_,
+        colname = NA_character_,
+        locnum = c(1, 2),
+        rownum = NA_integer_,
+        styles = style
+      )
+
+  } else if (title_components == "title") {
```
jcheng5, line 1197: Same comment as above

rich-iannone: Done.

jcheng5 (approved):

Issue #749, opened 2021-04-26
When using `summary_rows()` with no stub, there is a possibility for column name duplication


rich-iannone:
When using `summary_rows()` with no stub, a new column is generated internally and given the name `rowname`. This can conflict with a column that has the same name, generating duplicate column names and not erroring at that instant. Instead, the error occurs at the build phases and the error messages are uninformative. 

Here is some code that demonstrates some of the ways this might happen:

```r
library(gt)
library(tidyverse)

# Case where the `rowname` column conflicts with the
# similarly named `rowname` column in the internal summary df
exibble %>%
  rename(rowname = row) %>%
  # gt(rowname_col = NULL) %>% # doesn't work, Error: Column name `rowname` must not be duplicated.
  # gt(rowname_col = NULL, groupname_col = "group") %>% # doesn't work, Error: Column name `rowname` must not be duplicated.
  # gt(rowname_col = "char") %>% # works, `mapply()` warnings appear
  # gt(rowname_col = "char", groupname_col = "group") %>% # works, `mapply()` warnings appear
  grand_summary_rows(
    columns = c(num, currency),
    fns = list(
      min = ~min(., na.rm = TRUE),
      max = ~max(., na.rm = TRUE),
      avg = ~mean(., na.rm = TRUE)),
    formatter = fmt_number
  )
```

PR #792, opened 2021-06-09
Fix for possible column name duplication when generating summary rows

rich-iannone:
When using `summary_rows()` with no stub, a new column is generated internally and given the name `rowname`. This can conflict with a column in the input data that has the same name, generating duplicate column names and downstream errors. This fix decreases the probability that internal names used in temporary tables will conflict with column names from the input table.

Fixes: https://github.com/rstudio/gt/issues/749


jcheng5 (changes requested):
Any `::xxx::` variables should be extracted into global variables instead of hardcoded literals.

Also add tests that collide with the old column names.

jcheng5 (dismissed):

jcheng5 (approved):

Issue #802, opened 2021-06-17
Several `cell_text()` parameters don't work when applied to `cells_body()` in a `tab_style()` call


nikhilpathiyil:
When making a call to cell_text() with cells_body() containing the targeted location, most of the parameters in cell_text() don't have any effect, except for "color". This is in version 0.3.0.
Reprex below.

```r
  gtcars %>%
    dplyr::filter(ctry_origin == "United Kingdom") %>%
    dplyr::select(mfr, model, year, hp) %>%
    gt() %>%
    tab_style(
      style = list(
        cell_text(align = 'left',
                  indent = px(25),
                  font = c("sans-serif", gt::default_fonts()),
                  size = px(50),
                  weight = "bold",
                  color = "red") # the "color" parameter is the only one that works
      ),
      locations = cells_body(
        columns = c(hp),
        rows = c(1,2)
      )
    )
```

rich-iannone:
Able to reproduce this. The issue here is the use of the `list()`. When using the following:

```r
gtcars %>%
  dplyr::filter(ctry_origin == "United Kingdom") %>%
  dplyr::select(mfr, model, year, hp) %>%
  gt() %>%
  tab_style(
    style = 
      cell_text(
        align = 'left',
        indent = px(25),
        font = c("sans-serif", gt::default_fonts()),
        size = px(50),
        weight = "bold",
        color = "red"
      ),
    locations = cells_body(
      columns = hp,
      rows = 1:2
    )
  )

```

the table has all of the styling applied.

Issue #661, opened 2020-10-18
`google_font()` does not seem to work when passed as a list into `tab_style()` 


kaustavSen:
When passing `google_font()` as a list into `tab_style()`, it does not seem to work, i.e. the font face does not change. 

```r
library(gt)

mtcars %>%
  head() %>% 
  gt() %>% 
  tab_style(
    style = list(
      cell_text(font = google_font("Fira Code")),
      cell_fill(color = "#F5F3F1")
    ),
    locations = cells_body(everything())
  )
```
![image](https://user-images.githubusercontent.com/16508583/96361694-84998000-1145-11eb-8a90-ba8177847504.png)

However, if I pass it as two separate calls to `tab_style()` and **not** as a `list()`:

* one for changing the font-face
* second for updating the background color of the cells

then it works:

```r
mtcars %>%
  head() %>% 
  gt() %>% 
  # First call to change font-face
  tab_style(
    style = cell_text(font = google_font("Fira Code")),  # without list()
    locations = cells_body(everything())
  ) %>% 
  # Second call to update background color
  tab_style(
    style = cell_fill(color = "#F5F3F1"),  # without list()
    locations = cells_body(everything())
  )
```
![image](https://user-images.githubusercontent.com/16508583/96361757-05f11280-1146-11eb-9f75-bbe765424136.png)

rich-iannone:
Ran into this myself recently and it’s no good. I hope to fix this soon.

PR #805, opened 2021-06-21
`cell_text()` fix

rich-iannone:
This PR upgrades any `tab_style()` `style` input that’s not in a list to be in one. This fixes the internal font list interception scheme and makes `tab_style()` more robust to different types of input. 

Fixes: https://github.com/rstudio/gt/issues/661
Fixes: https://github.com/rstudio/gt/issues/802


R/tab_create_modify.R
```diff
@@ -1036,12 +1036,33 @@ tab_style <- function(data,
   # Perform input object validation
   stop_if_not_gt(data = data)
 
-  # Intercept font styles that require registration
-  if ("cell_text" %in% names(style)) {
+  # Upgrade `style` to be within a list if not provided as such
+  if (!inherits(style, "list")) {
```
jcheng5, line 1040:
```suggestion
  if (inherits(style, "cell_styles")) {
```
The `!inherits(style, "list")` was confusing to me because `cell_styles` pass `is.list()`

tests/testthat/test-tab_style.R
```diff
@@ -473,4 +473,90 @@ test_that("using fonts in `cell_text()` works", {
...
+    dplyr::filter(ctry_origin == "United Kingdom") %>%
+    dplyr::select(mfr, model, year, hp) %>%
+    gt()
+
+  # Expect no difference in output when using styles within a list or without
+  expect_equal(
+    gtcars_tbl %>%
+      tab_style(
+        style =
+          cell_text(
```
jcheng5, line 488: This test needs a `fonts` argument with a non-single-element value

R/tab_create_modify.R
rich-iannone: Now done. Thanks!

tests/testthat/test-tab_style.R
rich-iannone: Now added in.

jcheng5 (approved):

PR #828, opened 2021-08-19
Fix google font specifying all font options

kylebutts:


kylebutts:
I realize I never signed the CLA way back when. The same issue is still here with the same fix, so this should be good to merge!

PR #847, opened 2021-11-03
Use reusable GHA workflows

schloerke:


schloerke:
This tests on R 3.4 which has issues with `gt`

rich-iannone (approved):
Looks great. Thank you.

PR #848, opened 2021-11-06
Adds `.col_names` param to `cols_label()`

### LINKED ISSUE #830: character vector for column labels?
With labeling columns I understand that both uses of `cols_label()` below are valid but isn't the second, also requiring a named list, superfluous?  

```r
gt() %>%
    cols_label(my_label = "My Label", another_one = "Another Label")

gt() %>%
    cols_label(.list= list(my_label = "My Label", another_one = "Another Label"))
```

It seems that allowing a simple character vector would be an enhancement by reducing typing.  Naturally, this would require that all columns get new labels and the character vector be equal to `ncol()` of the data frame.
```r
gt() %>%
    cols_label(col_names = c("My Label", "Another Label"))
```
Thanks.

jmbarbone:
Resolves #830 

A few checks are completed:
* `.col_names` defaults to `NULL`
* If not `NULL`, `...` and `.list` should not have `length()`
* `.col_names` must be the same length as the `colnames` from the original data
* `.col_names` should not have `names` -- gives a warning as the names would be overwritten and vector is expected to be in order

Local tests didn't show any conflicts with this

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,7 +2,7 @@
 
 ## New features
 
-* `cols_label()` gains a new argument `.col_names()` to allow for an unnamed vector (of equal length as the column names in the original data) to be used as new labels (@jmbarbone, #830)
+* `cols_label()` gains a new argument `.col_names` to allow for an unnamed vector (of equal length as the column names in the original data) to be used as new labels (@jmbarbone, #830)
 
 # gt 0.3.1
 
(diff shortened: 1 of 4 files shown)


rich-iannone:
Thanks for the PR, but after a pretty long think about this one, I decided that this might be potentially difficult for the user during iteration; columns could be moved around (by gt or through the API), and columns could get used for the stub or as groups. At that point, it's hard to keep track of the original column order for renaming. But the next best thing to this is probably the `.list` argument.

Issue #833, opened 2021-09-06
Footnotes are rendered in separate lines/paragraphs with footnotes.sep = " "


0xlevel:
I would like to put the footnotes consecutively (inline) in one cell to adhere to APA style. Therefore I set
`footnotes.sep = " "`, to override the default linebreak as stated in the man-page (repex below).
In RStudio and in the knitted HTML each footnote is embedded in a `<p>`-Tag and therefore displayed in separate lines (with the separator at the end, so the test case is fine ;) ).

Is this a bug or expected behavior?
If the latter, the help page for `tab_options` might be misleading: 
> footnotes.sep | The separating characters between adjacent footnotes in the footnotes section. The default value produces a linebreak.

Is there a way to render all footnotes consecutively in one line? — Besides using CSS to set the ".gt_footnotes" class to `display: inline`.

```r
library(gt, magrittr)
data.frame(A = 1:3) %>%
  gt() %>%
  tab_footnote(
    footnote = "This is column A.",
    locations = cells_column_labels(
      columns = A
    )
  ) %>%
  tab_footnote(
    footnote = "This is not column B.",
    locations = cells_column_labels(
      columns = A
    )
  ) %>%
  tab_options(
    footnotes.sep = " "
  )
```
Resulting html for the table footer:

```html
<tfoot>
    <tr class="gt_footnotes">
      <td colspan="1">
        <p class="gt_footnote">
          <sup class="gt_footnote_marks">
            <em>1</em>
          </sup>
           
          This is column A.
           
        </p>
        <p class="gt_footnote">
          <sup class="gt_footnote_marks">
            <em>2</em>
          </sup>
           
          This is not column B.
           
        </p>
      </td>
    </tr>
  </tfoot>
```

btw: I <3 gt! Thank you!

rich-iannone:
This is an unfortunate bug that crept in awhile back. It's marked as a critical fix so it'll be one of the next things we'll get to fixing up. Thanks for reporting this!

0xlevel:
Good news! Thank you, I am looking forward to further developments of gt. :)

Issue #593, opened 2020-05-22
Cannot use multiple `tab_source_note` in a LaTeX table (works in HTML)


vincentarelbundock:
We can use multiple calls to `tab_source_note` to insert multiple notes at the bottom of a table. These notes appear correctly in HTML, but only the first note appears in LaTeX output.

Example:

``` r
library(gt)

tab <- data.frame(x = 1:2, y = 2:3) %>%
       gt %>%
       tab_source_note('note 1') %>%
       tab_source_note('note 2')

tab %>% as_latex %>% cat
#> \captionsetup[table]{labelformat=empty,skip=1pt}
#> \begin{longtable}{cc}
#> \toprule
#> x & y \\ 
#> \midrule
#> 1 & 2 \\ 
#> 2 & 3 \\ 
#> \bottomrule
#> \end{longtable}
#> \begin{minipage}{\linewidth}
#> note 1\\ 
#> \end{minipage}
```

<sup>Created on 2020-05-22 by the [reprex package](https://reprex.tidyverse.org) (v0.3.0)</sup>

The result is correct in HTML:

![tab](https://user-images.githubusercontent.com/987057/82691932-be189e80-9c2c-11ea-86ec-f47bb91f0d57.png)

The problem is related to Line 326 of the `utils_render_latex.R` file, which only selects the first element of the list of notes. In my `gt` fork, I removed that line and it fixed the problem:

https://github.com/rstudio/gt/blob/master/R/utils_render_latex.R#L326

PR #864, opened 2022-01-18
Fixes to footer (footnotes and source notes) across all output formats

rich-iannone:
This PR aims to fix various issues with the table footer. For HTML, the `footnotes.sep` option (in `tab_options()`) is not honored. For LaTeX, more than one source note is not possible (source notes are truncated to one). For RTF, multiple source notes are rendered but they aren't separated from each other. 

Fixes: https://github.com/rstudio/gt/issues/593
Fixes: https://github.com/rstudio/gt/issues/833


R/utils_render_latex.R
```diff
@@ -365,9 +365,7 @@ create_footer_component_l <- function(data) {
     footnotes <-
       paste0(
         footnote_mark_to_latex(footnotes_tbl[["fs_id"]]),
-        footnotes_tbl[["footnotes"]] %>%
-          unescape_html() %>%
-          markdown_to_latex()
+        unlist(lapply(footnotes_tbl[["footnotes"]], process_text, context = "latex"))
```
jcheng5, line 368: Actually can you change this to a `vapply` with `character(1)`? Since that shape is actually quite important.

rich-iannone: Good idea. Now fixed!

jcheng5 (approved):

PR #866, opened 2022-01-27
Add ability to set global locale in `gt()`

### LINKED ISSUE #682: Option to set locale globally
### Setting locale for fmt_\* globally
Im using gt increasingly in html reports in Europe, therefore I would like to set the locale globally, for example via an option that switches every standard gt fmt_\* to use the locale specified in the option. 

That would make (working with) gt code in non-US locales less verbose and reduces the need to use purrr::partial

rich-iannone:
This PR adds the `locale` argument to `gt()`. The value (`NULL` or a locale string) is immutably set in the gt object's `"_locale"` key. It's validated within the `gt()` function. Formatter functions will automatically use the global setting, if any, or, will use any non-NULL `locale` value provided during within a formatter (`fmt_*()`) function (overriding the global `locale`).

Fixes #682


R/format_data.R
```diff
@@ -168,13 +170,17 @@ fmt_number <- function(data,
   # Perform input object validation
   stop_if_not_gt(data = data)
 
+  # Resolve the `locale` by choosing the default (possibly set in `gt()`) or
+  # overriding with the `locale` set here
+  locale <- resolve_locale(data = data, locale = locale)
```
jcheng5, line 175: Since `validate_locale()` always seems to follow `resolve_locale()`, consider doing the validation right in `resolve_locale()`. You could add `mustWork = TRUE` as a param if you think it's possible future uses might want to opt out of validation.

jcheng5 (dismissed):

R/format_data.R
rich-iannone: Now fixed.

jcheng5 (approved):

Issue #856, opened 2021-11-29
"transparent" for color is no longer accepted


kylebutts:
Hi -- I think an update removed the ability to use "transparent" as a color. Is this a bug or an intended change?

``` r
gt::gt(mtcars) |> 
    gt::tab_options(
        table.border.top.color = "transparent"
    )
#> Error: Problem with `mutate()` column `value`.
#> ℹ `value = { ... }`.
#> x An invalid color name was used (`"transparent"`):
#>  * Only R/X11 color names and CSS 3.0 color names can be used
```

<sup>Created on 2021-11-29 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

rich-iannone:
It was something that was considered but was put off til later because of `transparent` handling across output formats. However, I do think it should be implemented.

Issue #839, opened 2021-09-23
Unexpected color error with CSS shorthand hex


jat255:
In some existing (previously working) code, we got an error after updating GT to 0.3.1, that I think is a regression introduced by #787.

Using shorthand hex codes for colors results in an error:

```R
tab_style(
      style = list(
        cell_text(weight = "bold",
                  color = "#555")
      ),
```

Gives the following error:

```
WARN [2021-09-23 16:54:48.776] Error in : An invalid color name was used (`"#555"`):
* Only R/X11 color names and CSS 3.0 color names can be used - function (m) { logger::log_warn(m$message) }
Warning: Error in : An invalid color name was used (`"#555"`):
 * Only R/X11 color names and CSS 3.0 color names can be used
  1: shiny::runApp
```

This is easy to workaround for the time being (just change to full hex colors instead), but thought it was important to report, as those are valid CSS color identifiers.

rich-iannone:
We didn't yet implement short hexadecimal color codes but we've certainly thought about it. Seeing that it's being used makes us definitely want to make it work now. Thanks for reporting!

PR #333, opened 2019-08-01
Easier access to footnote mark options

rich-iannone:
This PR makes it easier to access the option for setting footnote marks (through the `opt_footnote_marks()` function). The internal documentation has been also been clarified as to how to provide the marks (i.e., vector or keyword).

Fixes https://github.com/rstudio/gt/issues/332.

PR #870, opened 2022-01-31
Improve color handling and add tests

rich-iannone:
These changes allow the use of shorthand hexadecimal color values (e.g, `#333`) , and, the use of the `transparent` color name.

Fixes: https://github.com/rstudio/gt/issues/856
Fixes: https://github.com/rstudio/gt/issues/839


jcheng5 (changes requested):
1. Use regex to implement short-to-long hex color conversion
2. Find out what happens if "transparent" is used with latex/rtf

jcheng5 (approved):

Issue #790, opened 2021-06-07
Add option to override `tab_source_note()`


marianschmidt:
I find it quite neat that gt allows to override (delete) existing optional table parts like headers by using existing gt objects and re-applying the `tab_*()` and setting the label to `NULL`. This comes handy for different journals sometimes requiring to have headers and footers outside the table object. However, this currently only seems to work for `tab_header()` and not for `tab_source_note()`. Would be nice to also have the ability to reset existing source notes.

``` r
library(gt)

tab_1 <-
  gtcars %>%
  dplyr::select(mfr, model, msrp) %>%
  dplyr::slice(1:5) %>%
  gt() %>%
  tab_header(title = "A simple Table",
             subtitle = "with a subtitle") %>%
  tab_source_note(
    source_note = "From edmunds.com"
  )
```

overriding headers works, but overriding tab\_source\_note does not work:

``` r
tab_1 %>%
  tab_header(title = NULL,
             subtitle = NULL) %>%
  tab_source_note(source_note = NULL)
```

<div id="gqehiatkcg" style="overflow-x:auto;overflow-y:auto;width:auto;height:auto;">

<table class="gt_table">
  
  <thead class="gt_col_headings">
    <tr>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">mfr</th>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">model</th>
      <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1">msrp</th>
    </tr>
  </thead>
  <tbody class="gt_table_body">
    <tr><td class="gt_row gt_left">Ford</td>
<td class="gt_row gt_left">GT</td>
<td class="gt_row gt_right">447000</td></tr>
    <tr><td class="gt_row gt_left">Ferrari</td>
<td class="gt_row gt_left">458 Speciale</td>
<td class="gt_row gt_right">291744</td></tr>
    <tr><td class="gt_row gt_left">Ferrari</td>
<td class="gt_row gt_left">458 Spider</td>
<td class="gt_row gt_right">263553</td></tr>
    <tr><td class="gt_row gt_left">Ferrari</td>
<td class="gt_row gt_left">458 Italia</td>
<td class="gt_row gt_right">233509</td></tr>
    <tr><td class="gt_row gt_left">Ferrari</td>
<td class="gt_row gt_left">488 GTB</td>
[...]

PR #875, opened 2022-02-03
Allow use of `NULL` to remove any previously set source notes and footnotes

rich-iannone:
This will allow `NULL` to be valid input in `tab_source_note()` and `tab_footnote()`. In these, `NULL` isn't already used to signify a missing option, and, this change will bring the behavior closer to that of `tab_header()`.

Fixes: https://github.com/rstudio/gt/issues/790

### DIFF
--- a/R/dt_source_notes.R
+++ b/R/dt_source_notes.R
@@ -18,6 +18,13 @@ dt_source_notes_init <- function(data) {
 
 dt_source_notes_add <- function(data, source_note) {
 
+  # There is the option to remove all source notes with
+  # `tab_source_note(source_note = NULL)`; this will
+  # essentially reinitialize the list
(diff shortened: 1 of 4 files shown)


jcheng5:
After discussing with @rich-iannone--I don't like this approach with functions where repeated calls result in additive effects (i.e. calling `tab_footnote` multiple times on a cell results in multiple footnotes, not just the most recent footnote). It's surprising to me that if you pass in a `NULL`, suddenly it affects all the previous calls as well. I'd prefer a `tab_clear_footnote` or something--or maybe `tab_footnote(replace = TRUE)`?

rich-iannone:
Closing this in favor of the better approach afforded by https://github.com/rstudio/gt/pull/1073.

Issue #827, opened 2021-08-19
Missing values not supported in `summary_rows()`?


eutwt:
It looks like a summary row value can't be missing, even though missing values are displayed correctly (without error or warning) as part of a normal column. Is this known/expected?

``` r
library(gt)

example <- gt(data.frame(a = c(1, NA, 3)))

summary_rows(example, columns = a, fns = list(~ sum(.)))
#> Error: Problem with `mutate()` column `a`.
#> ℹ `a = (function (x) ...`.
#> ℹ `a` must be size 1, not 0.
```

<sup>Created on 2021-08-19 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

rich-iannone:
This is definitely not something I'm happy with and I do think it's a bug. Workaround is to use `summary_rows(example, columns = a, fns = list(~ sum(., na.rm = TRUE)))` but it's not ideal. Thanks for reporting this.

Issue #471, opened 2020-01-17
grand_summary_rows errors on all NA columns


smingerson:
latest from github (GithubSHA1: 883df9ad23e49e7e7ede9ea3cce91940fc59f189)  

My use case is an empty table showing what information will be displayed. The end user will enter info in a Shiny app and then the table will update. My workaround will be to populate the table with 0s. Edit: this also happens with 0. The error happened in the print method, when it is applying the functions to the columns.

```r
library(gt)
grand_summary_rows(gt(data.frame(a = c(1,2), b = NA_real_)), columns = vars(b),
                   fns = list(max = ~max(.)))
#> Error: Column `b` must be length 1 (the number of rows), not 0
```
Created on 2020-01-17 by the reprex package (v0.3.0)

rich-iannone:
Thanks for reporting this and sorry for waiting until now to respond. I'm able to reproduce the error in my environment and I do believe this requires fixing.

jkgrain:
I am running into the same issue.  The following reprex might help trace the problem.  I think it has to do with the mean of a vector that is all NA returns NaN even with na.rm=TRUE.
`
library(dplyr)
library(gt)

x <- c(1:6)
y <- c(6:2, NA)
z <- rep(NA, 6)

df <- tibble(x,y,z)

df <- mutate_if(df, is.logical, as.numeric)

df %>%
  gt() %>%
  grand_summary_rows(
    columns = c(1:3),
    fns = list(
      Avg = ~ mean(., na.rm = T)
    )
  ) 

mean(df$x, na.rm = TRUE)
mean(df$y, na.rm = TRUE)
mean(df$z, na.rm = TRUE)
`

jkgrain:
I was able to trap for the problematic column with this variation on the code:

```
library(dplyr)
library(gt)

x <- c(1:6)
y <- c(6:2, NA)
z <- rep(NA, 6)

df <- tibble(x,y,z)

df %>%
  gt() %>%
  grand_summary_rows(
    columns = c(1:3),
    fns = list(
      Avg = ~ mean(., na.rm = T)
    )
  ) 

df %>%
  gt() %>%
  grand_summary_rows(
    columns = which(sapply(df, class) != "logical"),
    fns = list(
      Avg = ~ mean(., na.rm = T)
    )
  ) 
```

PR #887, opened 2022-02-09
Allow summary rows to be generated in the case of all-`NA` columns

rich-iannone:
This PR allows for aggregations that might return an `NaN` (common case is using `mean(., na.rm = TRUE)` on an all-`NA`column) to be changed to `NA`. This removes the evaluation error that would otherwise occur and provides an opportunity to replace the `NA` with an explanatory replacement string.

Fixes: https://github.com/rstudio/gt/issues/471
Fixes: https://github.com/rstudio/gt/issues/827


R/dt_summary.R
```diff
@@ -251,7 +251,13 @@ dt_summary_build <- function(data,
               context = context
             )
 
-          formatter(x)
+          formatted <- formatter(x)
```
jcheng5, line 254: As currently written, the options from the main gt table (locale, etc.) won't be respected by this formatter call, because we're using a temporary gt table.

jcheng5 (changes requested):
Unit test that:

1. Ensures that summary functions returning vectors of length != 1 should always result in error
2. Test that NA and NaN returned from summary functions is fine, and render as missing text

R/dt_summary.R
rich-iannone: Now changed.

jcheng5 (approved):

Issue #884, opened 2022-02-08
Enhance `cols_merge_uncert()` so that different upper and lower bounds can be used


rich-iannone:
The `cols_merge_uncert()` function works nicely in cases where the uncertainty can be expressed by a single value after a plus-minus sign. However, there are cases where uncertainties differ in the positive direction and the negative direction.

To express properly this we need to extend the function to allow examination of two columns for uncertainty values (lower bound and upper bound to uncertainty), check if they are different, and use a different notation for expressing such uneven uncertainty values. This is typeset with superscripted and subscripted values for the bounds with prominent plus and minus signs. Both uncertainty values must begin at the same position (just after the measurement value) but have enough vertical space between them.

PR #888, opened 2022-02-10
Enhance the `cols_merge_uncert()` function

rich-iannone:
The `cols_merge_uncert()` function works nicely in cases where the uncertainty can be expressed by a single value after a plus-minus sign. However, there are cases where uncertainties differ in the positive direction and the negative direction and this PR expands the scope of `cols_merge_uncert()`.

Fixes: https://github.com/rstudio/gt/issues/884


inst/css/gt_styles_default.scss
```diff
@@ -366,6 +366,14 @@
     font-size: 65%;
   }
 
+  .gt_two_val_uncert {
+    display: inline-block;
+    line-height: 1em;
+    text-align: right;
+    font-size: 60%;
+    vertical-align: -0.25em;
```
jcheng5, line 374: Add left margin to this class, and dispose of the outer span

rich-iannone: Now done.

R/utils_render_common.R
```diff
@@ -327,18 +325,89 @@ perform_col_merge <- function(data,
...
+
+        pattern_unequal <- "<<1>>(+<<3>>, -<<2>>)"
+      }
+
+      # Determine rows where NA values exist
+      na_1_rows <- is.na(data_tbl[[mutated_column]])
+      na_lu_rows <- is.na(data_tbl[[lu_column]])
+      na_uu_rows <- is.na(data_tbl[[uu_column]])
+      na_lu_or_uu <- na_lu_rows | na_uu_rows
+      lu_equals_uu <- data_tbl[[lu_column]] == data_tbl[[uu_column]] & !na_lu_or_uu
```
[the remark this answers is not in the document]
rich-iannone: Decided to pass `NA`s straight though because it's the most noticeable and least invasive way to go. Plus, people can modify `NA` with `fmt_missing()` or `text_transform()` so the missing values may be meaningful/useful.

R/utils_render_common.R
```diff
@@ -355,8 +357,7 @@ perform_col_merge <- function(data,
 
         pattern_unequal <-
           paste0(
-            "<<1>><span style=\"margin-left: 0.1em;\">",
-            "<span class=\"gt_two_val_uncert\">",
+            "<<1>><span><span class=\"gt_two_val_uncert\">",
```
jcheng5, line 360: Let's remove the outer span entirely

R/utils_render_common.R
```diff
@@ -376,10 +377,12 @@ perform_col_merge <- function(data,
       na_lu_rows <- is.na(data_tbl[[lu_column]])
       na_uu_rows <- is.na(data_tbl[[uu_column]])
       na_lu_or_uu <- na_lu_rows | na_uu_rows
+      na_lu_and_uu <- na_lu_rows & na_uu_rows
+      na_one_of_lu_uu <- na_lu_or_uu & !na_lu_and_uu
```
jcheng5, line 381: Seems like this one can be removed

jcheng5 (approved):

R/utils_render_common.R
rich-iannone: Done. Thanks!

R/utils_render_common.R
rich-iannone: Now removed.

PR #897, opened 2022-02-15
Allow for footnote placement options in `tab_footnote()`

rich-iannone:
This PR enhances `tab_footnote()` by adding a `placement` argument. This provides the user with options for how to affix footnote marks to the table content. Two fixed options for this are `"left` or `"right"`, where the placement is to the absolute left or right of the cell content. The default option is `"auto"` whereby **gt** will choose a preferred left-or-right placement depending on the alignment.


R/utils.R
```diff
@@ -210,6 +210,118 @@ get_currency_str <- function(currency,
...
+resolve_footnote_placement <- function(data,
+                                       colname,
+                                       rownum,
+                                       input_placement,
+                                       cell_content,
+                                       context) {
+
+  if (input_placement %in% c("left", "right")) {
+    return(input_placement)
+  }
```
jcheng5, line 222: Throw if `input_placement` is not "auto"?

R/utils.R
```diff
@@ -210,6 +210,118 @@ get_currency_str <- function(currency,
...
+      rownum = rownum,
+      context = context
+    )
+
+  # If cell alignment has been set then always choose 'right' placement
+  if (!is.null(names(cell_alignment)) && names(cell_alignment) == "user_set") {
+    return("right")
+  }
+
+  if (cell_alignment == "right") {
```
jcheng5, line 237: Let's just say auto means mark placement is opposite of cell alignment.

rich-iannone: Done. However, now there needs to be some followup work done to get the alignments of the summary cells.

rich-iannone: Done.

R/utils.R
```diff
@@ -210,6 +210,102 @@ get_currency_str <- function(currency,
...
+    if (nrow(styles_tbl) < 1) {
+      return(column_alignment)
+    }
+
+    html_styles <-
+      vapply(
+        styles_filtered_tbl$styles,
+        FUN.VALUE = character(1),
+        USE.NAMES = FALSE,
+        FUN = styles_to_html
```
jcheng5, line 285: I'm curious, why do you render the styles to html instead of looking at `styles_filtered_tbl$styles` directly?

rich-iannone: For a statement that's a bit more complicated like this: 

```r
exibble %>%
  gt() %>%
  fmt_number(num) %>%
  tab_footnote(footnote = "note", locations = cells_body(group, 1)) %>%
  tab_style(style = cell_text(align = "right"), locations = cells_body(group, 1)) %>%
  tab_style(style = cell_text(size = "smaller", align = "left"), locations = cells_body(group, 1)) %>%
  tab_style(style = "text-align: right; background: green;", locations = cells_body(group, 1))
```

The final alignment for the cell is `right` because `html_styles` will produce `"text-align: left; font-size: smaller; text-align: right; background: green;"`. That said, this code in the PR still gets this wrong because of the final `gsub()` statement.

I will rewrite this to use `styles_filtered_tbl$styles` because that is a better strategy anyway.

rich-iannone: Now rewritten, and many tests were added.

R/utils.R
```diff
@@ -210,6 +210,110 @@ get_currency_str <- function(currency,
...
+  # This is a user-defined string with CSS style rules that should look
+  # something like this: "text-align: right; background: green;"
+  cell_style <- cell_styles_list[[1]]$cell_style
+
+  # Return the value of the last `text-align` property, if present
+  if (!is.null(cell_style) && grepl("text-align:", cell_style)) {
+
+    style_vec <- rev(unlist(strsplit(cell_style, ";\\s+")))
+    is_text_alignment <- grepl("text-align", style_vec)
+    cell_alignment <- gsub("text-align:\\s+|;", "", style_vec[is_text_alignment])[1]
```
jcheng5, line 304: Alternate parsing:

```
str <- "text-align: right !important; background: green; text-align: left;"
m <- gregexec("(?:^|;)\\s*text-align\\s*:\\s*([\\w-]+)\\s*(!important)?", str, perl = TRUE)
regmatches(str, m)
# Pick last !important, or if no !important, then last anything
```

rich-iannone: Now changed. Thanks!

jcheng5 (approved):

Issue #911, opened 2022-03-11
Add the `sub_large_vals()` function


rich-iannone:
The `sub_large_vals()` function should be added as a way to substitute values above a certain cutoff (either positive or negative values) with specified text.

Issue #910, opened 2022-03-11
Add the `sub_small_vals()` function


rich-iannone:
The `sub_small_vals()` function should be added as a way to substitute values below a certain cutoff (either positive or negative values) with specified text.

Issue #904, opened 2022-03-04
Add the `sub_zero()` function


rich-iannone:
We ought to add a function to substitute zero values with something else. We’ll call it `sub_zero()`:

![](http://2.bp.blogspot.com/--fzvdSxz5Fc/VQjud1kh_rI/AAAAAAAANpw/1PaN7MGnBv8/s1600/Sub-Zero%2B4.gif)

PR #912, opened 2022-03-14
Add small collection of `sub_*()` functions

rich-iannone:
The `sub_*()` group of functions are meant to make simple and common substitutions of data in a **gt** table with not-too-much effort. The functions added are:

- `sub_missing()`
- `sub_zero()`
- `sub_small_vals()`
- `sub_large_vals()`

The `sub_missing()` function will replace `fmt_missing()` (deprecating the latter function).

Fixes https://github.com/rstudio/gt/issues/904
Fixes https://github.com/rstudio/gt/issues/910
Fixes https://github.com/rstudio/gt/issues/911


jcheng5:
Changes requested:

1. Add `prepend=FALSE` argument to `fmt`. If `TRUE`, formatter will be added to the front of the existing list of formatters instead of the end (last one wins instead of first one wins).
2. Introduce `fmt_replace` that just takes static values. e.g. `fmt_replace(value = list(html = "<b>hi</b>", default = "hi"))`
3. `sub_` functions will call `fmt_replace(prepend=TRUE)` instead of having a separate list

jcheng5:
Note to self, advantages of `sub_small_vals` over `fmt_replace` are:

1. No need to know about `prepend = TRUE/FALSE`
2. Can apply to multiple columns in one call
3. Can easily use fancy context-specific symbols

R/substitution.R
```diff
@@ -0,0 +1,609 @@
...
+#' 3-18
+#'
+#' @import rlang
+#' @export
+sub_small_vals <- function(
+    data,
+    columns = everything(),
+    rows = everything(),
+    threshold = 0.01,
+    small_pattern = "<{x}",
```
jcheng5, line 321:
```suggestion
    small_pattern = if (sign == "+") "<{x}" else md("<*abs*({x})"),
```

R/substitution.R
```diff
@@ -0,0 +1,609 @@
...
+#' @section Function ID:
+#' 3-16
+#'
+#' @import rlang
+#' @export
+sub_missing <- function(
+    data,
+    columns = everything(),
+    rows = everything(),
+    missing_text = "---"
```
jcheng5, line 54:
```suggestion
    missing_text = md("---")
```

and replace `context_missing_text` with `process_text`

R/substitution.R
```diff
@@ -0,0 +1,609 @@
...
+          process_text(
+            context_small_vals_text(
+              threshold = threshold,
+              small_pattern = small_pattern,
+              sign = sign
+            ),
+            context = "html"
+          )
+            ,
+          NA_character_
```
jcheng5, line 366: Extract out a common (inner) function

R/substitution.R
```diff
@@ -0,0 +1,609 @@
...
+    columns = {{ columns }},
+    rows = {{ rows }},
+    prepend = TRUE,
+    fns = list(
+      # Any values of `x` that are above the threshold will be processed
+      # according to the `large_pattern`, the `threshold` value (interacts with
+      # the `large_pattern`, and the sign (changes the default `large_pattern`))
+      html = function(x) {
+        ifelse(
+          is.numeric(x) &
```
jcheng5, line 543: This `&` doesn't short-circuit. Let's perform a separate check for `is.numeric(x)` and return `rep_len(NA, length(x))`

R/substitution.R
rich-iannone: After testing, found that this cannot be done.

R/substitution.R
rich-iannone: I used this minor variation: `if (sign == "+") "<{x}" else md("<*abs*(-{x})")`

R/substitution.R
rich-iannone: Now done. Thanks for this!

R/substitution.R
rich-iannone: Now done.

R/substitution.R
```diff
@@ -348,67 +375,59 @@ sub_small_vals <- function(
...
-          context_small_vals_text(
-            threshold = threshold,
-            small_pattern = small_pattern,
-            sign = sign
-          ),
-          NA_character_
+
+        if (!is.numeric(x)) {
+          return(rep_len(NA_character_, length(x)))
+        }
```
jcheng5, line 395: Move these inside the common function

R/substitution.R
```diff
@@ -527,6 +546,33 @@ sub_large_vals <- function(
...
+  sub_replace_large_vals <- function(
+    x,
+    threshold,
+    large_pattern,
+    sign,
+    context
+  ) {
+    ifelse(
+      !is.na(x) &
+        x != 0 &
```
jcheng5, line 561:
```suggestion
```

R/substitution.R
```diff
@@ -539,57 +585,59 @@ sub_large_vals <- function(
...
-              sign = sign,
-              context = "html"
-            ),
-            context = "html"
-          ),
-          NA_character_
+
+        if (!is.numeric(x)) {
+          return(rep_len(NA_character_, length(x)))
+        }
```
jcheng5, line 591: Also move

jcheng5 (approved):

Issue #879, opened 2022-02-04
Allow adding footnotes without a location parameter


rich-iannone:
### Discussed in https://github.com/rstudio/gt/discussions/843

<div type='discussions-op-text'>

<sup>Originally posted by **JohannesNE** October 27, 2021</sup>
It would be useful to add footnotes with `tab_footnote()` without specifying a location.
This should just add an entry to the footnote text without a superscript number.

My specific use case is to explain a symbol used in the table.

`tab_source()` can nearly do this, but it is displayed in a separate section</div>

PR #925, opened 2022-04-15
Allow addition of footnotes without supplying any `locations`

rich-iannone:
We sometimes find that notes in the footnotes section can be without reference to any particular part of the table. Most often, these notes without marks precede those that do contain marks (it would be confusing to mix the marked and unmarked notes together in any other way, especially if they appear in paragraph form).

This PR makes it so that the default locations parameter is `NULL` and a `footnote` (still a required argument) will have no reference to a table cell. The order of `tab_footnote()` calls is still preserved but is split between location-based footnotes and the non-location variety (which always appears first).

Here is an example:

```r
library(gt)

exibble %>%
  gt() %>%
  tab_footnote(footnote = "A footnote.") %>%
  tab_footnote(
    footnote = "A location note. Appears after the first two notes",
    locations = cells_body(1, 1)
  ) %>%
  tab_footnote(footnote = "A second footnote.") %>%
  opt_footnote_marks(marks = letters) %>%
  tab_options(footnotes.multiline = FALSE)
```

tab_footnote_pr

Fixes: https://github.com/rstudio/gt/issues/879


R/utils_render_latex.R
```diff
@@ -10,7 +10,15 @@ latex_packages <- function() {
...
-  paste0("\\textsuperscript{", mark, "}")
+
+  vapply(
+    mark,
+    FUN.VALUE = character(1),
+    USE.NAMES = FALSE,
+    FUN = function(x) {
+      if (is.na(x)) "" else paste0("\\textsuperscript{", x, "}")
+    }
+  )
```
jcheng5, line 21:
```suggestion
  ifelse(is.na(mark), "", paste0("\\textsuperscript{", mark, "}"))
```

jcheng5 (approved):

PR #963, opened 2022-06-15
Improved errors and warnings with `cli::cli_[abort|warn]()`

rich-iannone:
With this PR, we are aiming to obtain improved formatting of errors and warnings by using the `cli` framework (`cli::cli_[abort|warn]()`). It's going to be a virtually 1:1 match to the older formatting but the styling improvements will lead to an overall better consistency in the messaging.


cscheid (approved):
lgtm!

PR #977, opened 2022-07-01
[Accessibility] Closing #679: Use <td> instead of <th> for title and subtitle

jooyoungseo:
This PR addresses #679: now screen readers don't get confused when using title and subtitle element.

## Reprex

``` r
library(gt)

# Create gt table with title and subtitle
airquality[1:10, ] %>%
  gt::gt() %>%
  gt::tab_header(
    title = "Title",
    subtitle = "Subtitle"
  )
```

## Results

`title` and `subtitle` now get treated as `<td>`. This is good for screen reader a11y.

@rich-iannone, I still need your help to verify whether there are any CSS tweaks that we need to consider. I don't want to introduce any visual breaks.

### DIFF
--- a/R/utils_render_html.R
+++ b/R/utils_render_html.R
@@ -287,7 +287,7 @@ create_heading_component_h <- function(data) {
 
   title_row <-
     htmltools::tags$tr(
-      htmltools::tags$th(
+      htmltools::tags$td(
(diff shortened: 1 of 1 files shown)


rich-iannone (approved):
lgtm

PR #979, opened 2022-07-02
[Accessibility] Closing #678: Use <th scope="row"> for gt_stub

jooyoungseo:
This PR addresses #678.

Although this works nicely for screen readers, we need to verify whether this PR needs any CSS tweaks.


rich-iannone (approved):
lgtm!

Issue #885, opened 2022-02-08
Saving table as PNG doesn't keep some font styling (i.e. bold) 


gponce-ars:
## Description

Running the example below (extracted from the documentation) you can see that the PNG does not keep the strong/bold style in the header as you expect it and see it in the HTML output.

## Reproducible example
```
library(gt)
gt_table<- exibble %>% 
              gt() %>% 
              tab_spanner(
                          label = md("**Testing style**"),
                          columns = c(datetime, currency, row)
                  )
  
gtsave(gt_table, filename = '/home/devfe/test.png')
gtsave(gt_table, filename = '/home/devfe/test.html')

```

## Expected result

A PNG showing the table with the expected font style.

## Session info
```
R version 4.1.0 (2021-05-18)
Platform: x86_64-apple-darwin13.4.0 (64-bit)
Running under: macOS Big Sur 10.16

Matrix products: default
BLAS/LAPACK: /Users/gponce/miniconda3/envs/python39/lib/libopenblasp-r0.3.15.dylib

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base

other attached packages:
[1] gt_0.3.1

loaded via a namespace (and not attached):
 [1] magrittr_2.0.1    tidyselect_1.1.1  munsell_0.5.0     colorspace_2.0-2
 [5] R6_2.5.1          rlang_0.4.12      fansi_1.0.2       dplyr_1.0.7
 [9] tools_4.1.0       grid_4.1.0        checkmate_2.0.0   gtable_0.3.0
[13] utf8_1.2.2        DBI_1.1.1         htmltools_0.5.1.1 ellipsis_0.3.2
[17] assertthat_0.2.1  digest_0.6.29     tibble_3.1.6      lifecycle_1.0.1
[21] crayon_1.4.2      purrr_0.3.4       ggplot2_3.3.5     sass_0.4.0
[25] vctrs_0.3.8       glue_1.6.1        compiler_4.1.0    pillar_1.6.4
[29] backports_1.4.1   generics_0.1.1    scales_1.1.1      pkgconfig_2.0.3
```

Issue #621, opened 2020-07-14
Add ability to use webshot2 when it lands on CRAN?


stephhazlitt:
I ran into a bug with `webshot` when saving a gt table to `png` with `gtsave()` &mdash; the **bold formatting** was lost. I found this [similar issue](https://github.com/davidgohel/flextable/issues/133) and managed to export the table with the bold formatting by saving as an `.html` and then using `webshot2::webshot` to render the `png`. 

---

1) make a nice gt table with bolded title row
``` r
library(gt)
library(dplyr)

test_table <- countrypops %>%
  filter(year == "2017") %>% 
  slice_max(population, n = 10) %>% 
  select(country_name, population) %>% 
  gt() %>% 
   tab_style(
    style = cell_text(weight = "bold"),
    locations = cells_body(
      columns = vars(country_name, population),
      rows = population == max(population))
    )
test_table
```

2) save to `png`, but sadly no bold title row for me 😢 
```r
gtsave(test_table,
       filename = "test_table.png")
```

3) save to `html` and used webshot2 to save to `png` 🎉 

```r
gtsave(test_table, 
        filename = "test_table.html")

webshot2::webshot("test_table.html", "test_table.png")
```

This issue might be something do to with my computing environment and `webshot` however just placing this here in case this might be is a useful enhancement for `gtsave()` to be able to leverage `webshot2` when it hits CRAN.

SessionInfo() 👇 

```
R version 4.0.2 (2020-06-22)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Mojave 10.14.6

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib
LAPACK: /Library/Frameworks/R.framework/Versions/4.0/Resources/lib/libRlapack.dylib

locale:
[1] en_CA.UTF-8/en_CA.UTF-8/en_CA.UTF-8/C/en_CA.UTF-8/en_CA.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] testthat_2.3.2 devtools_2.3.0 usethis_1.6.1 

loaded via a namespace (and not attached):
 [1] Rcpp_1.0.5          pillar_1.4.6        compiler_4.0.2     
 [4] later_1.1.0.1       prettyunits_1.1.1   remotes_2.1.1      
[...]

PR #985, opened 2022-07-07
Use **webshot2** for generating image files with `gtsave()`

rich-iannone:
This changes the image-generating implementation of `gtsave()` when specifying a PNG or PDF. Here, **webshot2** will be used instead of **webshot** and this will provide higher-fidelity images (due to better rendering with headless Chrome compared to PhantomJS). This change also allows for customization of the `selector` option during invocation of the `webshot2::webshot()` function (defaulting to `"table"`, which is the same as previously).

Fixes: https://github.com/rstudio/gt/issues/621
Fixes: https://github.com/rstudio/gt/issues/885


cscheid (approved):
Assuming tests pass, this is good.

PR #987, opened 2022-07-09
Add a function to generate styled tables from presets

rich-iannone:
This PR adds the `opt_stylize()` function to enable the styling of a table based on a numbered style and color.

Fixes: #238


R/opts.R
```diff
@@ -917,6 +917,299 @@ opt_table_font <- function(
...
+preset_colors <-
+  list(
+    gray =    c("#F4F4F4", "#D5D5D5", "#929292", "#5F5F5F", "#000000"),
+    blue =    c("#EDF7FC", "#89D3FE", "#00A1FF", "#0076BA", "#004D80"),
+    orange =  c("#FFFFDB", "#FEFC66", "#FAE231", "#F8BA00", "#FF9400"),
+    aqua =    c("#EBFBF9", "#A5FEF2", "#14E6CF", "#01A99D", "#007C77"),
+    pink =    c("#FCF2F7", "#FF8CC6", "#EF5FA7", "#CB2A7B", "#99195F"),
+    green =   c("#EDF6E8", "#9AFD69", "#60D937", "#1EB100", "#027101"),
+    red =     c("#FEEDEC", "#FF968D", "#FF644E", "#ED220D", "#B51800")
+  )
```
cscheid, line 1057: Would it make sense to document where these came from?

R/opts.R
```diff
@@ -917,6 +917,299 @@ opt_table_font <- function(
...
+      location_hlines_color = get_all_preset_colors_at_index(4),
+      column_labels_background_color = "#FFFFFF",
+      stub_background_color = get_all_preset_colors_at_index(4),
+      stub_border_style = "solid",
+      stub_border_color = get_all_preset_colors_at_index(4),
+      data_hlines_style = "none",
+      data_hlines_color = get_all_preset_colors_at_index(2),
+      data_vlines_style = "none",
+      data_vlines_color = get_all_preset_colors_at_index(2),
+      summary_row_background_color = "#FFFFFF",
```
cscheid, line 1088: I wonder if there's a design that plays somewhat better with CSS and other format-specific styling infra. (`#ffffff` background will instantly break with dark mode or any other background CSS setting..)

R/opts.R
```diff
@@ -917,6 +917,299 @@ opt_table_font <- function(
...
+      data_hlines_style = "none",
+      data_hlines_color = preset_colors$gray[[2]],
+      data_vlines_style = "none",
+      data_vlines_color = preset_colors$gray[[2]],
+      summary_row_background_color = "#FFFFFF",
+      grand_summary_row_background_color = preset_colors$gray[[2]],
+      row_striping_background_color = get_all_preset_colors_at_index(1),
+      table_outline_color = NA_character_
+    )
+  )
```
cscheid, line 1193: ## general design

It's possible I'm totally missing something, but I'm wondering about this design:

- where did these styles come from?
- can they be extended?

## performance

In addition, is this a sequence of `bind_rows` each creating a single-row tibble? This predetermined table gets initialized every time we call `library(gt)`, is that right? If so, does it make sense to have a script during the build step of the package that creates a .csv file or something like it and then we access that directly at runtime when needed? I'm sure this one initialization step won't take a lot of time by itself, but add a couple of dozen of these and we're now talking about maybe tenth of a second of loading time for every single `gt` user out there, regardless of whether they use this or not.

cscheid (commented):
I have a couple of questions this time.

R/opts.R
rich-iannone: I will document this in the `X09-styles_colors_params.R` script (in `data-raw`). Basically I chose some base colors, made one darker variation of each and then three lighter variations. This was in Sketch for the most part:

color_variations_sketch

I'll put in some notes w.r.t. the color metrics for each (lightness, contrast with black/white text, etc.) in that file soon.

rich-iannone: Also, I changed the color name `aqua` to `cyan`. It was closer to `aquamarine` than `aqua` but `cyan` is even closer.

R/opts.R
rich-iannone: Do you know of any testing tools to get a handle on this? Definitely interested in accommodating dark mode with the color choices.

rich-iannone:
One thing I need to do is check contrast levels (with white text on dark backgrounds). I don't think I have enough contrast in some cases and I'd like to pass all the checks at https://webaim.org/resources/contrastchecker/ before finalizing this.

R/opts.R
rich-iannone: Styles are all in a Sketch doc I put together quite a while back. With this iteration they can't be extended by the user but I'd like to make that possible in a follow-up PR.

cscheid:
> One thing I need to do is check contrast levels (with white text on dark backgrounds). I don't think I have enough contrast in some cases and I'd like to pass all the checks at https://webaim.org/resources/contrastchecker/ before finalizing this.

Cool, that's great. Please ask @batpigandme about a11y. It's a good important thing to check for those, but we should be using the same tooling across the org (it's possible you have already, in which case no need to check again!)

rich-iannone:
Hey @batpigandme , I've been working on this PR for the `opt_stylize()` function that adds a few themes with various color combinations and I could use your expertise on the matter.

The function essentially styles a **gt** table using a base `color` and a `style` preset (any number from 1 to 6). With the output tables I was able to visually check all of the resulting color combinations of computed text color with their matching background colors and then go a bit further with some contrast checking tools. Here's an example of how to produce a table with the new theme feature (and a screen capture of the table is below):

```r
exibble %>%
  summary_rows(
    groups = "grp_a",
    columns = c(num, currency),
    fns = list(
      min = ~min(., na.rm = TRUE),
      max = ~max(., na.rm = TRUE)
    )) %>%
  grand_summary_rows(
    columns = currency,
    fns = list(
      total = ~sum(., na.rm = TRUE)
    )) %>%
  tab_source_note(source_note = "This is a source note.") %>%
  tab_footnote(
    footnote = "This is a footnote.",
    locations = cells_body(columns = 1, rows = 1)
  ) %>%
  tab_header(
    title = "The title of the table",
    subtitle = "The table's subtitle"
  ) %>%
  opt_stylize(style = 6, color = "cyan")
```

opt_stylize_blue_1

I was able to manually check every combination of text and background color using the following two things: 

- the WebAIM Contrast Checker tool (at https://webaim.org/resources/contrastchecker): I required a WCAG AA passing
grade
- the Chrome Dev Tools accessibility indicator for contrast: I required a passing checkmark

Is this enough? Or did I miss something, perhaps a better piece of tooling? Happy to show all of this to you in a video call anytime.

batpigandme:
Great job, Rich! If the color contrasts pass those, you're good to go!

rich-iannone:
Thanks, @batpigandme ! I feel much better about eventually merging this in now. 

@cscheid would you be able to re-review (when you’re back, that is) and possibly approve?

cscheid (approved):
lgtm

Issue #926, opened 2022-04-19
[accessibility] Assign `id`s to summary row heads (`<th>`) and apply as `headers` to `<td>`s


batpigandme:
Expansion of https://github.com/rstudio/gt/issues/678 as it applies to, e.g. summary lines (as in the examples in the [creating-summary-lines vignette](https://gt.rstudio.com/articles/creating-summary-lines.html) where you're creating groupwise summary lines). 

These will constitute tables with multi-level headings, and need the `<th>`s with each scope (`col`, `row`, `colgroup`, `rowgroup`) to have `id`s that are applied to the individual `<td>` cells.

The examples in the [W3C WAI Tables with Multi-Level Headers](https://www.w3.org/WAI/tutorials/tables/multi-level/) (particularly Example 2) match nicely with the structure you have for the groupwise summary lines (see [full code for Complete “Table with three headers related to each data cell” Example](https://www.w3.org/WAI/tutorials/tables/examples/threeheaders/)). 

Wanted to file this now, since it might be useful for the `<th>` elements to get ids by default, so that they can be applied to individual cells for multi-level-header tables, and [tables with irregular headers](https://www.w3.org/WAI/tutorials/tables/irregular/).

**Edit** Sorry about filing as bug rather than enhancement! 😳  I can't change it after the fact.

rich-iannone:
Thanks for filing the issue and providing lots of detail! This and the small collection of other accessibility issues will be worked on for the `v0.6.0` release.

PR #988, opened 2022-07-14
[Accessibility] closing #926

jooyoungseo:
- [x] Replaced `<td>` with `<th>` for summary_rows and group_rows.
- [x] Assigned `id` attribute to each of the table headers denoted with `<th>`.
- [x] Assign `headers` attribute to regular table cells denoted with `<td>` to semantically be associated with corresponding headers.

Fixes: #926


jooyoungseo:
@rich-iannone  -- Still `row_id` does not change its value. Why do we need a new variable, `col_id`? Isn't it the same as `as.character(x)`?

jooyoungseo:
@rich-iannone  -- Test with the following code after installing this PR:

## Reprex
``` r
library(tidyverse)
library(gt)

exibble %>% gt()


# Create a gt table using the `exibble` dataset
exibble_a <-
    exibble %>%
    select(-c(fctr, date, time, datetime)) %>%
    sub_missing()



# Create groupwise summary rows for both
# groups (using `groups = TRUE`); use the
# `mean()`, `sum()`, and `sd()` functions
# (only for the `num` column)
exibble_b <-
    exibble_a %>%
    summary_rows(
        groups = TRUE,
        columns = num,
        fns = list(
            average = "mean",
            total = "sum",
            SD = "sd"
        )
    )

exibble_b
```

rich-iannone (approved):
LGTM!

PR #990, opened 2022-07-21
Adjust family names to fit with .Rd templating

rich-iannone:
Each exported function in gt is part of a family and given a family name (with the `@family` roxygen tag). When generated as .Rd files the family names are incorporated like this in the `See Also` help section:

`Other <family_name>: ...[.Rd links to other functions in the family]...`

Previously, the family names were written as standalone titles but they really need to be written to fit within this structure (lower case, preferably ending with 'functions'). This PR doesn't change familiar membership, it just modifies all of the family names in the source .R files (and the .Rd files were rebuilt).

### DIFF
--- a/R/data_color.R
+++ b/R/data_color.R
@@ -126,7 +126,7 @@
 #' `r man_get_image_tag(file = "man_data_color_2.png")`
 #' }}
 #'
-#' @family Format Data
+#' @family data formatting functions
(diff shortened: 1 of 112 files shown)


cscheid (approved):
112 files! Was this secretly APM training for SC2 or something?

(lgtm :) )

Issue #991, opened 2022-07-21
Ensure that spanner ID can't take a column name


rich-iannone:
Currently, we use `check_spanner_id_unique()` to validate whether an ID value for a new column spanner is unique. The uniqueness check only considers the closed set of existing column spanner ID values. However, we need to ensure that the column spanner IDs are unique across all cells that use the `<th>` tag (this includes column names and spanner column names).

This is required to unblock issue #926 since ID values for all `<th>` tags need to be unique for the purpose of linking the column and row headers.

cc @jooyoungseo

PR #992, opened 2022-07-22
Ensure that unique IDs are maintained across columns spanners and column headers

rich-iannone:
Currently, we use `check_spanner_id_unique()` to validate whether an ID value for a new column spanner is unique. The uniqueness check only considers the closed set of existing column spanner ID values. The changes here ensure that the column spanner IDs are unique across all cells that use the `<th>` tag (this includes column names and spanner column names).

Fixes: #991

### DIFF
--- a/R/tab_create_modify.R
+++ b/R/tab_create_modify.R
@@ -200,7 +200,8 @@ tab_spanner <- function(
     return(data)
   }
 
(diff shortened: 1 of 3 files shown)


cscheid (approved):
lgtm

PR #995, opened 2022-08-02
Improve implementation of all `vec_*()` functions

rich-iannone:
The `vec_*()` series of functions is useful for creating formatted values from vectors (totally independent of using tables, which is great for formatting values in an R Markdown or Quarto doc). Here, we include a `"plain"` `output` option. It can use Unicode characters and is suitable for printing in the R console. We also now have an `"auto"` output option (the default) which will determine the output context (e.g., Quarto rendering a PDF document) and choose the output, defaulting to `"plain"` in the interactive case. 

This PR also continues the work on using preferring the use of Unicode characters instead of HTML entities to represent certain non-ASCII characters (e.g., multiplication sign, minus sign, etc.).

The use of CSS classes in formatted values was limited since the classes used are inaccessible outside of a table context. For example, when using `vec_fmt_scientific()` with a vector of numeric values, a class would be present in the output and it likely wouldn't be available when embedding the output values in an HTML document. This is solved by inlining CSS styles to the relevant `style` attribute.


cscheid (approved):
lgtm

PR #996, opened 2022-08-05
Remove dependency on checkmate

rich-iannone:
The `checkmate` package's assertion functions were not really used comprehensively in gt and we have robust checking throughout anyway. So, it seemed prudent to remove the dependency on that package and provide suitable assertion replacements where needed, and that's what this PR delivers.


cscheid (approved):
lgtm

PR #997, opened 2022-08-07
Enhance automatic column alignment

rich-iannone:
With automatic column alignment (set to `"auto"` by default in the `gt()` function), columns (and their column labels) will be set on the basis of the content in the columns. The aim of this PR is to refine that a bit by changing the alignments of dates and time (to `"right"` alignment) but, also, making predominantly number-based strings align to the right. This ensures that string-based dates/times/datetimes receive a better appearance before formatting (so even if formatting isn't necessary, it'll still look formatted in terms of alignment).


cscheid (approved):
lgtm

PR #1011, opened 2022-08-17
Add return value for `gtsave()`

rich-iannone:
This adds a return value (`invisible(TRUE)`) to the bottom of the `gtsave()` function body. This clears up the issue of unwanted printing in R Markdown / Quarto documents and usefully returns `TRUE` in case the user needs to log the outcome of a `gtsave()` call. This also adds to and cleans up documentation in `gtsave()`.

Fixes https://github.com/rstudio/gt/issues/974

### DIFF
--- a/R/export.R
+++ b/R/export.R
@@ -38,19 +38,22 @@
 #' document. The LaTeX and RTF saving functions don't have any options to pass
 #' to `...`.
 #'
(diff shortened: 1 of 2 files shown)


cscheid (approved):
lgtm

Issue #1004, opened 2022-08-13
Warning in print method: `length(x) = 2 > 1' in coercion to 'logical(1)`


ddsjoberg:
## Description

I the recent release of R 4.2.1 (or maybe 4.2.0?), additional messaging to users have been included when a `if ()` statement includes a logical with length greater than one.

It appears that this messaging is now being triggered in the gt print method when certain attributes are applied to the gt table. Example below! 

## Reproducible example

``` r
packageVersion("gt")
#> [1] '0.6.0.9000'
R.version
#>                _                                
#> platform       x86_64-w64-mingw32               
#> arch           x86_64                           
#> os             mingw32                          
#> crt            ucrt                             
#> system         x86_64, mingw32                  
#> status                                          
#> major          4                                
#> minor          2.1                              
#> year           2022                             
#> month          06                               
#> day            23                               
#> svn rev        82513                            
#> language       R                                
#> version.string R version 4.2.1 (2022-06-23 ucrt)
#> nickname       Funny-Looking Kid

x <- 
  gtsummary::trial |>
  dplyr::select(trt) |> 
  gtsummary::tbl_summary() |>
  gtsummary::modify_table_styling(
    columns = label,
    rows = label == "Drug B",
    footnote = "i.e. placebo"
  )

tbl <-
  gt::gt(data = x$table_body, groupname_col = NULL, caption = NULL) |>
  gt::tab_style(
    style = gt::cell_text(indent = gt::px(10), align = "left"),
    locations = gt::cells_body(columns = "label", rows = 2:3)) |>
  gt::tab_footnote(
    footnote = gt::md("i.e. placebo"), 
    locations = gt::cells_body(columns = "label", rows = 3L))

print(tbl)
#> Warning messages:
#> 1: In locname == "data" && colname == .env$colname :
#>   'length(x) = 2 > 1' in coercion to 'logical(1)'
#> 2: In locname == "data" && colname == .env$colname :
#>   'length(x) = 2 > 1' in coercion to 'logical(1)'
#> 3: In locname == "data" && colname == .env$colname && rownum == .env$rownum :
#>   'length(x) = 2 > 1' in coercion to 'logical(1)'
```

<sup>Created on 2022-08-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

## Session info

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
[...]

PR #1012, opened 2022-08-17
Use `&` instead of `&&` in `filter()` stmt

rich-iannone:
This PR fixes an error in a `filter()` statement. The `&&`s should rightly be `&`s. This clears a warning with newer versions of R and correctly filters the table that this statement operates on.

Fixes: https://github.com/rstudio/gt/issues/1004

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -257,7 +257,7 @@ get_alignment_at_body_cell <- function(
   styles_filtered_tbl <-
     dplyr::filter(
       styles_tbl,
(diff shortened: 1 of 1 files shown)


cscheid (approved):
LGTM. Just making sure: did you check the rest of the codebase for similar issues?

rich-iannone:
I checked the .R file in which this error was found for similar defects (none found) but I’ll make a point to check all the other files.

Issue #993, opened 2022-07-24
fmt_markdown breaks a line and put footnote reference onto new line


Dobrokhotov1989:
I've faced an issue when trying to use footnote on the cells formatted with `fmt_markdown()`. With `fmt_markdown()` linebreak between cell content and footnote reference (i.e. <sup>1</sup>) is added. In the example below, `fmt_markdown()` does nothing useful but shows the problem.

```
  library(tidyverse)
  library(gt)
  
  mpg %>%
    sample_n(10) %>%
    select(manufacturer, model, displ, cty, hwy, year) %>%
    gt() %>%
    tab_footnote(
      footnote = "This is old car",
      locations = cells_body(columns = model,
                             rows = year < 2000)
    ) %>%
    cols_hide(columns = year) %>%
    fmt_markdown(columns = model)
```
![Rplot01](https://user-images.githubusercontent.com/45655146/180629338-111c344e-ddb1-4d2a-aff4-a27e73020c06.png)

Issue #893, opened 2022-02-13
Footnote marker appears in new line if formatted with Markdown


turbanisch:
With default formatting, the footnote marker appears right after the cell content:
```r
library(tidyverse)
library(gt)

table <- mtcars %>% 
  head(5) %>% 
  select(1:3) %>% 
  gt() %>% 
  tab_footnote(footnote = "A footnote.", 
               locations = cells_body(1, 3)) 

table
```
<img width="135" alt="image" src="https://user-images.githubusercontent.com/51969412/153770788-9b631a22-6b1b-4b72-98da-e78fee45a521.png">

If the column where the footnote marker appears is formatted using Markdown, however, the marker appears on a new line:
```r
table %>% 
  fmt_markdown(columns = 1)
```
<img width="127" alt="image" src="https://user-images.githubusercontent.com/51969412/153770750-8097c0ac-5abd-46b7-8aba-15afa92a10c5.png">

``` r
sessionInfo()
#> R version 4.1.2 (2021-11-01)
#> Platform: aarch64-apple-darwin20 (64-bit)
#> Running under: macOS Monterey 12.1
#> 
#> Matrix products: default
#> BLAS:   /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/lib/libRblas.0.dylib
#> LAPACK: /Library/Frameworks/R.framework/Versions/4.1-arm64/Resources/lib/libRlapack.dylib
#> 
#> locale:
#> [1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8
#> 
#> attached base packages:
#> [1] stats     graphics  grDevices utils     datasets  methods   base     
#> 
#> loaded via a namespace (and not attached):
#>  [1] digest_0.6.29   withr_2.4.3     magrittr_2.0.2  reprex_2.0.1   
#>  [5] evaluate_0.14   highr_0.9       stringi_1.7.6   rlang_1.0.1    
#>  [9] cli_3.1.1       rstudioapi_0.13 fs_1.5.2        rmarkdown_2.11 
#> [13] tools_4.1.2     stringr_1.4.0   glue_1.6.1      xfun_0.29      
#> [17] yaml_2.2.2      fastmap_1.1.0   compiler_4.1.2  htmltools_0.5.2
#> [21] knitr_1.37
```

<sup>Created on 2022-02-13 by the [reprex package](https://reprex.tidyverse.org) (v2.0.1)</sup>

[...]

PR #1013, opened 2022-08-17
Better integrate footnote marks with table cell text rendered with `fmt_markdown()`

rich-iannone:
This PR ensures that table body cell text generated from `fmt_markdown()` can have footnote marks inline with the produced text. This fix works both for right and left placement of the footnote marks. New testthat tests were added to address these changes.

Fixes: https://github.com/rstudio/gt/issues/993
Fixes: https://github.com/rstudio/gt/issues/893

### DIFF
--- a/R/z_utils_render_footnotes.R
+++ b/R/z_utils_render_footnotes.R
@@ -534,8 +534,7 @@ set_footnote_marks_stubhead <- function(data,
 #' Apply footnotes to the data rows
 #'
 #' @noRd
-apply_footnotes_to_output <- function(data,
-                                      context = "html") {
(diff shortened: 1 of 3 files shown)


cscheid (approved):
I don't have enough context to give a fuller review here. If it fixes the immediate bug then that's good (hence I'm approving)

I'm just generally confused about `fmt_markdown()`? Is GT processing markdown on its own?

rich-iannone:
It is processing markdown on its own with the commonmark package. This has both its advantages and disadvantages (user surprise that some things don’t render the same as in R Markdown). This is a bigger issue that we can talk about somewhere down the line.

PR #1017, opened 2022-08-19
Fix spanner ID generation in `tab_spanner_delim()`

rich-iannone:
In more complex situations, `tab_spanner_delim()` would fail to generate unique ID values. This is now fixed by adding an index to the first-duplicated ID value and incrementing by 1 for subsequent duplications. Now this works:

```r
library(tidyverse)
library(gt)

dplyr::tibble(
      a.i.x = "",
      b.i.x = "",
      a.j.x = "",
      b.j.x = "",
      a.i.y = "",
      b.i.y = "",
      a.j.y = "",
      b.j.y = "",
      a.i.z = "",
      b.i.z = "",
      a.j.z = "",
      b.j.z = ""
    ) %>%
      gt() %>%
      tab_spanner_delim(
        delim = ".",
        split = "first"
      )
```

tab-spanner-delim-fix

Also, using `split = "last"` also works. Added several testthat tests for this.

Fixes: https://github.com/rstudio/gt/issues/1000

### DIFF
--- a/R/tab_create_modify.R
+++ b/R/tab_create_modify.R
@@ -391,6 +391,7 @@ tab_spanner_delim <- function(
       expr = {{ columns }},
       data = data
     )
+
   if (!is.null(columns)) {
     colnames_spanners <- base::intersect(all_cols, columns)
   } else {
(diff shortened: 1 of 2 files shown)


cscheid (approved):
lgtm

PR #1035, opened 2022-09-01
Add the `fmt_roman()` function

rich-iannone:
This adds the `fmt_roman()` function and the corresponding `vec_fmt_roman()` function. This formatter takes numeric and integer values as inputs and outputs the corresponding Roman numerals.


cscheid (approved):
lgtm

Issue #1028, opened 2022-08-29
Feature request: `day_m` option for `date_style` in `fmt_date()`


mine-cetinkaya-rundel:
## Prework

- [x] Read and abide by **gt**'s [code of conduct](https://www.contributor-covenant.org/version/2/0/code_of_conduct/) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/rstudio/gt/issues) (both open and closed).

## Proposal

I'd love a 15th option for `date_style` in `fmt_date()`: `day_m`.

`"day_m"`: `29 Feb`

This will help keep the width of a date column narrower, which I could use in places like [this](https://sta199-f22-1.github.io/).

rich-iannone:
Like the idea and this will be implemented soon!

PR #1042, opened 2022-09-06
Include `"day_m"` option in `fmt_date()` and `vec_fmt_date()`

rich-iannone:
This includes the `"day_m"` option for formatting dates with `fmt_date()` and `vec_fmt_date()`. This makes it possible to have a very short date (w/ implied year) in the table stub.

Now we can something like this:

```r
exibble %>%
  dplyr::select(date, time) %>%
  gt(rowname_col = "date") %>%
  fmt_date(
    columns = date,
    date_style = "day_m"
  ) %>%
  cols_align(
    align = "left",
    columns = stub()
  ) %>%
  sub_missing(
    columns= stub(),
    missing_text = ""
  )
```

new-date-option

Fixes: https://github.com/rstudio/gt/issues/1028

### DIFF
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -2251,11 +2251,12 @@ fmt_bytes <- function(
 #' 7. `"day_m_year"`: `29 Feb 2000`
 #' 8. `"day_month_year"`: `29 February 2000`
 #' 9. `"day_month"`: `29 February`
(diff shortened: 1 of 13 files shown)


cscheid (approved):
lgtm

PR #1043, opened 2022-09-07
Replace stringr functions with base R equivalents

rich-iannone:
This PR removes the few stringr functions that remained in the package. A handful of simple replacement functions are used instead.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -42,7 +42,6 @@ Imports:
     rlang (>= 1.0.2),
     sass (>= 0.4.1),
     scales (>= 1.2.0),
-    stringr (>= 1.4.0),
     tibble (>= 3.1.6),
     tidyselect (>= 1.1.1)
 Suggests:
--- a/R/format_data.R
+++ b/R/format_data.R
@@ -3686,13 +3686,15 @@ fmt_markdown <- function(
         markdown_to_xml(x)
       },
       default = function(x) {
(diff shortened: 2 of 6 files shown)


cderv:
@rich-iannone Interested by this work for **knitr** too: https://github.com/yihui/knitr/issues/1549

I don't know if all the functions are covered by your switch here or not, but definitely some overlap. 

Only constraint for **knitr** - performance. We need to be sure that we don't loose performance in the rendering process. Maybe also an issue with **gt**

cscheid (approved):

rich-iannone:
> @rich-iannone Interested by this work for **knitr** too: [yihui/knitr#1549](https://github.com/yihui/knitr/issues/1549)
> 
> I don't know if all the functions are covered by your switch here or not, but definitely some overlap.
> 
> Only constraint for **knitr** - performance. We need to be sure that we don't loose performance in the rendering process. Maybe also an issue with **gt**

@cderv Many of the stringr functions are covered here by the replacements. I could definitely do the same for knitr, not a problem. A lot of the replacements in this PR are concentrated in one part of the code path (inlining styles in HTML tags). It's not any slower and that whole section will be replaced soon anyway with a fast JS-mediated function.

cderv:
> Many of the stringr functions are covered here by the replacements. I could definitely do the same for knitr, not a problem. 

Awesome !

PR #1062, opened 2022-09-20
Use the APCA contrast algorithm to better select light or dark text in `ideal_fgnd_color()`

rich-iannone:
This updates the `ideal_fgnd_color()` util function add the APCA contrast also. This determines Lc contrast between overlaid text on a background color. The background is fixed and the text color is swapped to the alternative (`light` or `dark`) that maximizes contrast. The code here is adapted from the SAPC APCA (Accessible Perceptual Contrast Algorithm). The WCAG algo has been added here as a (non-default) alternative in `data_color()`. Here is a comparison table demonstrating the differences between the WCAG and APCA contrast schemes.

contrast-compare

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -113,6 +113,7 @@ Collate:
     'summary_rows.R'
     'text_transform.R'
     'utils.R'
+    'utils_color_contrast.R'
     'utils_formatters.R'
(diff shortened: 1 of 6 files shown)


cscheid (approved):
Looks good!

Issue #703, opened 2020-12-22
Format any cell based on value 


rjake:
Just found `gt` and really like it. Using all the `select` helpers is 👌 
The only thing I haven't found is a simple solution to format cells of any column based on the cell value. In my app, the number of columns will vary based on the user's selection. In this reprex, I want to bold any value <= 5.

Some of the solutions I've found weren't easy to implement:
This solution uses `rlang`  + `purrr`
https://community.rstudio.com/t/highlighting-values-in-gt/80608/2

This one uses a for loop across each column:
https://stackoverflow.com/a/63945239/4650934

I was hoping to use something like:
```
tab_style(
  style = ...,
  locations = cells_body(
    columns = ...,
    condition = ~(.x <= 5))
  )
)
```

I hope you'll consider adding a helper function to make this easy to do. Thanks for all the hard work.

```r
library(tidyverse)
library(gt)

set.seed(123)
matrix(sample(1:30, 30), 3, 10) %>% 
  as_tibble() %>% 
  gt()
```

<img src="https://user-images.githubusercontent.com/14003618/102846975-6d923b00-43df-11eb-8e86-908cff890d35.png" width="300">

rich-iannone:
This seems like something that ought to be possible (and I would like this feature a lot). Thank you for suggesting it. I’m hoping that our ongoing work in improving the tidyselect support/semantics in this package will pave the way for your suggested code to work.

PR #1066, opened 2022-09-23
Add the `sub_values()` function for subbing in arbitrary text in body cells based on value or match

rich-iannone:
This adds the `sub_values()` function which allows for more flexibility in substituting text based on cell data values (different than `text_transform()`, which occurs later in the render process and operates on fully formatted text). This function allows for text to replace either exact values the user specifies or those (character-based) values found though regex matching. Plenty of tests were added and the PR also fixes some deficiencies discovered in LaTeX and RTF escaping.

Fixes: https://github.com/rstudio/gt/issues/703


cscheid (approved):
lgtm

Issue #694, opened 2020-11-22
HTML/Markdown in group heading rows is not processed


abielr:
Currently HTML or Markdown stored in a column that is then defined as the row group (groupname_col) is not rendered properly (the characters are escaped). This appears to occur in `dt_stub_df.R::dt_stub_df_init()` at the following stage, which converts the column to plain text even if the `gt::html` or `gt::md` functions were used.

    row_group_labels <-
      apply(
        data_tbl[, groupname_col],
        MARGIN = 1,
        paste, collapse = row_group.sep
      )

Would like to request that users be able to pass HTML/MD into the row group.

rich-iannone:
This seems quite reasonable. While it’s not yet possible, I think the way this ought to be done is to allow formatters (`fmt_markdown()` in this case) to transform columns that are used for group row labels and stub labels.

PR #1097, opened 2022-10-31
Provide option to process Markdown in `"rowname_col"` and `"groupname_col"` when using `gt()`

rich-iannone:
This PR adds the ability to treat columns containing row names or group names as Markdown text. This provides the `process_md` argument in the `gt()` function. If this option is taken then the supplied names will be treated to ensure that there are unique IDs for later targeting (for `tab_style()`, `tab_footnote()`, etc.).

Fixes: https://github.com/rstudio/gt/issues/694


cscheid (approved):
lgtm

Issue #1102, opened 2022-11-01
Add function to display all ID values/labels for a gt table


rich-iannone:
It can become increasingly difficult to recall the ID values associated with different labels in a gt table. There are also processes where gt will generate ID values (e.g., `tab_spanner_delim()`, etc.) while ensuring that duplicate ID values aren't produced. For the latter case, it is impossible to know what those ID values are unless one were to examine to correct component of the `gt_tbl` object.

Because it's important to know these ID values for targeting (for styles, creating borders, adding footnote marks), we should have a convenient function (`tab_info()`) that summarizes (by location) the ID values and their labels.

PR #1104, opened 2022-11-03
Add the `tab_info()` function

rich-iannone:
This PR adds the `tab_info()` function for summarizing id/label values in an informational gt table.

Fixes: https://github.com/rstudio/gt/issues/1102


cscheid (approved):
lgtm!

Issue #863, opened 2022-01-18
How to set background colors for the cells of multiple columns based on string contents


ahbon123:
Hello, say I have a small dataset as follows:

```
library(gt)
library(tidyverse)

id <- c(1,2,3,4,5)
res1 <- c("true", "true", "false", "true", "false")
res2 <- c("false", NA, NA, "true", "true")
df <- data.frame(id, res1, res2)

df %>% 
  gt()
```

For columns `res1 ` and `res2`, let's say if value content are `true`s, I'll need to highlight that cell with red background color, if `false`s, highlight with green color, for other cases, keep their colors as original.

Is it possible to achieve that using gt package?

rich-iannone:
We need an easier way to set basic style attributes based on values. I think a new function for this should be the way to do it. The `sub_values()` function provides a nice (and forgiving) way to target values so I believe the new function ought to be modeled on that.

ahbon123:
Thank you very much, I believe this function will have lots of application scenarios.

jthomasmock:
Tagging on to this, there are a few requests I've seen on Twitter and `gtExtras` re: color the background of column 1 based on column 2. Right now, `data_color()` only operates on the current column, `text_transform()` only works on the current column (although this can be "abused"), and to properly color the background of a column you end up needing to use `tab_style()` multiple times.

I have a prototype that works like below:

```r
set.seed(37)
base_tab <- head(mtcars) |>
  dplyr::slice_sample(n = 6) |>
  dplyr::arrange(desc(mpg)) |>
  gt::gt()

base_tab |>
  gt_color_by_col(mpg, cyl, domain = range(gtExtras::gt_index(base_tab, mpg)))
```

![image](https://user-images.githubusercontent.com/29187501/200356656-b363a793-c869-47e3-912b-4c4a69ac8da1.png)

Essentially it loops through the column, values, and rows to apply a `tab_style()` with the minimal snippets of code below:

```r
fn_data_clr_by_col <- function(gt_object, col2, value, row_id, color_add){

  gt_object |>
    tab_style(
      style = list(
        cell_fill(color = color_add),
        cell_text(color = gt:::ideal_fgnd_color(color_add))
      ),
      locations = cells_body({{ col2 }}, rows = row_id)
    )

}
```

```r
for (i in 1:length(full_vals)){
    gt_object <- fn_data_clr_by_col(gt_object, {{ col2 }}, full_vals[i], i, color_add[i])
  }
  return(gt_object)
```

[...]

PR #1108, opened 2022-11-08
Add `tab_style_body()` variant to style by targeted body cells

rich-iannone:
For easier ways to set basic style attributes based on values (in body cells and surrounding areas), the `tab_style_body()` function (based on `tab_style()`) is added in this PR. The `sub_values()` function provides a nice (and forgiving) way to target values so the new function added here is modeled on that.

Fixes: https://github.com/rstudio/gt/issues/863


cscheid (approved):
lgtm!

Issue #1093, opened 2022-10-28
`as_raw_html` issue with `cells_stub` styling


cgoo4:
Hi @rich-iannone - Great package!

I'm getting misaligned data when combining `as_raw_html` with `cells_stub` styling per the minimal reprex below:

``` r
library(tibble)
library(gt)

mtcars |> 
  rownames_to_column() |> 
  gt(rowname_col = "rowname", groupname_col = "cyl") |> 
  tab_style(
    style = cell_text(color = "blue", weight = "bold"),
    locations = cells_stub()
  ) |> 
  as_raw_html()
```
<img width="730" alt="image" src="https://user-images.githubusercontent.com/32419319/198693071-f92ae36c-e65e-44ee-85d7-5741e1f6aa54.png">

<sup>Created on 2022-10-28 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.2.1 (2022-06-23)
#>  os       macOS Big Sur ... 10.16
#>  system   x86_64, darwin17.0
#>  ui       X11
#>  language (EN)
#>  collate  en_GB.UTF-8
#>  ctype    en_GB.UTF-8
#>  tz       Europe/London
#>  date     2022-10-28
#>  pandoc   2.19.2 @ /Applications/RStudio.app/Contents/MacOS/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version    date (UTC) lib source
#>  assertthat    0.2.1      2019-03-21 [1] CRAN (R 4.2.0)
#>  cli           3.4.1      2022-09-23 [1] CRAN (R 4.2.1)
#>  colorspace    2.0-3      2022-02-21 [1] CRAN (R 4.2.0)
#>  DBI           1.1.3      2022-06-18 [1] CRAN (R 4.2.0)
#>  digest        0.6.30     2022-10-18 [1] CRAN (R 4.2.1)
#>  dplyr         1.0.10     2022-09-01 [1] CRAN (R 4.2.1)
#>  evaluate      0.17       2022-10-07 [1] CRAN (R 4.2.0)
#>  fansi         1.0.3      2022-03-24 [1] CRAN (R 4.2.0)
[...]

Issue #915, opened 2022-03-24
Weird html output when using `tab_style()` with `as_raw_html()`


Beck466:
## Description

I am making a gt table which colours text in a row depending on the value in a column.  gt behaves as expected when making the table, however when the table is used with `as_raw_html` the first cell of the first row is moved outside of the table (and other entries in that row are moved one to the left) if that row has been styled (not: works perfectly with gt 0.3.0 but not with gt 0.4.0

With gt 0.3.0:
![image](https://user-images.githubusercontent.com/72534062/159841822-0597685f-e9b1-4626-8c02-e03e2d574ab0.png)

With gt 0.4.0:
![image](https://user-images.githubusercontent.com/72534062/159841907-008c201a-3df4-4952-b063-74921bf29091.png)

## Reproducible example

``` r
library(tidyverse)
library(gt)

df <- tribble(~Site, ~Plant, ~Circuit, ~Frequency, ~Weeks, ~colour,
              "Site_1", "plant_1",  "circuit_1",    "Frequent", 28, "green",
              "Site_1", "plant_2",  "circuit_b",    "Frequent", 28, "green",
              "Site_1", "plant_3",  "circuit_c",    "Frequent", 28, "green",
              "Site_1", "plant_4",  "circuit_d",    "Frequent", 28, "green",
              "Site_1", "plant_5",  "circuit_e",    "Frequent", 26, "green",
              "Site_2", "plant_6",  "circuit_f",    "Frequent", 28, "black",
              "Site_2", "plant_7",  "circuit_g",    "Frequent", 28, "black",
              "Site_2", "plant_8",  "circuit_h",    "Frequent", 28, "green",
              "Site_2", "plant_9",  "circuit_i",    "Frequent", 28, "green",
              "Site_3", "plant_10", "circuit_j",    "Frequent", 28, "green",
              "Site_3", "plant_11", "circuit_k",    "Frequent", 28, "black",
              "Site_3", "plant_12", "circuit_l",    "Frequent", 28, "green",
              "Site_3", "plant_13", "circuit_m",    "Frequent", 28, "green")

df %>% 
  gt(groupname_col = "Site") %>% 
  tab_style(
    style = cell_text(weight = "bold"),
    locations = cells_column_labels(columns = everything())) %>% 
  tab_style(
    style = list(
      cell_text(weight = "bold"),
      cell_fill(color = "lightblue")),
    locations = cells_row_groups()) %>%
  tab_style(
    style = list(
[...]

Issue #858, opened 2021-12-10
table id when using `as_raw_html()`


mattkerlogue:
I've been using {gt} for a lot of internal reports, and relying on external styling of the output via a separate CSS file, as such I use `as_raw_html(inline_css = FALSE)` at the end of a {gt} pipeline to output just the HTML table.

In a live project I've had some trouble with one specific table that needs to be handled slightly differently to the main CSS file's style definitions. I hoped that setting `gt(id = "special-table")` would result in the id coming through to the output HTML, but it seems the id is inserted in a containing `<div>` (which makes sense for the general usage).

As referenced in [#816](https://github.com/rstudio/gt/issues/816#issuecomment-930172380), I'm actually using v0.2.2 because I'm wanting to apply custom styling (to {gt} generated classes) and later versions enclose the table within a `<div>` that includes style information which overrides the external stylesheet. By using v0.2.2 the object at the end of my pipeline is not contained within its own `<div>`, it's the raw `<table>` object (a behaviour I'd like to retain). I wondered if it is possible however to get the id applied to the table object when using `as_raw_html()`.

The code:

```r
library(gt) # v0.2.2

gtcars |> 
  gt(id = "test-table") |> 
  as_raw_html(inline_css = FALSE)
```

Produces:
```
<table class="gt_table">
  
  <thead class="gt_col_headings">
    <tr>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">mfr</th>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">model</th>
...
```

I'd hope for something like:
```
<table class="gt_table" id="test-table">
  
  <thead class="gt_col_headings">
    <tr>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">mfr</th>
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1">model</th>

```

Current less-than-ideal-hacky-workaround:

```r
library(gt) #v0.2.2

add_tab_id <- function(x, id) {
  x <- gsub("(<table class=\"gt_table\")(>)", 
            paste0("\\1 id=\"",id,"\"\\2"),
            x)
  htmltools::HTML(x)
}

gtcars |> 
  gt() |> 
  as_raw_html(inline_css = FALSE) |>
  add_tab_id(id = "test-table")
```

rich-iannone:
[...]

Issue #837, opened 2021-09-13
`as_raw_html()` doesn't include images


bzigterman:
When I run the code below to make an html table with `gt`, the images appear fine in Rstudio's viewer, but when I run them through `as_raw_html`, the image references don't make it into the html. Unless I'm doing something wrong, this appears to be a bug in `gt`.

This also happens when I use `local_image`. Curiously, it doesn't happen when I add `inline_css = FALSE` to `as_raw_html`. 

```
library(gt)
library(tidyverse)

standings_table <-
  dplyr::tibble(
    pixels = px(seq(10, 35, 5)),
    image = seq(10, 35, 5)
  ) %>%
  gt() %>%
  text_transform(
    locations = cells_body(columns = image),
    fn = function(x) {
      web_image(
        url = "https://www.mlbstatic.com/team-logos/142.svg",
        height = as.numeric(x)
      )
    }
  )

standings_table
standings_table_html <- as_raw_html(standings_table)
standings_table_html
```

banbh:
I have encountered this problem too, but I wonder if it is related to the *size* of the image (i.e., the data URI). See below for a minimal working example. What seems to happen is that the smaller image ('dot') is included but the larger one is not.
```R
library(tidyverse)
library(gt) # gt_0.7.0

images <- c(dot = '<img src="data:image/png;base64,iVBORw0KGgoAAA
ANSUhEUgAAAAUAAAAFCAYAAACNbyblAAAAHElEQVQI12P4
//8/w38GIAXDIBKE0DHxgljNBAAO9TXL0Y4OHwAAAABJRU
5ErkJggg=="/>',
            test = local_image(file = test_image(type = "png")))

tibble(
  name = names(images),
  image = 1:2
) %>%
  gt() %>%
  text_transform(
    locations = cells_body(columns = image),
    fn = function(x) {
      images[as.numeric(x)]
    }
  ) ->
  my_table

print(my_table) # OK: Displays both images

str_count(as_raw_html(my_table), pattern = '<img') # == 1: Only the 'dot' smaller is included
# expect table to be bigger than on of its contained images!
stopifnot(nchar(as_raw_html(my_table)) > nchar(images[2])) # Fails!
```

banbh:
[...]

PR #1114, opened 2022-11-09
Enhance CSS-inlining within `as_raw_html()`

rich-iannone:
The `as_raw_html()` function is useful for generating an HTML string for email message bodies and for HTML embedding purposes (expressly with `inline_css = TRUE`). While the CSS-inlining has been adequate it is: (1) slow, and (2) underpowered and can't keep up with additions to the SCSS styles. This is solved by using an external package that uses the `juice` JS library. This solution is far more performant and correct. Given the degree to which inlining problems are being seen (this especially impacts HTML email sending with gt tables), it is necessary to improve this part of the package.

Fixes: https://github.com/rstudio/gt/issues/1093
Fixes: https://github.com/rstudio/gt/issues/915
Fixes: https://github.com/rstudio/gt/issues/858
Fixes: https://github.com/rstudio/gt/issues/837
Fixes: https://github.com/rstudio/gt/issues/540
Fixes: https://github.com/rstudio/gt/issues/455


cscheid (approved):
lgtm

Issue #1105, opened 2022-11-04
Add more margin space to the top, bottom, left, and right edges of the gt table


ahbon123:
Hello,

I used the code below to plot and save a gt table. Now I am facing a problem that the top, bottom, left and right edges of the table are too compact (the space is too small).

```
library(gt)
p <- mtcars %>% 
  select(cyl:wt) %>% 
  head() %>% 
  gt()

gtsave(p, file='test.png', vwidth=1000, vheight=600, zoom=1)
```
Out:

![image](https://user-images.githubusercontent.com/15428667/199900078-a504c640-2baa-4d7e-be0c-4aee78752159.png)

Thus I want to add more margin space (increase the area between the two blue boxes as figure above shows) to the four edges of the table.

How can I achieve this? Maybe I'm missing some parameters, but I didn't find the right ones in [tab_options()](https://gt.rstudio.com/reference/tab_options.html). Thanks.

Issue #590, opened 2020-05-20
Table margin bottom too small in RNotebook


MokeEire:
The bottom margin between a `gt()` table in an RNotebook chunk (with no specified chunk options) and markdown text which follows it is too small.  

```r
mtcars %>% 
  rownames_to_column() %>% 
  mutate(hp_per_us_ton = hp/(2*wt)) %>% 
  arrange(-hp_per_us_ton) %>% 
  select(rowname, mpg, hp, wt, hp_per_us_ton) %>% 
  head(10) %>% 
  gt()
```

Looks like this:

![rmarkdown issue](https://user-images.githubusercontent.com/19700454/82392321-6c87ce00-99f8-11ea-8c31-95cd1cdd2d2c.PNG)

rich-iannone:
I was reminded of this recently and I think it definitely does not look good at all. Also, occurs in RMarkdown documents (and probably in other contexts too).

MokeEire:
The following addition to a CSS file solved it for me:

```
.gt_table {
    margin-bottom: 20px;
    margin-top: 20px;
}
```

qdread:
The CSS solution was helpful for me also in the context of `print()`ing a large number of gt tables in a loop within a code chunk, to put padding between them.

PR #1116, opened 2022-11-10
Add container padding options

rich-iannone:
This adds two options in `tab_options()` for setting the left/right and top/bottom padding on the table's container (HTML only). This is necessary since the lack of default padding (and any means to set it) results in tables being uncomfortable close to paragraphs above and below.

Fixes: https://github.com/rstudio/gt/issues/1105
Fixes: https://github.com/rstudio/gt/issues/590


cscheid (approved):
lgtm!

Issue #1054, opened 2022-09-12
span labels not emitted correctly to rtf


gmbecker:
To reproduce:
```
library(gt)

mydf <- data.frame(x = 5, y = "hi", z = "what?")
mygt <- mydf |>
  gt() |>
  tab_spanner(label ="spaannning", columns = c(y, z))

rtf_str <- as_rtf(mygt)
```

Then `cat` `rtf_str` to the file of your choice and open the file in an editor/viewer that supports RTF.

For me, I tried TextEdit  (Version 1.17 (380.2)) and LibreOffice 7.3.5.2. Both of these display (different) incorrect tables:

![gt_rtf_loaded_screenshot](https://user-images.githubusercontent.com/908721/189775730-aa2a41f9-58c6-45c2-9794-7be8b66c1b3f.png)

Looking at `rtf_str` we see:

```
 rtf_str
{\rtf\ansi\ansicpg1252{\fonttbl{\f0\froman\fcharset0\fprq0 Courier New;}{\f1\froman\fcharset0\fprq0 Times;}}{\colortbl;\red211\green211\blue211;}

\paperw12240\paperh15840\widowctrl\ftnbj\fet0\sectd\linex0
\lndscpsxn
\margl1440\margr1440\margt1440\margb1440
\headery720\footery720\fs20

\trowd\trrh0\trhdr

\pard\plain\uc0\qc\clvertalc \clpadfl3\clpadl25 \clpadft3\clpadt85 \clpadfb3\clpadb25 \clpadfr3\clpadr85 \cellx3120
\intbl {\f0 {\f0\fs20 }}\cell

\pard\plain\uc0\qc\clvertalc \clpadfl3\clpadl25 \clpadft3\clpadt85 \clpadfb3\clpadb25 \clpadfr3\clpadr85\clbrdrb\brdrs\brdrw20\brdrcf1\clmgf \cellx6240
\intbl {\f0 {\f0\fs20 spaannning}}\cell

\pard\plain\uc0\qc\clvertalc \clpadfl3\clpadl25 \clpadft3\clpadt85 \clpadfb3\clpadb25 \clpadfr3\clpadr85\clbrdrb\brdrs\brdrw20\brdrcf1\clmrg \cellx9360
\intbl {\f0 {\f0\fs20 spaannning}}\cell

\row

<snip>

```
[...]

PR #1121, opened 2022-11-11
Fix display of spanner column labels in different applications that render RTF

rich-iannone:
This fix ensures that spanner column labels are rendered correctly in a wider variety of RTF applications (OpenOffice, Apple TextEdit, MS Word, etc.). Previously, Word could work around the non-ideal RTF generated but the fix ensures that more renderers display spanner column labels properly.

Fixes: https://github.com/rstudio/gt/issues/1054


cscheid (approved):
lgtm

PR #1122, opened 2022-11-11
Allow the table stub to participate in column merging operations

rich-iannone:
This PR allows the stub contents to freely merge with other columns (using the `cols_merge*()` functions). With this, the stub won't be accidentally hidden since the logic behind that was altered to exclude the hiding of columns designated as the stub.


cscheid (approved):
lgtm

Issue #1118, opened 2022-11-10
Handling missing values in a col_merge


erhoppe:
Thanks for the great package! It's super useful. I wondered if it would be possible to add an option for dealing with missing data in `cols_merge` so that you could either override `sub_missing` or somehow provide an alternative pattern for cells where you only have missing data. 

I'm encountering a situation where I have a subgroup of data for a particular column that is missing data, so my summary with the median, min, and max are all missing. I would like to just have a single "n.p" in that cell rather, but currently, it doesn't appear as if this is possible.
```
library(tidyverse)
library(gt)

tibble(
  category = c("a","b","c","d"),
  median = c(5,10,8, NA_real_),
  min = c(1,6,4,NA_real_),
  max = c(8,25,12,NA_real_),
) %>% 
  gt(rowname_col = "category") %>% 
  cols_merge(columns = c(median, min, max),
             pattern = "{1}<br>({2},{3})") %>%
  tab_spanner(
    label = "Measurement",
    columns = c(median)
  ) %>% 
  cols_label(
    median = html("Median<br>(min,max)")
  ) %>% 
  sub_missing(columns =
                c(median),
              rows = everything(), missing_text = "n.p.") %>%
  sub_missing(columns =
                c(min,max),
              rows = everything(), missing_text = "") %>% 
  tab_source_note(
    source_note = "n.p. = annotation not provided by authors"
  )
```

<img width="442" alt="Screen Shot 2022-11-10 at 1 15 29 AM" src="https://user-images.githubusercontent.com/29584603/201049791-3aa33334-b59e-4953-a54b-d1c08c78595b.png">

Would something like this be possible?

rich-iannone:
Thanks for filing this issue and providing a great example! This has come up a few times and I think we need some extra syntax in the `pattern` to enclose a run of text. Where any NAs are encountered, the entire span would be omitted.

erhoppe:
That sounds like a great solution for it. I look forward to seeing what you come up with. :) Thanks again!

PR #1144, opened 2022-11-24
Add secondary pattern support for `pattern` arg

rich-iannone:
This provides support for secondary patterns whereby segments of a pattern may be removed should there be a missing value within the new `<<`/`>>` delimiters.

For example, `"<<{1} <<{2} << ({3}) >>>>>>"` might yield `"a b (c)"` given no position having missing values. The same pattern would yield `"a "` with `{2}` being missing and `"a b "` with `{3}` missing.

Fixes: https://github.com/rstudio/gt/issues/1118


erhoppe:
Sweet! Thanks so much for addressing this so quickly! :D I really appreciate it.

rich-iannone:
@erhoppe you’re welcome! I forgot to follow up on this in the Issue but I’m glad you noticed this. Let me know if there are any shortcomings or other issues!

PR #1152, opened 2022-12-01
Cols hide missing

mrcaseb:
# Summary

This PR introduces the exported function `cols_hide_missing()`. It checks if columns in the table data are missing all rows, i.e. all values equal `NA`, and hides them using `cols_hide()`.

It is obviously possible to omit such columns before entering gt but the "all-NA" columns could still be useful during formatting of other columns.

I keep the PR a draft as the cols_hide documentation needs to be updated for cols_hide_missing. I didn't do this yet because I wanted to discuss the general interest in this function first.

# Checklist

- [x] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/0/code_of_conduct/).


rich-iannone:
Thanks for the contribution but after thinking about this for some time, it's possible to create an equivalent tidyselect expression and supply it to `cols_hide()` for the same effect.

mrcaseb:
> Thanks for the contribution but after thinking about this for some time, it's possible to create an equivalent tidyselect expression and supply it to `cols_hide()` for the same effect.

Oooh. I never thought about that! Do you suggest something like this?

```r
tidyselect::where(~ any(!is.na(.x))
```

rich-iannone:
Yeah, that would do it! Again sorry to close but it’s in the interest of limiting new functions (as we are super high in exported function count now).

mrcaseb:
> Yeah, that would do it! Again sorry to close but it’s in the interest of limiting new functions (as we are super high in exported function count now). 

Oh no worries. I prefer limiting exported functions as well and appreciate the tidyselect suggestion as well as your work.

PR #1470, opened 2023-10-16
Improve performance of `render_html()`

mgirlich:
# Summary

Rendering slightly bigger tables (e.g. 50k rows) as HTML is very slow. The two main culprits are `create_body_component_h()` and {htmltools} (in particular `tagWrite()` but also creating the tags themself).

Two small benchmarks

```r
diamonds_1000 <- ggplot2::diamonds |>
  head(1000) |>
  gt::gt()
diamonds <- ggplot2::diamonds |>
  gt::gt()

prepare <- function(data) {
  data <- gt:::build_data(data = data, context = "html")
  data <- gt:::add_css_styles(data = data)
  gt:::create_body_component_h(data = data)
}

bench::mark(
  diamonds_1000 = prepare(diamonds_1000),
  diamonds = prepare(diamonds),
  check = FALSE
)

# CRAN
# # A tibble: 2 × 13
#   expression        min median `itr/sec` mem_alloc `gc/sec` n_itr
#   <bch:expr>    <bch:t> <bch:>     <dbl> <bch:byt>    <dbl> <int>
# 1 diamonds_1000   1.14s  1.14s    0.878     5.37MB     2.63     1
# 2 diamonds       56.42s 56.42s    0.0177  175.21MB     2.11     1

# This PR
# # A tibble: 2 × 13
#   expression         min   median `itr/sec` mem_alloc `gc/sec` n_itr
#   <bch:expr>    <bch:tm> <bch:tm>     <dbl> <bch:byt>    <dbl> <int>
# 1 diamonds_1000 146.87ms 148.81ms     6.33     7.72MB     6.33     4
# 2 diamonds         7.46s    7.46s     0.134  301.76MB     6.03     1
```

# Checklist

# Notes

`create_body_component_h()` is quite huge. I think it would make sense to split parts of it into separate functions, e.g. a function for the group headings, the summaries, the data itself. I didn't do this as I didn't want to change the style too much.

The PR is probably quite a bit easier to understand by looking at the commits one by one. Otherwise the changes are hard to follow.

There are a couple of places in `render_row_data()` that deal with spaces and creating HTML. This feels quite tedious and error prone. It would be great to use {htmltools} for such a task but this would require to increase the performance of {htmltools} much more.

I also opened an [issue](https://github.com/rstudio/htmltools/issues/413) in {htmltools} regarding the performance

### DIFF
--- a/R/utils_render_html.R
+++ b/R/utils_render_html.R
@@ -1428,10 +1428,13 @@ create_body_component_h <- function(data) {
 
     body_rows <- lapply(
       seq_along(body_rows_vec),
-      \(i) {
-        list(htmltools::tags$tr(
-          class = if (!is.na(row_classes[[i]])) row_classes[[i]],
-          htmltools::HTML(body_rows_vec[[i]])
+      function(i) {
+        list(htmltools::tag(
+          "tr",
+          varArgs = list(
+            class = if (!is.na(row_classes[[i]])) row_classes[[i]],
+            htmltools::HTML(body_rows_vec[[i]])
+          )
         ))
       }
     )
--- a/R/utils_render_html.R
+++ b/R/utils_render_html.R
@@ -1240,14 +1240,10 @@ create_body_component_h <- function(data) {
     # Is this the first row of a group?
     group_start <- !is.null(g) && group_row_start == i
 
-    # Is this the first row of a group where there is a two-column stub?
-    group_start_two_col_stub <- has_two_col_stub && group_start
-
     #
     # Create a body row
     #
 
-    has_rtl_i <- has_rtl[i, ]
     # If any characters come from a RTL script, ensure that a
     # left alignment is transformed to a right alignment
     has_rtl_i <- has_rtl[i, ]
@@ -1385,7 +1381,8 @@ create_body_component_h <- function(data) {
       summaries_present &&
       !is.null(group_has_summary_rows) &&
       group_has_summary_rows &&
-      group_start_two_col_stub &&
+      has_two_col_stub &&
+      group_start &&
       !is.null(group_summary_row_side) &&
       !is.na(group_summary_row_side) &&
       group_summary_row_side == "top"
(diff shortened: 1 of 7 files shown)


R/utils_render_html.R
```diff
@@ -1063,419 +1063,400 @@ create_body_component_h <- function(data) {
   }
 
   # Create a default vector of row span values for group labels as a column
-  row_span_vals <- rep_len(list(NULL), n_cols_total)
+  row_span_vals <- rep_len(NA_integer_, n_cols_total)
```
mgirlich, line 1066: Using vectors instead of lists is much nicer to work with and also faster.

R/utils_render_html.R
```diff
@@ -1063,419 +1063,400 @@ create_body_component_h <- function(data) {
...
+  # end fixme
+
+  group_idx <- rep(list(NULL), n_rows)
+  for (i in seq_len(n_groups)) {
+    start <- groups_rows_df$row_start[[i]]
+    end <- groups_rows_df$row_end[[i]]
+    group_idx[seq(start, end)] <- i
+    group_ids[seq(start, end)] <- groups_rows_df$group_id[[i]]
+  }
+  groups_list <- as.list(groups_rows_df)
```
mgirlich, line 1226: Accessing fields from a list is much faster than fields of a data frame

rich-iannone:
@mgirlich This is great work and I really appreciate it!

The only problem I can see is that adding `vtcrs` as a dependency would bring the total number to 21, which is something that CRAN flags during updates (and ultimately won't, to my knowledge, accept). However, the `tibble` dependency could be removed with some minor, extra work. If you'd be willing to replace the two instances where tibble is used, then the `vtcrs` addition will be okay.

From a quick search, `tibble` functions are used here:

- https://github.com/rstudio/gt/blob/9c6a46c84c05a111bf9b62e8f07222d2dd79029b/R/utils_render_rtf.R#L415C21-L415C21
- https://github.com/rstudio/gt/blob/9c6a46c84c05a111bf9b62e8f07222d2dd79029b/R/z_utils_render_footnotes.R#L357C5-L357C5

rich-iannone (approved):
lgtm!

PR #1478, opened 2023-11-02
Make gtsave return file path invisibly.

olivroy:
# Summary

Have gtsave follow more closely `ggsave()` by returning the file path instead of `TRUE`.

This is a naive solution. I guess the code could be cleaned up a little bit for `gtsave()` since the related htmltools issue seems to have been solved. May no longer require this fs code? 

Some tests seem to be a little redundant too? I could take another (more thorough) shot at the `gtsave()` code and tests if you think this is worth pursuing.

# Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -2,6 +2,8 @@
 
 * The performance of rendering bigger tables as HTML has been improved and is now up to 3 times faster than before (@mgirlich, #1470).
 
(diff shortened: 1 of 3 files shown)


rich-iannone:
Thanks for this PR! Agree it’s much better to return the image path rather than `TRUE`. Also agree that some of the surrounding code could be revisited for a cleanup!

rich-iannone (approved):
Lgtm!

Issue #1486, opened 2023-11-10
cols_nanoplot not rendering correct values when all values are the same


mtjhrsk:
## Description

Look like cols_nanoplot is not rendering correct values when all the values are the same in a row.

## Reproducible example
``` r
library(dplyr)
library(gt)

simdat <- data.frame (id  = c("A", "B", "C", "D"),
                  sim_1 = c("2", "3", "","10"),
                  sim_2 = c("2", "2", "","10"),
                  sim_3 = c("2", "1", "","10"),
                  sim_4 = c("2", "2", "1","10")
                  ) %>% 
  mutate_at(vars(starts_with("sim_")), funs(as.numeric))

simdat %>% 
  gt(rowname_col = "id") %>%
    cols_hide(columns = c(starts_with("sim_"))) %>% 
  cols_nanoplot(
        columns = starts_with("sim_"),
    new_col_name = "nanoplots",
    new_col_label = md("Estimate over time"),
    missing_vals = "zero") 
```

This is the resulting table:

![tabplot](https://github.com/rstudio/gt/assets/3810592/b825c6aa-2682-4981-8911-6d560b1531eb)

## Expected result

Lines A and D should be straight.

## Session info

sessionInfo()
R version 4.3.2 (2023-10-31 ucrt)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19045)

Matrix products: default

attached base packages:
[1] stats     graphics  grDevices utils    
[5] datasets  methods   base     

other attached packages:
[1] reprex_2.0.2    dplyr_1.1.3    
[3] gt_0.10.0.9000  tidyr_1.3.0    
[5] RPostgres_1.4.6

loaded via a namespace (and not attached):
 [1] config_0.3.2      sass_0.4.7       
 [3] utf8_1.2.4        generics_0.1.3   
 [5] xml2_1.3.5        hms_1.1.3        
 [7] digest_0.6.33     magrittr_2.0.3   
 [9] grid_4.3.2        evaluate_0.23    
[11] fastmap_1.1.1     blob_1.2.4       
[13] processx_3.8.2    DBI_1.1.3        
[15] ps_1.7.5          purrr_1.0.2      
[17] fansi_1.0.5       scales_1.2.1     
[19] cli_3.6.1         rlang_1.1.2      
[21] crayon_1.5.2      bit64_4.0.5      
[23] munsell_0.5.0     commonmark_1.9.0 
[...]

PR #1503, opened 2023-12-01
Fix issue where constant-value nanoplots show jittered values

rich-iannone:
This PR addresses an issue where non-changing values in a nanoplot deceptively look different (within a very small range). The changes here recognize that all values are the same; the y-value range will get adjusted around the non-changing values.

Fixes: https://github.com/rstudio/gt/issues/1486

### DIFF
--- a/R/utils_plots.R
+++ b/R/utils_plots.R
@@ -164,6 +164,13 @@ generate_nanoplot <- function(
   y_scale_max <- get_extreme_value(y_vals, stat = "max")
   y_scale_min <- get_extreme_value(y_vals, stat = "min")
 
+  if (y_scale_min == y_scale_max && is.null(expand_y)) {
+
+    # Expand the `y` scale, centering around the `y_scale_min` value
+    expand_y_dist <- (y_scale_min / 10) * 2
+    expand_y <- c(y_scale_min - expand_y_dist, y_scale_min + expand_y_dist)
+  }
+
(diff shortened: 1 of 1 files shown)


mtjhrsk:
Hi,
thanks for looking into this!
I does work, but I am still getting incorrect jittered values when all the input values for a row are 0. Ofc this could be solved by replacing them with NAs, but wanted to flag this in case it causes some other issues.

`library(dplyr)
library(gt)

simdat <- data.frame (id  = c("A", "B", "C", "D"),
                      sim_1 = c("2", "3", "0","10"),
                      sim_2 = c("2", "2", "0","10"),
                      sim_3 = c("2", "1", "0","10"),
                      sim_4 = c("2", "2", "0","10")
) %>% 
    mutate_at(vars(starts_with("sim_")), funs(as.numeric))

simdat %>% 
    gt(rowname_col = "id") %>%
    cols_hide(columns = c(starts_with("sim_"))) %>% 
    cols_nanoplot(
        columns = starts_with("sim_"),
        new_col_name = "nanoplots",
        new_col_label = md("Estimate over time"),
        missing_vals = "zero") `

Result: 
nano

Expected: values for C should form a straight line.

rich-iannone:
I discovered the same thing (after the PR merged, of course)... I think I'll make it so that non-changing values are treated more specially (no matter what the value is). I will open a new issue about this soon. Thanks for testing this so quickly after the 'fix'!

Issue #1554, opened 2024-01-25
Footnote symbol in cell appears before value instead of after


etiennebacher:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

The symbol in superscript that indicates the footnote appears before the value in the cell, rather than after. Seems close to https://github.com/rstudio/gt/issues/1541 but I can't reproduce the issue reported there.

## Reproducible example

Using the example in `?cells_body()` (with fewer rows):
``` r
library(gt)
gtcars |>
  head(n = 2) |> 
  dplyr::select(mfr, model, year, hp) |>
  gt() |>
  tab_footnote(
    footnote = "Highest horsepower.",
    locations = cells_body(
      columns = hp,
      rows = hp == max(hp)
    )
  ) |>
  opt_footnote_marks(marks = c("*", "+"))
```

![image](https://github.com/rstudio/gt/assets/52219252/18e4fbbe-6b79-4755-b006-144589b5cccd)

## Expected result

The `*` symbol in the cell should appear after the value "647", not before. The [example on the website](https://gt.rstudio.com/reference/cells_body.html#examples) shows the correct behavior.

## Session info

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.2 (2023-10-31 ucrt)
#>  os       Windows 10 x64 (build 19044)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  English_Europe.utf8
#>  ctype    English_Europe.utf8
#>  tz       Europe/Paris
#>  date     2024-01-25
#>  pandoc   3.1.1 @ C:/Program Files/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version     date (UTC) lib source
#>  cli           3.6.2       2023-12-11 [1] CRAN (R 4.3.2)
[...]

PR #1555, opened 2024-01-25
Update several images in the documentation

rich-iannone:
This updates a few images used in the documentation, mostly related to typesetting changes in the table footer.

Fixes: https://github.com/rstudio/gt/issues/1554


images/man_cells_body_1.png
olivroy, line 1: The new table looks a bit weird, the numbers should all be aligned? Is this intentional?

rich-iannone: Could you specify which image doesn't look right to you?

olivroy: image

rich-iannone: Thanks for showing this (because I thought you meant something else). Yeah, I don't like it and it needs a fix. This problem can happen everywhere there's a superscript so I need to find a pretty universal solution. Could you open a separate issue for this?

olivroy: Yeah, it seems pretty complicated. Thanks for all this great work. It is overall very awesome.

PR #1624, opened 2024-04-18
Fix argument name typo

olivroy:
In #1504, I made a small typo. The variable name is `call`, not `error_call`

Currently, yields this error.

! `call` must be a call or environment, not a function.


rich-iannone (approved):
LGTM!

rich-iannone:
Thanks!

olivroy:
@rich-iannone A small request if you have some time.

 I wonder if you have access to a Windows machine. If so, could you run all tests locally. Because there is a lot of noise in test output, and I am not sure if these are false positive or not. If you could take a look and skip / silence tests where appropriate, it would make contributing a little easier.

PR #1632, opened 2024-04-24
Improve on `cols_hide()` messages on edge cases and allow empty input

### LINKED ISSUE #1631: `cols_hide()` has a documentation issue and error msg
## Description

``` r
library(gt)

exibble |> 
  gt() |> 
  cols_hide()
#> Error in translate_legacy_resolver_expr(quo, null_means): l'argument "expr" est manquant, avec aucune valeur par défaut
```

<sup>Created on 2024-04-24 with [reprex v2.1.0](https://reprex.tidyverse.org)</sup>

The message when you forget to supply input to `cols_hide()` is a bit ugly and the docs is misleading.

Docs are this currently

```r
Columns to target

⁠<column-targeting expression>⁠ // default: everything()
```

## Expected result

I think that the documentation should be updated and say that columns must be supplied.

I also think it would be better to use `rlang::check_required()` in `cols_hide()` to avoid an ugly message.

I don't think having the documented `everything()` as a default is a good idea.

I think the implentation of the function is (almost) okay.

olivroy:
# Summary

Add `cols_hide()` and put better user-facing messages when something goes wrong.

## Outstanding 

`exibble |> gt() |> cols_hide(everything())` still gives an ugly error, but I will leave that to you!

Edit: I just discovered that it worked as expected in gt 0.10.0, but stopped working in gt 0.10.1. 

It now gives Error in has_rtl[i, ] : subscript out of bounds

Not affected by this PR.

## Question

Is there a way to bench mark gt?

```r
# Rough testing proved that my change improved speed, but not convinced..
gt(exibble) |> cols_hide(columns =  fctr)


```

Improved messages

```r
# before
gt(exibble) |> cols_hide(columns = fcd) 
#> Error in resolve_cols_i(expr = { : Can't select columns that don't exist.
#> ✖ Column `fcd` doesn't exist.

# this PR
Error in `cols_hide()`:
! Can't select columns that don't exist.
✖ Column `fcd` doesn't exist.

```

- fixes #1631 

# Checklist


rich-iannone:
I found that the `cols_move*()` fns had similar issues with incorrect documentation. I pushed a commit that applies the same behavior as `cols_hide()` but I'm thinking now that those functions should instead error if there is nothing supplied to `columns` (the last commit here returns data unchanged).

Do you think there should be shared behavior between the column moving and hiding functions?

olivroy:
I like having  `cols_hide()` working (it worked in 0.10.0)

because 

```r
exibble |>
  gt()
  cols_hide(
    fctr
  )
```
But then if you change your mind, you can just comment it out.

That's actually how I came across this bug.

```r
exibble |>
  gt()
  cols_hide(
    # fctr
  )
```

I have never used `cols_move()`. I tend to use `relocate()` in dplyr steps instead.

However, as a reference, `data |> relocate()` does nothing.

So, I like how you handled it in the last commit.

rich-iannone (approved):
LGTM!

olivroy:
I know it is difficult without a reprex, but do you have any clue on where to start for  #1635. It is actually the reason why I made these contributions today.

It has given me headaches for almost a week now. I am still pretty much clueless (apart from using 0.10.0, which I don't really like)

If anything of what I mentioned there rings a bell, please let me know!

Issue #1481, opened 2023-11-06
A test is failing


olivroy:
A date-time test is failing currently on R 4.3.1 on Windows. gt 0.10.0 and dev version. Happy to provide more info, but I don't really understand the bug.

``` r
data_tbl <-
    dplyr::tibble(time = c(
        "12:35:23", "15:01:34", "09:45:23", "01:32:00"
    ))
gt::gt(data_tbl) |> 
    gt::fmt_time("time", time_style = 3)
#> Error in format.POSIXlt(as.POSIXlt(x, tz), format, usetz, ...): output string exceeded 2048 bytes
```

<sup>Created on 2023-11-06 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

<details style="margin-bottom:10px;">
<summary>
Session info
</summary>

``` r
sessioninfo::session_info()
#> ─ Session info ───────────────────────────────────────────────────────────────
#>  setting  value
#>  version  R version 4.3.1 (2023-06-16 ucrt)
#>  os       Windows 10 x64 (build 19044)
#>  system   x86_64, mingw32
#>  ui       RTerm
#>  language (EN)
#>  collate  French_Canada.utf8
#>  ctype    French_Canada.utf8
#>  tz       America/Toronto
#>  date     2023-11-06
#>  pandoc   3.1.1 @ C:/~/AppData/Local/Programs/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)
#> 
#> ─ Packages ───────────────────────────────────────────────────────────────────
#>  package     * version date (UTC) lib source
#>  bigD          0.2.0   2022-09-05 [1] CRAN (R 4.3.0)
#>  cli           3.6.1   2023-03-23 [1] CRAN (R 4.3.0)
#>  digest        0.6.33  2023-07-07 [1] CRAN (R 4.3.1)
#>  dplyr         1.1.3   2023-09-03 [1] CRAN (R 4.3.1)
#>  evaluate      0.23    2023-11-01 [1] CRAN (R 4.3.2)
#>  fansi         1.0.5   2023-10-08 [1] CRAN (R 4.3.1)
#>  fastmap       1.1.1   2023-02-24 [1] CRAN (R 4.3.0)
#>  fs            1.6.3   2023-07-20 [1] CRAN (R 4.3.1)
#>  generics      0.1.3   2022-07-05 [1] CRAN (R 4.3.0)
[...]

PR #1634, opened 2024-04-24
silence tests on Windows

olivroy:
# Summary

Since the logs are noisy due to file not found.

Now, the only remaining failures in tests is #1481, which is genuine, but with this PR, it works better to run tests locally on windows.

The image snapshot error is annoying too, so I skipped it on windows. I opened r-lib/waldo#191 for options.

Basically, render related warnings are silenced on Windows. I would recommend you investigate a bit more if you have time. But gt is generally working well. I think it is safe to ignore these warnings when running tests locally.

It will make contributing in the future a bit easier!

These are mostly in .Rbuildignore, so won't change much. They are only run on CI on Ubuntu. So the only way to get that is locally on Windows.

I also took advantage to do some test refactoring to make the warning testing a bit more explicit

# Checklist


rich-iannone (approved):
LGTM!

PR #1639, opened 2024-04-25
Update condition

olivroy:
Test-coverage action seems much slower since yesterday. Could this be the reason why? https://github.com/rstudio/gt/actions/workflows/test-coverage.yaml

this condition is 4x faster, so will see if it helps


olivroy:
Was not the issue. Doesn't matter too much. Maybe it is just the change to R 4.4?

Issue #1659, opened 2024-05-13
Color columns based on another column


kwakuduahc1:
## Prework

* [*] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [*] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.
* [*] For any problems you identify, a [minimal reproducible example](https://www.tidyverse.org/help/) so the maintainer can troubleshoot. A reproducible example is:
    * [*] **Runnable**: post enough R code and data so any onlooker can create the error on their own computer.
    * [*] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
    * [*] **Readable**: format your code according to the [tidyverse style guide](https://style.tidyverse.org/).

## Question

What would you like to know?

How do I color other columns based on the value of another column? The issue is described in #1119 but I could not find a description of how to do that and I could not find it in the documentation.

Below is a how it should work

```
mtcars  %>% 
  gt() %>% 
  data_color(
  # Help here
   # color **disp and hp** columns where __cyl__ > 4
  )
```
Thank you in advance.

olivroy:
`data_color()` is quite powerful, I recommend you reading the examples in the documentation https://gt.rstudio.com/reference/data_color.html

```r
# To color disp and hp colors, when cyl > 4, based on their respective values
gt::gt(
  mtcars
) |> 
  gt::data_color(
    columns = c(disp, hp), # columns to color
    rows = cyl > 4, # rows to color
  )
```
![image](https://github.com/rstudio/gt/assets/52606734/6f95c987-62cb-4644-83a4-8270f412aa6f)

```r
# To color disp and hp colors, based on the value of column mpg
gt::gt(
  mtcars
) |> 
  gt::data_color(
    columns = c(mpg),  # colored based on mpg
    rows = cyl > 4, # rows to color
    target_columns = c(disp, hp) # columns to color
  )
```
![image](https://github.com/rstudio/gt/assets/52606734/ddd2b0d8-4bab-41bf-9a43-144c4f3ea569)

The function allows for more customization and options too

kwakuduahc1:
[...]

PR #1660, opened 2024-05-13
Error early if rows could not resolve to anything in data_color()

olivroy:
# Summary

Consider this,

```r
mtcars |>
  gt() |>
  data_color(rows = cyl == 5)
```

Currently, this errors with a cryptic error:

```r
Error in `get_contrast_ratio()`:
! At least one color defined for each of
  `color_1` and `color_2`.
```

I have left a review with an alternative approach! Please let me know which one you think is better

Discovered while answering #1659 

# Checklist


olivroy (commented):
Alternative PR! (would not be a breaking change, since this currently errors)

Maybe less suprising this way

rich-iannone:
This is a bit tricky API-wise. I like the erroring if you supply a specific row index, row name, or set of them. However, if providing an expression it would be nice to no-op, which is just like the behavior for columns. For rows, I don't know that we currently make the distinction between expression and explicit selection. Also, I don't know if this is the current behavior for other functions with a `rows` argument (mostly `fmt_*()` and IIRC some `cols_*()` fns too).

So, it's worth discussing this a little! I know what I proposed would be a breaking change but I hope it would be a welcome one that fixes some inconsistencies.

olivroy:
Good idea!

In this case, let's go with the alternative PR (that is the status quo)? and I will open a new issue to do that in a folow-up PR.

NEWS.md
```diff
@@ -1,5 +1,7 @@
 # gt (development version)
 
+* `data_color()` throws a more informative error message if `rows didn't resolve to anything (@olivroy, #1659).
```
rich-iannone, line 3:
```suggestion
* `data_color()` throws a more informative error message if `rows` didn't resolve to anything (@olivroy, #1659).
```

rich-iannone:
Disregard the above suggestion, as we're going with a different PR (feel free to close this one).

olivroy:
Will you open a new PR?  The "alternative" PR is this PR in its current state, after I accepted the suggestions https://github.com/rstudio/gt/pull/1660#pullrequestreview-2053232141.

My suggestion would be to merge this which basically just changes

```r
# before
mtcars |>
  gt() |>
  data_color(rows = cyl == 5)

Error in `get_contrast_ratio()`:
! At least one color defined for each of
  `color_1` and `color_2`.

# this PR
mtcars |>
  gt() |>
  data_color(rows = cyl == 5)

Error in `data_color()`:
!  `rows` resulted in an empty selection.
```

And address outstanding issue in a follow-up (issue #1665 )?

Does that make sense?

rich-iannone:
Yes, it makes sense! (Sorry for the confusion here.)

rich-iannone (approved):
LGTM!

PR #1669, opened 2024-05-19
[wip] Shorten help for `rows` and `columns` arguments.

olivroy:
Not yet ready for review !

# Summary

The first goal was to remove the mention the tidyselect::one_of (superseded by any_of + all_of.

Than, I realized there was a bit of duplication, so I looked into that.

* When the default is `everything()`, shorten how the default is explained + don't link to `everything()` twice in man
* Don't link to `c()` too often.
* Personally, I have never used tidyselect helpers to select rows, (an expression is almost always more useful, so I want to highlight that), and maybe reduce prominence in man.
* Possibly a topic would be more appropriate where all the ways to use `rows` and `columns` is explained (since that knowledge applies to all of gt, and not to any particular functions.
* Ideally, I think it would be great to link to the topic, and only explain if there are exception to usage in that particular function (i.e. change of default, optional vs required etc.)
* I may close this and reopen a PR where I gather data on how it is documented currently.  to use `@inheritParams` safely without changing anything in man. than reopen a PR with the same intention as this one, but with minimal changes

I am opening to check-out my progress. 

I will review shortly to ensure consitency, and will use `@inheritParams` as much as possible, unless some special info is shared.

i.e. in `cols_hide()`, I think it is worth mentioning how to hide columns with all missing values. 8901a1f461a352b2473d057b5e6640ea88d6d05a

Before requesting review, I want to double-check that the style used is consistent. 

I used https://github.com/r-lib/roxygen2/pull/1594 to document, for faster iteration.


olivroy:
Closing for now. Will maybe come back to this once doc duplication is reduced. (i.e. more PRs like #1676 )

PR #1694, opened 2024-06-05
Refactor documentation and test nomenclature

olivroy:
Sorry.. this got out of hand again. Please let me know if you agree with this. If not, I can send smaller separate PRs of the relevant bits!

TL;DR: The goal is basically to simplify the `x()` function to `x()` where applicable as recommended in the [documentation tidyverse style guide](https://style.tidyverse.org/documentation.html#r-code). (only the function bullet)

This PR started as a follow-up to #1683 as I wanted to make clarifications and improved some things I did there after taking a fresh look.
cbb1dca8179c371d260190568eac7d7add1e746e and 018b232b793e99f50856bf7232c80336fa0d1c8d do this. I finally discovered the way to avoid spurious ([snapshot](https://github.com/r-lib/testthat/issues/1958) git diffs while working on Windows: Set Posix line ending in Rproj (13b2666ad31fe955c132ac6574017d9b3cba45c6). Used https://www.tidyverse.org/blog/2020/03/roxygen2-7-1-0/ as ref too.

I'd like to get your views on how I edited the shiny related functions docs 64e80b348b61087a530ab3ecb37c997b93943293. (I removed details that felt redundant? Let me know if you agree. Otherwise, I will rework it!

Changes 
* Renamed `is_non_empty_string()` to `is_non_empty_chr()` for clarity (accepts length > 1) e0c4a10e11b4e4c7b77c7376ea304c286efb74c2 (string is usually for character + length = 1)
* renamed tests to use the convention ab0ed09f8142a9e11a675e1605a21da437f9a453

* Moved some test helper functions into own [test helper files](https://testthat.r-lib.org/articles/special-files.html#helper-files) to improve testing workflow (i.e. functions are loaded with `load_all()` af9cdd6
* Removed duplicate definition of test helper in tests/testthat/helper.R bcbb80dafedb023d53f249630328d0f3ee791f91
* Removed the check for xml2 installation since it is import now (I may do a follow-up where I remove `check_suggests()` with appropriate `skip_if_not_installed()` bcbb80dafedb023d53f249630328d0f3ee791f91
* Did minor test refactoring where I saw it useful bcbb80dafedb023d53f249630328d0f3ee791f91
* simplified some dplyr expressions in examples d2aa113762cefe65bc081e130c34f4a5f9ae4f6b, bc70cdb4b10310a594863a00c75890811d22caff

<details><summary>I verified that they are all correct with this code
</summary>
<p>

```r
v1 <-  towny |>
[...]

PR #1700, opened 2024-06-07
Edits pkgdown + fixup of some docs

olivroy:
Regression from #1694 caught by skimming through package index.

Also renamed the locations topic to include a link to main. Adding `text_case_when()` to the list of functions that use `locations`.

~Somehow you can add links in `@family` tags! but `pkgdown::has_concept()` doesn't really recognize it (but that is okay, we can just revert to using `starts_with("cells_")`, like it was before #1658~  Reverted this bit with `gert::git_reset_hard("HEAD~2")` ! (advice from https://github.com/r-lib/pkgdown/issues/2647#issuecomment-2155442376)

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -301,7 +301,7 @@
 
 * The `docx` output format is now better detected in R Markdown and Quarto (#1040). (#1084, thanks @cderv!)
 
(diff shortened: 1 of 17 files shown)


rich-iannone (approved):
This is great, thank you!

Issue #1510, opened 2023-12-17
Setting cols_width causes column to "disappear"


schwa021:
## Description

I can set the column widths of all but one column. When I set the last column width a column disappears from the table. The reprex and images below will make this more understandable.

## Reproducible example

I have the following tibble named `tdat ` that has columns (Model, name, err_Treated, ..., coverage_Control)

**Model,	name,	err_Treated,	err_Control,	width90_Treated,	width90_Control,	coverage_Treated,	coverage_Control**
Match, A, 12, 8, 5 ,36, 0.91, 0.82
Match, B,	5,	6,	29,	25,	0.87,	0.81
Match, C,	10,	9,	42,	41,	0.79,	0.82
BART, A,	7,	7,	37,	37,	0.89,	0.87
BART, B,	6,	5,	26,	26,	0.87,	0.9
BART, C,	9,	9,	49,	48,	0.93	,0.87

(sorry - I don't know of a better way to reprex this)

I use the following code to create and format a gt() table:
 ``` r
  tbl_cover <-
    tdat |>
    gt(groupname_col = "name") |>
    cols_label(
      err_Treated = "Treated",
      err_Control = "Control",
      width90_Treated = "Treated",
      width90_Control = "Control",
      coverage_Treated = "Treated",
      coverage_Control = "Control"
    ) |>
    fmt_percent(columns = starts_with("cover"), decimals = 0) |>
    # tab_header(str_replace_all(s, "_", " ")) |>
    tab_style(
      cell_text(weight = "bold", align = "left"),
      locations = list(cells_row_groups(), cells_title())
    ) |>
    tab_style(cell_text(align = "center"),
              locations = list(
                cells_body(columns = everything()),
                cells_column_labels(columns = everything())
              )) |>
    tab_style(cell_text(style = "italic"),
              locations = list(
                cells_column_labels(columns = matches("width|cover|err")),
                cells_body(columns = Model)
              )) |>
    tab_style(cell_text(align = "left", indent = px(25)),
              locations = cells_body(columns = Model)) |>
    tab_style(cell_text(color = "transparent"),
              locations = cells_column_labels(columns = Model)) |>
    cols_width(
      # err_Treated ~ px(100),
      err_Control ~ px(100),
      starts_with("cover") ~ px(100),
      starts_with("width") ~ px(100),
      Model ~ px(230)
    ) |>
    tab_spanner(label = "Mean Abs. Error",
                columns = starts_with("err")) |>
[...]

PR #1706, opened 2024-06-14
Add support for `rowname_to_stub` + `opt_interactive()`

### LINKED ISSUE #1702: Rownames not showing when gt is set to interactive
I would like rownames to appear on my gt table while being able to interactively sort and navigate through pages, but whenever I make the gt object interactive the rownames go away. 

Here is a simple reproducible example:

```
library(tibble)
library(gt)

gt(iris, rownames_to_stub = T)

gt(iris, rownames_to_stub = T) %>% 
  opt_interactive()
```

Is this result intentional or a bug?

olivroy:
fixes #1702

~Do you think it is worth adding a new test?~

```r
gt(iris, rownames_to_stub = T) %>% 
    opt_interactive()
```

~Maybe the width is too wide? I don't know how to tweak it.~

This PR

image

Before 

image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # gt (development version)
 
+* `opt_interactive()` now shows row names if `rownames_to_stub = TRUE` (@olivroy, #1702). 
+
 * `data_color()` throws a more informative error message if `rows` didn't resolve to anything (@olivroy, #1659).
 
 * PDF output now allows the font size of a table to be set using the table.font.size parameter in the tab_options function (#1472).  The font sizes of individual table cells (including those in the body, stubs, column headings, etc.) can be set using tab_style function. Several other options specified in tab_style are now reflected in PDF output.
--- a/R/render_as_i_html.R
+++ b/R/render_as_i_html.R
@@ -183,6 +183,8 @@ render_as_ihtml <- function(data, id) {
   column_labels_border_bottom_width <- opt_val(data = data, option = "column_labels_border_bottom_width")
   column_labels_border_bottom_color <- opt_val(data = data, option = "column_labels_border_bottom_color")
 
+  rownames_to_stub <- stub_rownames_has_column(data)
+
   emoji_symbol_fonts <-
     c(
       "Apple Color Emoji", "Segoe UI Emoji",
@@ -452,7 +454,7 @@ render_as_ihtml <- function(data, id) {
       data = data_tbl,
       columns = col_defs,
       columnGroups = colgroups_def,
-      rownames = NULL,
+      rownames = rownames_to_stub,
       groupBy = NULL,
       sortable = use_sorting,
       resizable = use_resizers,


rich-iannone:
This is great! I wouldn't worry about the width. There's a default width here that can be set for all widths in `tab_options()`, plus you can individually set them with `cols_width()`. The width rules are different in interactive tables because the underlying framework constructs a table without the `<table>` element, so there is no automatic sizing based on content.

rich-iannone:
Also, I wouldn't worry about the adding of a test here.

rich-iannone (approved):
LGTM!

rich-iannone:
I've reviewed and approved. If there is nothing more to change here, I'll merge once all tests pass.

olivroy:
Thanks for the review. All good on my side, but let's wait for test-coverage to pass! I have tried to address #1706, but realized it was not as straight forward as I had imagined. I may take another look later on.

rich-iannone:
Thanks for taking on these bugs! If you're interested in working on https://github.com/rstudio/gt/issues/1510, it's along the same lines (but not interactive) and also a pretty serious bug. I suspect the issue there is that we are not supplying enough column width values and columns get dropped because of that.

olivroy:
Hi @rich-iannone.

I may try again later, but here is the code I used in case it is useful to you

```r
tb <- mtcars %>% 
  dplyr::select(mpg, cyl, vs) %>% 
  head() %>% 
  gt(groupname_col = "vs", row_group_as_column = T)

one_width <- tb %>% 
  cols_width(
    mpg ~ px(100),
  )

all_widths <-  tb %>% 
  cols_width(
    mpg ~ px(100),
    cyl ~ px(100)
  )
#' Somehow the only thing inside the gt object that is not different  
waldo::compare(
  one_width,
  all_widths
)
waldo::compare(one_width %>% render_as_html(), all_widths %>% render_as_html(), x_arg = "working", y_arg =  "bug", max_diffs = Inf, )


one_width %>% render_as_html() %>% 
  .[1] %>% 
  stringr::str_split_1(pattern = "\n") %>% 
  .[1:8]

all_widths %>% render_as_html() %>% 
  .[1] %>% 
  stringr::str_split_1(pattern = "\n") %>% 
  .[1:8]

#' # As a comparison, with `row_group_as_column = FALSE`
#' 
tb2 <-  mtcars %>% 
  dplyr::select(mpg, cyl, vs) %>% 
  head() %>% 
  gt(groupname_col = "vs", row_group_as_column = F)


one_width2 <- tb2 %>% 
  cols_width(
    mpg ~ px(100),
  )

all_widths2 <-  tb2 %>% 
  cols_width(
    mpg ~ px(100),
    cyl ~ px(100)
  )
#' Somehow the only thing inside the gt object that is not different  

# same as difference between one_width and all_widths
waldo::compare(one_width2 %>% render_as_html(), all_widths2 %>% render_as_html(), x_arg = "one_width", y_arg = "all_widths")

# same as the comparison above, so the issue is not in how gt sets col widths, the issue lies in the render html part.
waldo::compare(
  one_width2,
  all_widths2
)
waldo::compare(one_width2 %>% render_as_html(), all_widths2 %>% render_as_html(), x_arg = "one_width", y_arg = "all_widths")


# not very useful
waldo::compare(
  all_widths2 %>% render_as_html(),
  all_widths %>% render_as_html(),
  x_arg = "working",
  y_arg = "bug",
  max_diffs = Inf
)

# not very useful.
waldo::compare(
  one_width2 %>% render_as_html(),
  one_width %>% render_as_html(),
  x_arg = "working",
  y_arg = "working row_group_as_col"
)

```

Next steps for checking,

* `cols_width()` is not the culprit here.
* There may be a `<col/>` missing?
* There are 2 consecutive `;` which means that something wrong is written
* The issue may be further than anticipated when doing the gt table

The issue may be caught more by a general html validator as locally, the rendered code doesn't tell us much...

Do you have tip on which helper function's output I should look at to really find out what is going wrong?

rich-iannone:
Thanks for looking into this one. For that issue, I think the problem is within `get_table_defs()`. We can continue the discussion there. As for this PR, I'll merge it now!

Issue #1292, opened 2023-04-04
tidyselect error in `gt::summary_rows()` without groups (in v0.9.0)


salim-b:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

See the error message. 

## Reproducible example

``` r
mtcars |>
        dplyr::select(gear) |>
        gt::gt(rownames_to_stub = TRUE) |>
        gt::summary_rows(fns = "sum")
#> Error:
#> ! `everything()` must be used within a *selecting* function.
#> ℹ See <https://tidyselect.r-lib.org/reference/faq-selection-context.html> for
#>   details.
#> Backtrace:
#>     ▆
#>  1. └─gt::summary_rows(...)
#>  2.   └─tidyselect::everything()
#>  3.     ├─vars %||% peek_vars(fn = "everything")
#>  4.     └─tidyselect::peek_vars(fn = "everything")
#>  5.       └─cli::cli_abort(...)
#>  6.         └─rlang::abort(...)
```

<sup>Created on 2023-04-04 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

## Expected result

I'd expect the above expression to return the same as `gt::grand_summary_rows(fns = "sum")`

<details><summary>Reprex:</summary>

``` r
mtcars |>
        dplyr::select(gear) |>
        gt::gt(rownames_to_stub = TRUE) |>
        gt::grand_summary_rows(fns = "sum")
```

<div id="toywfipzsh" style="padding-left:0px;padding-right:0px;padding-top:10px;padding-bottom:10px;overflow-x:auto;overflow-y:auto;width:auto;height:auto;">
<table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
  <thead>
    &#10;    <tr class="gt_col_headings">
      <th class="gt_col_heading gt_columns_bottom_border gt_left" rowspan="1" colspan="1" scope="col" id=""></th>
      <th class="gt_col_heading gt_columns_bottom_border gt_right" rowspan="1" colspan="1" scope="col" id="gear">gear</th>
    </tr>
  </thead>
  <tbody class="gt_table_body">
[...]

PR #1707, opened 2024-06-14
try to make summary rows work [wip]

olivroy:

- Ref: #1292


olivroy:
After all, I believe it would be better to have an error message that points to using `grand_summary_rows()` instead of making this work, so will leave it untouched.

PR #1711, opened 2024-06-17
fix usage of `rowname_col` for `opt_interactive()`

olivroy:
Addresses https://github.com/rstudio/gt/issues/1702#issuecomment-2173258752

After a bit of investigation, I also realized that setting `.rownames` doesn't work. https://github.com/glin/reactable/issues/378

Let me know if you can think of a better way to achieve this. 

This is a bit sketchy, because 

```r
row.names(mtcars) <- mtcars$vs
#> Error 
#> Duplicates in row.names are not allowed.
```

before #1706, it would be silently ignored.

### DIFF
--- a/R/render_as_i_html.R
+++ b/R/render_as_i_html.R
@@ -119,8 +119,10 @@ render_as_ihtml <- function(data, id) {
       "*" = "Failing that, look at whether all columns have been inadvertently hidden."
     ))
   }
+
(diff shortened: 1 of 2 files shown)


rich-iannone:
This is a pretty good fix!

rich-iannone (approved):
LGTM!

PR #1714, opened 2024-06-18
Optionally skip tests requiring 'rvest'

MichaelChirico:


rich-iannone:
This is fixed by https://github.com/rstudio/gt/pull/1718

Issue #1544, opened 2024-01-19
interactive - provide sticky headers


slodge:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

With interactive reactable tables sticky headers seem to come by default.

With gt I haven't been able to easily find a way to get sticky headers...

## Reproducible example

Example:
```
library(tidyverse)
library(gt)
library(reactable)
library(shiny)

g_output <- gt_output("iris_table")
g_output <- g_output |> 
  htmltools::tagAppendAttributes(style="height:400px;")

r_output <- reactable::reactableOutput("react_table", height = "400px")

ui <- fluidPage(
  shiny::fluidRow(
    column(width=4,
           h4("Great"),
           tags$div(id="great",
                    style="height:400px;",
                    g_output)
           ),
    column(width=4,
           h4("reactable"),
           tags$div(id="react",
                    r_output)
    )
  )
)

server <- function(input, output, session) {
  output$iris_table <- render_gt({
    iris |> 
      gt() |> 
      tab_options(
        container.height = px(400)
      ) |> 
      opt_interactive(use_pagination = FALSE)
  })
  
  output$react_table <- reactable::renderReactable({
    iris |> 
      reactable::reactable(pagination = FALSE)
  })
}

shinyApp(ui, server)
```

This gives me non-sticky headers in the gt half of the experiment:
![image](https://github.com/rstudio/gt/assets/533662/042897ed-a0f6-4efe-b1bd-d397735a2092)

## Expected result

This example (with some jquery runtime hacking) works more as expected:
```
library(tidyverse)
library(gt)
library(reactable)
library(shiny)

g_output <- gt_output("iris_table")
g_output <- g_output |> 
  htmltools::tagAppendAttributes(style="height:400px;")

r_output <- reactable::reactableOutput("react_table", height = "400px")

js_table_hack <- "$(document).ready(function() {
  $('#great').on('DOMNodeInserted', function(e) {
    const t = $(e.target);
    if (t.hasClass('ReactTable')) {
      //console.log('changing the height of an element...');
      //    console.log(t);
      t.css('height','100%');
      t.parent().css('height','100%');
    }
  });
[...]

PR #1723, opened 2024-06-19
Add `ihtml.height` for interactive custom widget height

olivroy:
To mitigate #1544, specifying manually height could be useful.

Before, there was no way to prevent overflowing with  `opt_interactive()`. Let me know if there is a better way to do this. I tried modifying the `height` value to `NULL`, or `"auto"` in the `reactable::reactable()` call in `render_as_ihtml()`, but couldn't find a way to do this well. The only way I found was with manual override. Figure a way to find out the RStudio viewer's size. But after my research, I figured that giving the option to user exactly how to set their widget's height was a good way to go?

I am not too familiar with setting options, and this will likely ignore static options set. I don't know which `tab_options()` are expected to have an effect in both interactive and non-interactive?

Before:

```r
mtcars %>% gt() %>% opt_interactive()
```
image

This PR:

```r
mtcars %>% gt() %>% opt_interactive(height = 300)
```
image

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,7 @@
 # gt (development version)
 
+* `opt_interactive()` gains `height` to help specify your widget's height (@olivroy, #1544).
+
 * `opt_interactive()` now shows row names if `rownames_to_stub = TRUE` (@olivroy, #1702). 
 
 * `data_color()` throws a more informative error message if `rows` didn't resolve to anything (@olivroy, #1659).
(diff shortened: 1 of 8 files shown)


rich-iannone:
This is actually ideal! There is `container_height` but it doesn't quite fit, and I think a scoped `ihtml.*` property is better here.

rich-iannone (approved):
LGTM!

PR #1731, opened 2024-06-25
Depend on R 3.6 + other package cleanup

olivroy:
Ready for review! I don't think I will have anything more to add

# Summary

Depend on R 3.6 to be able to use features of R and remove workarounds. Not sure if gt fully worked on older versions. `isTRUE()` is already used in certain places for example 

* allows for removal of `tidy_sub()` and `tidy_gsub()` (for consistency in the codebase, it is possible to simply use `gsub()` and `sub()` in most cases.
* Tested that the `tidy_sub()`, `tidy_gsub()` test passed 
* Remove `is_false()` (`isFALSE()` introduced in R 3.5.)
* Allows for removal of `register_s3_method()`, see https://www.tidyverse.org/blog/2024/01/roxygen2-7-3-0/ for why/ (fiddle a bit with `as.tags()` in 8b7f9d8

Unrelated

* Don't ignore snapshots that are not ignored on CRAN in 64d0bb2 (to actually test that output is consistent across versions)
* Move `tidy_grepl()` to tests/testthat/helper.R since unused in code. https://testthat.r-lib.org/articles/special-files.html
* Change `gsub()` to `trimws()` for clarity dd5a47188eba1dc795b89c6135f6792d1bb434d9
* Change `gsub()` begin end to `paste0()` for clarity e93c5db.
* Comment about names not NULL. Maybe the condition is incorrect, and this may relate to #1510 (maybe not)
* Change of option. 18c10b5ff378d27d187ff2499dae4d4a3d065010 (I just find it clearer this way, let me know if you disagree)
* Refactoring in 78e6a296ff882a37a11ed6b233d231cb151c05a7 (caused previous build failures, resolved with 90337a0)
* Rename function for clarity in 02c97e3 to allow removing comments everywhere, and synthetize it in the helper function instead (since the comment is unrelated to individual functions)
* Simplify some dplyr code in 8da30ca
* Was able to [remove a workaround](2ee758d) introduced in https://github.com/rstudio/gt/pull/1634  due to edbd9f49c6c79e878c75dfae8d1bf7fe7f1f088d
* Remove Collate and move imports to `gt-package.R` to avoid repetition

# Checklist

FYI, the hard-to-debug failure was due to  https://github.com/r-lib/testthat/issues/1646


rich-iannone:
Thanks for all the great work! This is a nice round of cleanups and I agree with everything done here. Will review and merge shortly.

rich-iannone (approved):
LGTM!

PR #1742, opened 2024-07-03
Add `plot.gt_tbl()`

olivroy:
Just so that you can quickly do `exibble |> gt() |> plot()` for a reprex visual example.


rich-iannone (approved):
LGTM!

Issue #1705, opened 2024-06-14
`groupname_col` disappears with `opt_interactive()`


olivroy:
With gt 0.10.0.9000,

we can have 

```r
d <- mtcars |> dplyr::slice_head(n = 10)
```

| row_group_as_columns = FALSE | row_group_as_columns = TRUE |
|--------|--------|
| `gt::gt(d, groupname_col = "vs")`  |  `gt::gt(d, groupname_col = "vs", row_groups_as_columns = TRUE)` |
| ![image](https://github.com/rstudio/gt/assets/52606734/21230cf6-752d-4c05-93bd-dd573579787f) |  ![image](https://github.com/rstudio/gt/assets/52606734/7b0ba762-8347-4812-b792-44c628c387e8) |
 
Currently, `opt_interactive()` hides the `vs`  `groupname_col` column regardless.

## Here is my suggestion with `reactable::reactable()`

```r
# use `colDef()` to hide column name
# use rownames = FALSE since we are using a different column as row names
# From https://github.com/glin/reactable/issues/94
grp_fn  = reactable::JS("
      function(cellInfo) {
        return cellInfo.value
      }")

mtcars |> dplyr::slice(1:10)  |>reactable::reactable(groupBy = "vs",defaultExpanded = T, rownames = FALSE, columns = list(vs = reactable::colDef(name = "", grouped = grp_fn)))
```
![image](https://github.com/rstudio/gt/assets/52606734/c4200094-1b36-42ce-8b5c-3b649d6b104c)

I don't see a good solution for `row_group_as_columns = T` as I couldn't find an example of reactable merging a column, but would still be an improvement.

PR #1758, opened 2024-07-08
Respect `tab_stubhead()` in `opt_interactive()`+ fix rendering for missing row names

olivroy:
Follow-up to #1702, #1705. Fixes some rough edges for new feature introduced in #1711 and #1725

Intent is respecting `tab_stubhead()` in `opt_interactive()`, by adding `name` to groupname colDef() or rownames where appropriate.

I noticed that `sub_missing()` (and probably all `sub_*()` is ignored in `opt_interactive()`, but that will require more work another time!

For clarity, I also factored out the translations since the intermediate variables are not useful for the final output.

Before this PR, 
image

```r
exibble |> gt::gt(rowname_col = "num", groupname_col = "group") |> tab_stubhead("row name to show") |> opt_interactive()
```

With this PR, stubhead label shows and NAs are correctly marked as NA instead of `...6`

### DIFF
--- a/R/render_as_i_html.R
+++ b/R/render_as_i_html.R
@@ -58,34 +58,48 @@ render_as_ihtml <- function(data, id) {
   # Generate a `lang_defs` object to pass to the `language` argument
   lang_defs <- get_ihtml_translations(locale)
 
-  } else {
+  column_groups <- dt_boxhead_get_vars_groups(data = data)
+  # if dt_boxhead_get_vars_groups is fixed, this will no longer be necessary.
+  if (identical(column_groups, NA_character_)) {
+    column_groups <- NULL
   }
+  rownames_to_stub <- stub_rownames_has_column(data)
+  # value to use for rowname_col or groupname_col title
+  # Will use it for rowname_col only if groupname_col is undefined.
+  # Will use it only for the first groupname_col.
+  stub_label <- dt_stubhead_get(data)$label
+
+  if (is.null(stub_label)) {
+    # No label if stubhead label is undefined.
+    rowname_label <- ""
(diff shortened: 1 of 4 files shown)


R/render_as_i_html.R
```diff
@@ -55,94 +55,80 @@ render_as_ihtml <- function(data, id) {
...
   if (rownames_to_stub) {
     rowname_col <- dt_boxhead_get_var_stub(data)
     if (length(rowname_col) == 1) {
       # avoid base R error when setting duplicate row names.
-      attr(data_tbl, "row.names") <-  as.character(data$`_data`[[rowname_col]])
+      row_names <-  as.character(data$`_data`[[rowname_col]])
+      # Convert to NA string to avoid wrong output.
+      # TODO figure out if there is a way to get the sub_missing value.
+      # With data$`_substitutions`
+      row_names <- dplyr::coalesce(row_names, "NA")
```
olivroy, line 114:
```suggestion
      row_names <- dplyr::coalesce(row_names, "")
```

For consistency with `reactable()` in general, better to display nothing and we can fix this in the future with #1759

R/render_as_i_html.R
```diff
@@ -307,37 +293,47 @@ render_as_ihtml <- function(data, id) {
...
+        # Use the stubhead label for the first group
+        group_label <- groupname_label
+      } else {
+        # by default, don't name groupname_col for consistency with non-interactive
+        group_label <- ""
+      }
+
+      group_col_defs[[i]] <-
+        reactable::colDef(
+          name = stub_label,
```
olivroy, line 306:
```suggestion
          name = group_label,
```

PR #1760, opened 2024-07-08
actually fix NA row names in `opt_interactive()`

olivroy:
I should have tested before comitting this https://github.com/rstudio/gt/pull/1758#discussion_r1668777820...

If it is not too late, this version actually solves the problem I was trying to solve.


rich-iannone (approved):
Thanks!

PR #1774, opened 2024-07-12
Improve `text_*()` error messages

olivroy:
Thought I was done with error messages, but no!

I started this PR as a wish to improve test coverage.

But turns out more work was required to get the `text_*()` to have consistent error messages (similar to `tab_style()` and `tab_footnote()`.

While working on that, I discovered an error in tidyselect (r-lib/tidyselect#352) It means that some errors will improve naturally when the next tidyselect version will be out.

---
Test coverage 

https://app.codecov.io/gh/rstudio/gt/blob/master/R%2Ftext_transform.R

I also deleted dt_col_merge_init() as it is unused in the codebase.  https://app.codecov.io/gh/rstudio/gt/blob/master/R%2Fdt_cols_merge.R

I created the `text_transform_impl()` as a way to easily get `Error in `text_case_match()`

Coverage data https://app.codecov.io/gh/rstudio/gt/pull/1774

### DIFF
--- a/.Rbuildignore
+++ b/.Rbuildignore
@@ -171,6 +171,8 @@ tests/testthat/test-table_parts.R
 tests/testthat/_snaps/table_parts.md
 
 tests/testthat/test-text_transform.R
+tests/testthat/_snaps/text_transform.md
+
 tests/testthat/test-util_functions.R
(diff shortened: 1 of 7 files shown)


rich-iannone (approved):
LGTM!

PR #1778, opened 2024-07-12
Reorganize files 

olivroy:
To have a standardized way to test for output and facilitate workflows.

This is the first step and doesn't seem to conflict with open PRs.

I plan to do more in the future, but this is a first step.

The idea is to improve workflow with `usethis::use_r()` / `use_test()`

* Standardized latex tests to the `test-l_*` prefix
* Standardized rtf tests to the `test-rtf_*` prefix (noticed 2 similar test files that I combined
* Move `fmt()` into own file along with all helper functions for formatting (previously found in utils_formatters.R


rich-iannone (approved):
Thank you!

Issue #1767, opened 2024-07-11
Make `<br>` a universal linebreak across formats


olivroy:
* Relates to https://github.com/rstudio/gt/issues/1027
* Relates to https://github.com/ddsjoberg/gtsummary/pull/1898
* #512

I think one thing to make "just work" is recognizing `<br>` as a line break in all output contexts; this'll avoid any need to call `fmt_markdown()`. This could probably be done in the `process_text()` function.

We incidentally also need a similar bit of 'extra' markup to represent cross-output tab stops but that's for another issue.

_Originally posted by @rich-iannone in https://github.com/rstudio/gt/issues/1765#issuecomment-2221639787_

---

As of gt 0.11

* \n works for grid format (<br> just shows as <br>)
* <br> works for md/html
* No linebreak seems to work for docx (<br> shows as <br> and \n acts like a space)
* Haven't looked for latex yet

For rtf 

the `text_grob` argument of `as_gtable()` can be used.

https://github.com/rstudio/gt/pull/1776#issuecomment-2225651976

ddsjoberg:
If possible, would it makes sense to make `\n` the universal line breaker, since that's how it's already interpreted in many other spaces? Then no one needs to learn another

teunbrand:
I second the suggestion to make `\n` to universal line breaker. I don't understand why `<br>` would be the preferred option.

olivroy:
I think my main reasoning is that even if gt supports all formats, it still remains an html-first format. When I construct a complex gt table, I do it interactively in html, and hope for the best when trying to convert to other formats. I am doing so blindly (too much sometimes).

* html output is easier to preview
* tested more thoroughly

So getting a LaTeX output or grid or docx is a matter of converting the html to this other format as close as possible to the html output.

Similarly, ggplot2 is grid first. Therefore, to make it interactive, `plotly::ggplotly()` interprets the grid system and translates it to html. (i.e. When I am using ggplotly, I trust that it will convert ggplot2 grid to html properly. ggplots are much easier to preview as static.)

By default, html ignores line breaks and multiple consecutive spaces.
```r
# no line break
data.frame(
  x = c("x\ny")
) |> 
  gt()
```


```r
 data.frame(x = c("x\ny") |> 
   gt() |>
  gt::opt_interactive()
```
![image](https://github.com/user-attachments/assets/58bc9511-8fa1-433a-87fa-b53dab22d740)

grid is my new second favorite for easy previewing!

```r
# line break
[...]

PR #1788, opened 2024-07-15
Improvements to grid context `as_gtable()` + `plot.gt_tbl()`

olivroy:
# Summary

* this PR adds unicode support to some `sub_*()` + `cols_merge()` functions 
* Adds support for displaying symbols for currencies
* Add support for `fmt_markdown()`
* Convert `<br>` to \n #1767 
* Add `context = "grid"` to some functions
* Also a left-over from #1781

@teunbrand I'd love to have your views here in general.

* Does this seem reasonable?
* Should we use vdiffr to capture the tables as a useful visual reference?
* I am not too sure of what I did here. My testing is mostly through trial and error.
* Any concerns with unicode?

- follow-up to #1727, 
- #1767 

@rich-iannone I am not too sure if I am duplicating things here. Not clear to me how context = "plain" vs context ="grid" should act. 

It is just that grid renders unicode formats, since R 4.2 I think? 

Do you think we may add some conditions on R version for some grid features? i.e. if context = "grid" and R version < 4.2, use plain text rendering?

## Examples

This PR resolves the following to look the same in html and grid

```r
# this works with only a few warnings!
info_currencies() |> plot()
```

```r
gt_tbl <- exibble |> 
  gt() |> 
  cols_label(
    num = md("num<br>x")
  )

plot(gt_tbl)
```

Previously, we would see `<br>`


rich-iannone (approved):
LGTM!

teunbrand:
> Does this seem reasonable?

It doesn't look *un*reasonable, but I don't really know what `markdown::mark()` or `commonmark::markdown_text()` are doing, so feel free to dismiss.

> Should we use vdiffr to capture the tables as a useful visual reference?

You can, but it is an extra suggest dependency that I'm not sure pays off here. For testing purposes the `layout` object created in the line below should have most of the relevant components (i.e. labels/layout) to test:

https://github.com/rstudio/gt/blob/34fd17f7b1845201d86d11934ec0a4260c5a5a89/R/export.R#L1194

> Any concerns with unicode?

No grid should render unicode just fine even before R 4.2.0 I think so this shouldn't be cause for worry. Potentially it could be the case that a unicode glyph is not in the font, so it might render as missing characters in devices that don't have font fallbacks. But that's just a limitation, so there is nothing that {gt} can do here.

olivroy:
Maybe I can provide an explanation here. It feels a bit magical, but since the currencies are stored in html strings, I discovered that `markdown::mark(gt:::currencies$symbol, format = "text")` transformed the html codes to their utf-8 equivalent. (This allows us showing the currency symbol instead of their unescaped equivalent) GBP for British pounds for example.

PR #1789, opened 2024-07-15
Performance improvement in options, currency and rendering

olivroy:
performance improvements for retrieving currency and for gt tbl checks.


rich-iannone (approved):
LGTM!

PR #1791, opened 2024-07-16
Setup standard pkgdown workflow + install Quarto

olivroy:
This sets the ground to adding a Quarto vignette to the site, since pkgdown supports this since 2.1 (which I am working on!)

But I thought it would be useful to run the pkgdown workflow on PRs.


rich-iannone (approved):
LGTM!

PR #1795, opened 2024-07-16
No error for `data_color()` + empty tidyselect for consistency with `fmt_*()`

### LINKED ISSUE #1665: `data_color()` should return data unchanged if a tidyselect expressions resolved to 0
This is a bit tricky API-wise. I like the erroring if you supply a specific row index, row name, or set of them. However, if providing an expression it would be nice to no-op, which is just like the behavior for columns. For rows, I don't know that we currently make the distinction between expression and explicit selection. Also, I don't know if this is the current behavior for other functions with a `rows` argument (mostly `fmt_*()` and IIRC some `cols_*()` fns too).

So, it's worth discussing this a little! I know what I proposed would be a breaking change but I hope it would be a welcome one that fixes some inconsistencies.

_Originally posted by @rich-iannone in https://github.com/rstudio/gt/issues/1660#issuecomment-2111145355_

olivroy:
# Summary

After thinking about this a bit more, you are right that if we supply a tidyselect expression to `rows` or `columns` that resolves to nothing, we should just leave the data unchanged.

This PR takes care of this and adds test to verify consistency between `data_color()` and `fmt_number()`.

fixes #1665 and amends #1660

# Checklist


rich-iannone (approved):
LGTM!

PR #1803, opened 2024-07-17
Fix grid output with unique groups in `groupname_col` and  `row_group_as_column = TRUE`

olivroy:
Split of.

Basically, just due to the fact that like base subsetting,

```r
vctrs::vec_slice(c(1, 2), -integer(0))
#> numeric(0)
c(1, 2)[-integer(0)]
#> numeric(0)
```

which ends up creating an empty layout in this particular case. I have been bitten by this before ;) 

```r
# bug before this PR
df <- data.frame(x = 1:3, y = 4:6, row = c("A", "B", "C"))
gt(df, groupname_col = "row", row_group_as_column = T) |> plot()
#> Error in body[vertical] + n : 
 #> Non-numeric argument
```

This PR:  
image

Split from #1801 as I am still trying to understand what is going on. What currently works, and what doesn't. I'd like to narrow down the issues and I thought this was a straightforward bug fix!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -27,6 +27,10 @@
 
 * Fixed a bug in using `pct()` column widths with `as_gtable()` (@teunbrand, #1771)
 
+* Fixed a bug where `gt(row_group_as_column = TRUE)` would create the wrong layout with `as_gtable()` when all groups are unique (@olivroy, #1803).
+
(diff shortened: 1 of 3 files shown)


rich-iannone (approved):
LGTM!

Issue #1675, opened 2024-05-21
Documentation of topics


olivroy:
Some concepts of gt are valuable to learn as a whole

I introduced the `locations` topic in #1658.

I believe it would be interesting to have that for other parts of gt

* [ ] Specifying `columns`
* [ ] Specifying `rows`
* [ ] `locale` (awaiting #1593)
* [ ] `date` and `time` styles

Probably others

Remove long form info from functions and includes links everywhere

PR #1808, opened 2024-07-18
Update visual-tests-latex.R to show Quarto version

olivroy:
Seems like the visual tests are off to a great start! https://github.com/rstudio/gt/pull/1804#issuecomment-2236314015. Let's add the Quarto version there as an extra layer of safety / reproducibility.

### DIFF
--- a/scripts/visual-tests-latex.R
+++ b/scripts/visual-tests-latex.R
@@ -156,8 +156,9 @@ c(
   "## --------",
   # Comment / uncomment here to compare between released version and current.
   # "library(gt)",
-  "devtools::load_all(\".\")",
+  "pkgload::load_all(\".\")",
   "packageVersion('gt')",
+  "quarto::quarto_version()",
   "",
   "#'",
   "#' {{< pagebreak >}}",


rich-iannone:
This is all really great and super essential for finding new rendering issues!

rich-iannone (approved):

rich-iannone:
Just an idea, but do you think it's possible to re-purpose the HTML tests and use the `embedpdf` Quarto extension (example at: https://jmgirard.github.io/embedpdf/native.html, repo: https://github.com/jmgirard/embedpdf)?

This would involve making PDFs for each example and generating the Quarto website with the appropriate shortcodes. Seems like a fair amount of work (and there could be some issues along the way) but the payoff is that we get the full set of examples visualized as LaTeX/PDF outputs.

Let me know if you'd like me to take an initial stab at this, or, if the interest is there for you to rig something up. But first let me know if you think this is an overall good strategy.

olivroy:
I think it could be a good idea. However, 1 pdf per example would have to be produced, which will probably take time to do. Maybe not generating them via Quarto would be best in this case, to avoid too many computations.

I agree something more thorough should be done. However, I am not sure I am willing to invest a lot of time on that.  I'd like to remain more focused on understanding where to make changes when bugs occur, (i.e. at init, build or render stage), improve the general documentation (#1675), and make sure no regressions are created (by adding tests and refactoring some).

But for LaTeX output, I am very impressed by Ken's recent improvements! Let's hope he continues doing integration testing :)

rich-iannone:
Oh yeah, I didn’t consider that the volume of examples could make PDF generation all very slow. Let’s keep adding to the system already in place. Thanks for getting that implemented btw!

Agreed that Ken does excellent work!

PR #1811, opened 2024-07-24
Code quality fixes

MichaelChirico:
I took a pass at the `lintr::lint_package(lintr::all_linters())` output on {gt} as a way to contribute back.

I fixed those which seem most meaningful and tried to exclude the more purely stylistic suggested changes.

I'm happy to roll back any changes, add tests, etc, as seen fit, and to help add a CI for keeping up code quality going forward using {lintr}.

### DIFF
--- a/R/utils_render_grid.R
+++ b/R/utils_render_grid.R
@@ -948,14 +948,14 @@ render_grid_svg <- function(label, style, margin) {
   width <- height <- NULL
 
   # Try if any height is declared in style attribute
-  if (any(grepl("^height:", svg_style))) {
+  if (any(startsWith(svg_style, "height:"))) {
     height <- gsub("^height:", "", svg_style[grep("^height:", svg_style)]) %>%
       parse_fontsize(style$text_gp$fontsize) %>%
       grid::unit(.grid_unit)
   }
 
   # Try if any width is declared in style attribute
-  if (any(grepl("^width:", svg_style))) {
+  if (any(startsWith(svg_style, "width:"))) {
     width <- gsub("^width:", "", svg_style[grep("^width:", svg_style)]) %>%
       parse_fontsize(style$text_gp$fontsize) %>%
       grid::unit(.grid_unit)
@@ -1381,7 +1381,7 @@ parse_css <- function(data) {
   # Find first and last line of definitions
   start <- grep("\\{$", css)
   end   <- which(css == "}")
-  if (!length(start) == length(end)) {
+  if (length(start) != length(end)) {
     cli::cli_abort("Formatting in {.fn compile_css} is unexpected.")
   }
 
@@ -1408,7 +1408,7 @@ parse_css <- function(data) {
   classes <- Map(`:`, start + 1, end - 1)
   names(classes) <- names
   classes <- lapply(classes, function(x) unlist(split[x], FALSE))
-  classes <- classes[grepl("^gt_", names(classes))]
+  classes <- grep("^gt_", names(classes), value = TRUE)
 
   # There are two entries for gt_table that we merge here
   is_table <- which(names(classes) == "gt_table")
--- a/R/utils_render_html.R
+++ b/R/utils_render_html.R
@@ -51,7 +51,7 @@ footnote_mark_to_html <- function(
     sup_class <- "gt_footnote_marks gt_asterisk"
   }
 
-  is_sup <- grepl("\\^", spec)
+  is_sup <- startsWith(spec, "^")
 
   if (grepl(".", spec, fixed = TRUE)) mark <- paste0(mark, ".")
   if (grepl("(", spec, fixed = TRUE)) mark <- paste0("(", mark)
@@ -2074,7 +2074,7 @@ build_row_styles <- function(
   # colnum values. Check and throw early.
   if (
     !isTRUE(all(styles_resolved_row$colnum %in% c(0, seq_len(n_cols)))) ||
-    any(duplicated(styles_resolved_row$colnum))
+    anyDuplicated(styles_resolved_row$colnum) > 0L
   ) {
     cli::cli_abort(
       "`build_row_styles()` was called with invalid `colnum` values."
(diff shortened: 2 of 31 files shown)


olivroy:
These all look good to me! @rich-iannone we'll let CI say the final word

MichaelChirico:
Not seeing what's causing the R CMD check issue. I'll take a closer look later.

R/utils_render_grid.R
```diff
@@ -948,14 +948,14 @@ render_grid_svg <- function(label, style, margin) {
...
-  if (any(grepl("^height:", svg_style))) {
+  if (any(startsWith(svg_style, "height:"))) {
     height <- gsub("^height:", "", svg_style[grep("^height:", svg_style)]) %>%
       parse_fontsize(style$text_gp$fontsize) %>%
       grid::unit(.grid_unit)
   }
 
   # Try if any width is declared in style attribute
-  if (any(grepl("^width:", svg_style))) {
+  if (any(startsWith(svg_style, "width:"))) {
```
olivroy, line 958:
```suggestion
  if (any(grepl("^width:", svg_style))) {
```

R/utils_render_grid.R
```diff
@@ -948,14 +948,14 @@ render_grid_svg <- function(label, style, margin) {
   width <- height <- NULL
 
   # Try if any height is declared in style attribute
-  if (any(grepl("^height:", svg_style))) {
+  if (any(startsWith(svg_style, "height:"))) {
```
olivroy, line 951:
```suggestion
  if (any(grepl("^height:", svg_style))) {
```

olivroy (commented):
Let's try reverting those as the failures are due to font size problems with grid. If it doesn't work, I can investgate a bit more

R/utils_render_grid.R
MichaelChirico: I'd expect an error from `startsWith()` returning `NA` (--> `if (NA)` --> error), is there a `tryCatch()` somewhere obscuring that?

What about `any(na.rm=TRUE)` instead?

(not worth more than a few seconds' effort, so let's just revert if it requires deeper thought/investigation...)

olivroy: You are right. It is better to investigate locally, by running the tests in test-as_gtable.R and tracking down the issue manually to see where a NA or length 0  is introduced.

MichaelChirico: Let's leave it as a future exercise :)

MichaelChirico:
(apologies for committing to my fork's `master` btw, I know it makes it more inconvenient as a reviewer...)

R/utils_render_grid.R
```diff
@@ -1408,7 +1408,7 @@ parse_css <- function(data) {
   classes <- Map(`:`, start + 1, end - 1)
   names(classes) <- names
   classes <- lapply(classes, function(x) unlist(split[x], FALSE))
-  classes <- classes[grepl("^gt_", names(classes))]
+  classes <- grep("^gt_", names(classes), value = TRUE)
```
olivroy, line 1411:
```suggestion
  classes <- classes[grepl("^gt_", names(classes))]
```

We want classes, not `names(classes)`. This was the main bug

R/utils_render_html.R
```diff
@@ -51,7 +51,7 @@ footnote_mark_to_html <- function(
     sup_class <- "gt_footnote_marks gt_asterisk"
   }
 
-  is_sup <- grepl("\\^", spec)
+  is_sup <- startsWith(spec, "^")
```
olivroy, line 54:
```suggestion
  is_sup <- grepl("^",  spec, fixed = TRUE)
```

R/utils_render_latex.R
```diff
@@ -57,7 +57,7 @@ footnote_mark_to_latex <- function(
   if (grepl("\\(|\\[", spec)) mark <- sprintf_unless_na("(%s", mark)
   if (grepl("\\)|\\]", spec)) mark <- sprintf_unless_na("%s)", mark)
 
-  if (grepl("\\^", spec)) {
+  if (startsWith(spec, "^")) {
```
olivroy, line 60:
```suggestion
  if (grepl("^", spec, fixed = TRUE)) {
```

R/utils_render_rtf.R
```diff
@@ -807,7 +807,7 @@ footnote_mark_to_rtf <- function(
     rtf_raw(
       paste0(
       "{",
-      if (grepl("\\^", spec)) "\\super " else NULL,
+      if (startsWith(spec, "^")) "\\super " else NULL,
```
olivroy, line 810:
```suggestion
      if (grepl("^", spec, fixed = TRUE)) "\\super " else NULL,
```

R/utils_render_xml.R
```diff
@@ -755,7 +755,7 @@ footnote_mark_to_xml <- function(
     xml_r(
       xml_rPr(
         xml_baseline_adj(
-          v_align = if (grepl("\\^", spec)) "superscript" else "baseline"
+          v_align = if (startsWith(spec, "^")) "superscript" else "baseline"
```
olivroy, line 758:
```suggestion
          v_align = if (grepl("^", spec, fixed = TRUE)) "superscript" else "baseline"
```
This appears incorrect? Aren't we checking for the ^ character here, would `grepl("^", spec, fixed = TRUE)` be more accurate?

olivroy (commented):
Found the bug by fetching the PR locally, it was in `classes`. Also, I think the linter had some false positive 

`grepl("\\^", x)` is not the same as `startsWith(x, "^")`. I suggest changing it to `grepl("^", spec, fixed = TRUE)`

R/utils_render_grid.R
olivroy:
```suggestion
  classes <- classes[startsWith(names(classes), "gt_")]
```
This may work too, but I haven't tested it.

MichaelChirico: Argh, I suspected that line from the beginning but I re-read it 4+ times without spotting the mistake!

R/utils_render_xml.R
MichaelChirico: Sorry for the sloppy fixes 😨

rich-iannone:
Thank you @MichaelChirico for this great contribution! Really appreciate it. And thanks @olivroy for your review. Once CI passes, I'll merge this one in.

rich-iannone (approved):
LGTM!

PR #1813, opened 2024-08-01
Add support for rlang anonymous function to fmt() - closes #1762

### LINKED ISSUE #1762: Improve error message when user attempts to call purrr-style function in fmt?
I should know better (#1148) but I just tried to use `fmt()` with purrr-style functions. While I know that they are becoming outdated, it might still be worth generating an error message that is a bit clearer than the below? Apart from maybe making it explicit that this syntax is not supported, it would be even more helpful to know where the error occurs (as this was part of a long pipeline).

``` r
library(gt)
mtcars |> gt() |> fmt(mpg, fns = ~.x+1)
#> Error in fmt$func[[eval_func]](data_tbl[[col]][rows]): attempt to apply non-function
```

<sup>Created on 2024-07-10 with [reprex v2.0.2](https://reprex.tidyverse.org)</sup>

LukasWallrich:
This adds support for anonymous functions in rlang style to `fmt()`, in line with the discussion (and helpful suggestions by @olivroy) in #1762. It also documents explicitly that anonymous functions (of both base R and rlang flavors) can be used in `fmt()`

# Checklist

- [not major imho] I have listed any major changes in the [NEWS](https://github.com/rstudio/gt/blob/master/NEWS.md).

Fixes: #1762


rich-iannone (approved):
LGTM!

PR #1820, opened 2024-08-08
Update actions and badges + remove partial match + tidyselect warning

olivroy:
The tidyselect warning was hard to catch, I used: (relates to https://github.com/rstudio/gt/pull/1763

tests/testthat/setup.R as 

```r
options(
  warnPartialMatchArgs = TRUE,
  warnPartialMatchDollar = TRUE,
  warnPartialMatchAttr = TRUE,
  lifecycle_verbosity = "error"
)
```

We are all clear now.

Also the pkgdown action now installs Quarto by default, so we can remove the hack!


rich-iannone (approved):
LGTM!

Issue #1827, opened 2024-08-14
`cells_stubhead()` doesn't style the full stub when more than 1 spanner 


olivroy:
Works as expected
```r
library(gt)

exibble |>
  gt::gt(rowname_col = "row") |>
  tab_spanner("spanners", c(char, num)) |>
  # tab_spanner("spanne2rs", c(char, num)) |>
  
  tab_stubhead("Stub row") |>
  tab_style(
    style = cell_fill("#f0f0f0"),
    cells_stubhead()
  ) 
```

![image](https://github.com/user-attachments/assets/0b1033c0-82c4-4ebc-bfcf-2bc1fbddc36f)

The top left cell isn't styled when there are 2 spanner levels

```r

two_levels <- exibble |>
  gt::gt(rowname_col = "row") |>
  tab_spanner("spanners", c(char, num)) |>
  tab_spanner("spanne2rs", c(char, num)) |>
  
  tab_stubhead("Stub row") |>
  tab_style(
    style = cell_fill("#f0f0f0"),
    cells_stubhead()
  )
two_levels
```
![image](https://github.com/user-attachments/assets/a4163b9d-d13c-440f-9d78-497ef254efd8)

Surprisingly, this works as expected in grid!
```r
plot(two_levels)
```
![image](https://github.com/user-attachments/assets/e889cf25-e1d4-4abb-b9d9-6798c1779852)

I am using the latest gt version with 0.11.0.9000

PR #1829, opened 2024-08-14
Support `tab_style()` + `cells_stubhead()` in `opt_interactive()` + add visual test

olivroy:
While looking into #1827, I noticed that `cells_stubhead()` had no effect in `opt_interactive()`. This PR adds this!


rich-iannone:
Thank you!

PR #1841, opened 2024-08-17
Correctly output `vec_fmt_markdown()` inside Quarto

### LINKED ISSUE #1840: `vec_fmt_markdown()` should probably not be affected by Quarto
WIth the html validations issues I am trying to understand, I wanted to look on the [examples Quarto website](https://quarto-gt-examples.netlify.app)

`vec_fmt_markdown()` should probably not have a different result inside Quarto? https://quarto-gt-examples.netlify.app/gt-vec_fmt_markdown

* https://github.com/rstudio/gt/issues/1839
* https://github.com/rstudio/gt/issues/1773
* https://github.com/quarto-dev/quarto-cli/issues/10534#issuecomment-2293892173
* https://github.com/quarto-dev/quarto-cli/discussions/10512#discussioncomment-10361691

olivroy:
Fix #1840 

Basically, the idea is that since `vec_fmt_markdown()` calls `fmt_markdown()` (and not the other way around, we can use some code to trick `vec_fmt_markdown()` to think it is not inside Quarto.

I tried using `on.exit(Sys.unsetenv("QUARTO_BIN_PATH")` but it didn't work for some reason. Would love to not use withr here.

I know tidyselect imports it, so gt has an indirect dep on it, but since withr 3.0 has simplified its code, it is possible they remove that dep at some point.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -47,10 +47,11 @@
 
 * Fixed an issue where `tab_spanner_delim()` would fail to resolve a duplicate id (@olivroy, #1821).
 
-
 * Fixed an issue with multiple `text_replace()` calls would produce bad result with `cells_column_labels()` (@olivroy, #1824).
 
 * `tidyselect::where()`, `tidyselect::all_of()`, `tidyselect::any_of()` are now re-exported by gt.
+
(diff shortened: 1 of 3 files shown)


rich-iannone (approved):
Thank you!

PR #1842, opened 2024-08-18
Modify html element ids to use name rather than labels for html validity

olivroy:
# Summary

Opening for discussion as this PR does *not* fix the aftermentioned issues, but may be a good first step. (I would need advice for anything extra)

It seems difficult to get rid of the `<style>` error. Suggestion to use `<link>` in https://github.com/quarto-dev/quarto-cli/discussions/10512#discussioncomment-10388292

It seems that when we are using `md()`, we are emitting invalid ids for spanners.

Unfortunately, it seems that the example supplied in #1839 still yields an invalid html error. Also wondering if there is a more robust `valid_html_id()` in the deps. I wrote a very simple version, https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id, but not sure it will work well. It still states this has to be unique?

````qmd
---
title: Hello
format: html
keep-md: true
lang: en
---


https://github.com/rstudio/gt/pull/988

```{r}
#| echo: false
#| message: false
#| eval: true
devtools::load_all(".")
 # or pak::pak("rstudio/gt#1842")
# library(gt)
mtcars[1:3, 1:4] |> 
  gt::gt() |> 
  gt::tab_spanner(label = gt::md("Header"), columns = 2:3)
```
````

It seems that with this PR, we are down to a single html validity error 

image

down from a couple from https://github.com/quarto-dev/quarto-cli/issues/10534#issuecomment-2293892173

But, it doesn't fix the issue, I still get this when I render the docc
image

@cscheid in case you have an idea of how we could deal with this, please let me know!

# Related GitHub Issues and PRs (the PR doesn't fix any issue)

* https://github.com/rstudio/gt/issues/1839
* #1840
* https://github.com/rstudio/gt/issues/1773
* https://github.com/quarto-dev/quarto-cli/issues/10534#issuecomment-2293892173
*https://github.com/quarto-dev/quarto-cli/discussions/10512#discussioncomment-10361691
* It seems that id was added in #988, and never touched after. To me, it seems more reasonable to use the unique id instead of labels for id
* [HTML: Valid id attribute values?](https://stackoverflow.com/questions/70579/html-valid-id-attribute-values) 

# Checklist

### CODE REVIEW

**Comment by capnrefsmmat** [comment]
> I wrote a very simple version, https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id, but not sure it will work well. It still states this has to be unique?

Yes, HTML ID attributes have to be globally unique in the document. If you print the same gt table twice in one HTML file, you can't have the same IDs in it.

I suspect that's not possible without adding some kind of random value to the IDs, e.g. making the attribute `id="cyl-dsklfj34"`, and keeping track of the IDs used in the table headers.

**Comment by rich-iannone** [comment]
Just getting back into the fold here, and I looked into what might be causing this. The malformed ID through `md()` was bad so thank you @olivroy for addressing that in the PR! There's probably more to do there to make it truly unique (e.g., using the table ID as part of the string).

You're right that this doesn't solve the issue. If you supply a non-malformed ID through the `id` arg (or just through using your fix here) you run into the display bug whenever `md()` is used in the `tab_spanner()` call.

This can be traced to the `<span>` tag of the spanner label. It's closed early in the Quarto-processed version. Here's a screenshot of the source HTML for the correct version:

correct_span

And here is the table within a Quarto HTML output:

incorrect_span

**Comment by olivroy** [comment]
Thanks for the clear information! Is there any way we could avoid that Quarto close that span?  Maybe we don't need to emit the gt_from_md class since we emit base64 encoding?

**Comment by rich-iannone** [comment]
> Thanks for the clear information! Is there any way we could avoid that Quarto close that span? Maybe we don't need to emit the gt_from_md class since we emit base64 encoding?

We could certainly try that! Would you be okay with me reviewing the PR as it is? I figure we could keep the scope smaller if we try to solve with a new PR.

**Comment by cscheid** [comment]
> I suspect that's not possible without adding some kind of random value to the IDs, e.g. making the attribute `id="cyl-dsklfj34"`, and keeping track of the IDs used in the table headers.

I would just like to suggest to avoid _random_ values. You could have a globally unique stateful counter instead, and that would make consecutive document renders consistent. (Random values would also make snapshot testing super hard...)

**Comment by olivroy** [comment]
Yes! I think this PR is fine in any case!

* Basically, aims to remove empty ids (they are not useful?
* Aim to take id from table variable name instead of label 
* Reduce number of html table problems

A future PR will be easier to do given #1844 is merged. You will be able to quickly visualize the diffs inside RStudio with `snapshot_review()`.

I'd love to see your solution for this at the end.

Agreed that random ids kind of defeat the original purpose. The idea was for accessibility when they were implemented in #988.  I think my PR stays with this purpose.

**Comment by cscheid** [comment]
> This can be traced to the `<span>` tag of the spanner label.

Could we have a look at the unparsed output that `gt` generates? Pasting the `keep-md: true` output from the test document in this PR would suffice. It could easily be a Quarto bug.

**Comment by olivroy** [comment]
I noticed in https://github.com/quarto-dev/quarto-cli/issues/3340#issuecomment-1698012543 that somehow Quarto processes this

> In my testing, if `<div class='gt_from_md'><p>...</p></div>` is replaced with `<span data-qmd="..."></span>`, then quarto (after this bugfix) actually renders this crossref correctly.

**Comment by cscheid** [comment]
This is the problem:

```
<span class="gt_column_spanner"><div data-qmd-base64="SGVhZGVy"><div class='gt_from_md'><p>Header</p>
</div></div></span>
```

A div is a block-level element, and can't exist inside a span.

image

cscheid:
(I'll reiterate our request that the HTML is valid, otherwise Quarto doesn't have a chance to do the right thing.)

olivroy:
Oh! Thanks 

So, if we replace this `div` code by `span`, it should work? Excited to try it after this PR is merged!

https://github.com/rstudio/gt/blob/02fef72e8bf0f6f9ad190de807eda13c84907592/R/utils.R#L760-L766

rich-iannone:
I think we should go the opposite way. Replace with outer `<span>` with a `<div>`.  Some testing will need to be done but changing to `htmltools::tags$div(...)` in the `level_1_spanners[[length(level_1_spanners) + 1]] <-`  assignments in `utils_render_html.R` (line 814) resulted the table example working properly.

olivroy:
@rich-iannone 
Nakes sense! Please merge this! with this PR + your proposed fix + the idea I came up with in , I think we may be able to actually fix this

Issue #1837, opened 2024-08-16
`cols_width()` not working when using `cols_merge()` for PDF output


snhansen:
## Description

Widths set with `cols_width()` isn't respected when using `cols_merge()` and when the output is PDF.

## Reproducible example

Consider this Quarto document:

	---
	format: pdf
	---

	```{r}
	#| echo: false
	library(gt)
	sp500 |>
	  dplyr::slice(50:55) |>
	  dplyr::select(-volume, -adj_close) |>
	  gt() |>
	  cols_align(
		columns = everything(),
		align = "right"
	  ) |>
	  cols_merge(
		columns = c(open, close),
		pattern = "{1}-{2}"
	  ) |>
	  cols_merge(
		columns = c(low, high),
		pattern = "{1}-{2}"
	  ) |>
	  cols_width(
		date ~ px(100),
		c("open", "low") ~ px(200)
	  )
	```

## Expected result

The table created doesn't respect the widths set:

![image](https://github.com/user-attachments/assets/f7f21472-da11-4d3b-a02f-d5a6977af730)

When output as html, everything looks fine:

![image](https://github.com/user-attachments/assets/882da52b-de4c-4cfd-aa57-1fbf87d31eb7)

## Session info

	> sessionInfo()
	R version 4.4.0 (2024-04-24 ucrt)
	Platform: x86_64-w64-mingw32/x64
	Running under: Windows 10 x64 (build 19045)

	Matrix products: default

	locale:
	[1] LC_COLLATE=Danish_Denmark.utf8     LC_CTYPE=Danish_Denmark.utf8       LC_MONETARY=Danish_Denmark.utf8    LC_NUMERIC=C                      
	[5] LC_TIME=English_United States.1252

	time zone: Europe/Copenhagen
	tzcode source: internal

	attached base packages:
	[1] stats     graphics  grDevices utils     datasets  methods   base     

	other attached packages:
	 [1] marginaleffects_0.20.1 patchwork_1.2.0        glue_1.7.0             parameters_0.21.7      ggpomological_0.1.2    epoxy_1.0.0           
	 [7] gt_0.11.0              lubridate_1.9.3        forcats_1.0.0          stringr_1.5.1          dplyr_1.1.4            purrr_1.0.2           
	[13] readr_2.1.5            tidyr_1.3.1            tibble_3.2.1           ggplot2_3.5.1          tidyverse_2.0.0       

	loaded via a namespace (and not attached):
[...]

PR #1588, opened 2024-02-25
Floating table

AaronGullickson:
# Summary

This PR addresses issue #157, specifically the ability to create latex floating environment tables with the use of the `table` + `tabular` environments rather than the `longtable` environment. Currently, I believe this is a major limitation of `gt` because tables are often broken up across pages, and most people expect tables to operate in a similar manner to figures. 

I added an option to `tab_option` called `latex.use.longtable` which defaults to FALSE. I then use a few `ifelse` clauses in `utils_render_latex.R` to get the proper output. Specification of the positioning of the float works well from the quarto code chunk option `tbl-pos`. 

If I run the following code:

```r
head(state.x77) |> 
  as.data.frame() |>
  gt() |>
  tab_options(latex.use.longtable = FALSE) |>
  as_latex()
```

Here is the latex output with the option set to FALSE:

```
"\\begin{table}\n\\begin{tabular}{rrrrrrrr}\n\\toprule\nPopulation & Income & Illiteracy & Life Exp & Murder & HS Grad & Frost & Area \\\\ \n\\midrule\\addlinespace[2.5pt]\n3615 & 3624 & 2.1 & 69.05 & 15.1 & 41.3 & 20 & 50708 \\\\ \n365 & 6315 & 1.5 & 69.31 & 11.3 & 66.7 & 152 & 566432 \\\\ \n2212 & 4530 & 1.8 & 70.55 & 7.8 & 58.1 & 15 & 113417 \\\\ \n2110 & 3378 & 1.9 & 70.66 & 10.1 & 39.9 & 65 & 51945 \\\\ \n21198 & 5114 & 1.1 & 71.71 & 10.3 & 62.6 & 20 & 156361 \\\\ \n2541 & 4884 & 0.7 & 72.06 & 6.8 & 63.9 & 166 & 103766 \\\\ \n\\bottomrule\n\\end{tabular}\n\\end{table}"
```

Here is the latex output with the option set to TRUE:

```
[...]

PR #1867, opened 2024-08-30
Fix table widths for latex tables and add unit testing

snhansen:
# Summary

In the old code, `col_alignment` contains the aligment of stub columns and visible columns, in that order. In the loop `for (i in seq_along(col_alignment))`, the `i`th row of `colwidth_df` is referenced. However, there are multiple problems with that:

* `colwidth_df` contains hidden columns, so the `i`th row of that data frame doesn't not necessarily correspond to the `i`th item of `col_alignment`.
* stub columns is not necessarily at the beginning of `colwidth_df`

In this PR, 

* we change the function `create_colwidth_df_l()` to include column alignments.
* we rewrite `create_table_start_l()` to extract only the visible columns of `colwidth_df` and we make sure the order of the rows in `colwidth_df_visible` is `"row_group", "stub", "default"`. The loop is then taken over the rows of `colwidth_df_visible`.
* we add some unit testing in `test-l_cols_width.R`. We test five different tables where we test whether `col_widths()` works correctly in different scenarios, e.g. when merging columns, when using stubs with/without row groups as a column.

# Screenshots

Below are some before and after screenshots of the five examples in `test-l_cols_width.R` with `latex.use_longtable` being TRUE and FALSE in that order. When not using longtable, I've set the `table.width` manually in the screenshots, otherwise the resulting table will be a full-width table from which the column widths can't really be judged.

## Table 1

image

## Table 2

image

## Table 3

image

## Table 4

image
image

## Table 5

image

# Additional comments:

* Stub columns are automatically left-aligned for LaTeX output. This is so for the code in this PR as well as the old code. 
* It might be an idea not to go full-width for tables using tabular if all column widths are specified (https://github.com/rstudio/gt/discussions/1857).
* Tables using tabular are left-aligned by default. Are we able to control this? (https://github.com/rstudio/gt/discussions/1848)

- https://github.com/rstudio/gt/issues/1837

# Checklist


snhansen:
@olivroy: Of course! Should be done now.

olivroy:
I didn't notice before, but it seems like your changes affect other tests https://github.com/rstudio/gt/actions/runs/10629794644/job/29475689750. You'd have to review and accept those if they make sense!

snhansen:
@olivroy: Thanks, the tests did indeed cause a mistake on my part. I think it should be good now though. Could you press the button to run tests again? 🤞

olivroy:
I unfortunately don't have that power, but we will review soon! from what I can see, it looks pretty good. The testing seems thorough and the explanations are clear. (All tests also pass locally for me)

> Stub columns are automatically left-aligned for LaTeX output. This is so for the code in this PR as well as the old code.


If you wish to change that, feel free to open a new PR!

> It might be an idea not to go full-width for tables using tabular if all column widths are specified (https://github.com/rstudio/gt/discussions/1857).

Agreed. Full widths tables were never a thing before in gt.  See this comment about that https://github.com/rstudio/gt/pull/1588#issuecomment-1965220731


> Tables using tabular are left-aligned by default. Are we able to control this? (https://github.com/rstudio/gt/discussions/1848)

It would be great if there was a Quarto div for this. https://github.com/quarto-dev/quarto-cli/discussions/8705#discussioncomment-8444576 is a solution.

olivroy (approved):

rich-iannone (approved):
Looks great, thanks!

PR #1873, opened 2024-09-03
avoid generating 0 width table

### LINKED ISSUE #1532: zero width table created
Quarto example  (without bootstrap)

````markdown
---
title: "Test"
format: 
  html:
    theme: pandoc
keep-md: true
---

## Toto
```{r}
library(magrittr)
library(gt)
```


## Tata
```{r}
iris %>%
  gt() %>%
  cols_width(
    "Sepal.Length" ~ px(300),
    everything() ~ px(50)
  ) %>%
  as_raw_html()
```

````

The HTML table created by **gt** has a `width: 0` in style (`<table class="gt_table" style="``)

Also `width=` attribute is set (which is deprecated I believe). 

Not sure from where this zero 0 width is coming. 

Reported first at https://github.com/quarto-dev/quarto-cli/issues/8233

### PR DESCRIPTION
**Pull request by olivroy**
# Summary

fix #1532. The reveal.js slide is no longer empty.

But cderv is right, since col widths are ignored by Quarto processing, maybe we should point towards using `quarto.disable.processing = TRUE`

### CODE REVIEW

**Comment by rich-iannone** [comment]
Sorry for taking so long to weigh in on this I feel I can finally shed some light on where `width: 0px` came from. Have a look at this document: 

https://github.com/rstudio/gt/blob/master/tests/gt-examples/01-html-script/html-17-table_width_specifications.R

That contains some expectations/specifications of how different width settings (both on columns and for the table) affect the HTML table. If you go to the comment/explanation in 

https://github.com/rstudio/gt/blob/33e9d62bc8fccc480c603fc85e27011b5501939d/tests/gt-examples/01-html-script/html-17-table_width_specifications.R#L249 

it's shown that specifying all columns with a px value (same behavior if all pct values set), we intervene and use `table-width: 0px`. And the reason for this is to ensure that the precise px values are honored. This was a desired feature (though, I admit, it's not well communicated in the docs) since exactly set column widths are quite tricky (the `max-width` CSS doesn't seem to work well in a table). Through some research, the best solution I could find was to set `table-width: 0px`, but only when either px values were set on all columns or pct values were set on all columns.

With the following code

```r
library(gt)

exibble |> 
  dplyr::select(num, char) |> 
  gt() |>
  cols_width(everything() ~ px(50))
```

We can see the effect between the PR and `master`

PR:
image

`master`:
image

Since this worked quite well in browsers, I felt this was a solved problem. That `reveal.js` doesn't accept this approach is a problem though. I wonder if this change could be made only if `reveal.js` output can somehow be detected. Failing that, only if rendering through Quarto. And perhaps there could be something in `tab_options()` to enable/disable the automatic `table-width: 0px` behavior (maybe calling it something like `table.strict_column_widths`)?

olivroy:
Thanks for the clear explanation! I will close this for now and we can think about this more. We can add comments in the code documenting this as well at some point!

olivroy:
From https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#try_it it seems that `table-layout:auto` could help mitigate that? 

You are right that there is no need to touch things that are working well in 95% of cases.

rich-iannone:
> From https://developer.mozilla.org/en-US/docs/Web/CSS/table-layout#try_it it seems that `table-layout:auto` could help mitigate that?
> 
> You are right that there is no need to touch things that are working well in 95% of cases.

I’ll look into that. It’s good to do a fresh survey and some experimentation. Last time was ~4 years ago. I doubt much has changed behavior-wise but I’m sure my research was incomplete. 

@cderv is there a way (or could there be a way) to detect whether a Quarto render is using reveal.js? I’m thinking it might be determined through an existing or new environment variable.

cderv:
> is there a way (or could there be a way) to detect whether a Quarto render is using reveal.js? I’m thinking it might be determined through an existing or new environment variable.

It should be possible to check the same way you are doing in **knitr** by looking at the `pandoc_to()` format.

rich-iannone:
Thank you Christoph! @olivroy I think we should revive this PR and use that knitr function. I know that knitr is in Suggests but it's reasonable to check for Quarto and assume knitr is available (Quarto requires knitr, though CRAN may want explicit checking for suggested packages anyway).

cderv:
FWIW we'll probably make this information available as environment variable. It is planned, just not yert done 
- https://github.com/quarto-dev/quarto-cli/issues/2216

You can wait for this, but should not be that complicated to use knitr if it is already in suggest. Even for CRAN. Just protect by checking package is available (which will always be the case on quarto context)

But you are already using knitr unconditionnaly in `knitr_is_word_output()` and others I believe.

Issue #1875, opened 2024-09-04
The `extract_body()` function extracts a df with stub and hidden columns 


rich-iannone:
The intent of `extract_body()` is to provide a dataframe of body cells with a chosen level of rendering applied to those cells (e.g., formatting, text transformations, etc.). However, it should match the body of rendered table and right now it:

- includes hidden columns
- has no option to exclude non-body cells (i.e., stub cells)

It can be useful to have stub information and the stub can be formatted. But it would be useful to omit the stub cells via an option.

PR #1889, opened 2024-09-25
Enhance the `extract_body()` function by adding inclusion options

rich-iannone:
This update of the `extract_body()` function includes two new arguments: `incl_hidden_cols` and `incl_stub_cols`. Previously, the function always returned a data frame that contained hidden columns. And, though stub columns were included, there was no way to exclude them from the returned data frame.

Fixes: https://github.com/rstudio/gt/issues/1875


olivroy:
I agree with the changes! Maybe it would be worth adding `...` + `rlang::check_dots_empty()` to put `output` as last argument and force naming the argument explicitly to minimize breaking changes in the future. 

Since this is technically a breaking change, it is worth adding to news.

```r
gt_tbl <- gt(exibble)
# before this PR, this would work.
extract_body(gt_tbl, "unfmt_included", "html")
```

R/extract.R
```diff
@@ -107,6 +124,8 @@
 extract_body <- function(
     data,
     build_stage = NULL,
+    incl_hidden_cols = FALSE,
+    incl_stub_cols = TRUE,
```
olivroy, line 128:
```suggestion
    incl_stub_cols = TRUE,
    ...,
```

R/extract.R
```diff
@@ -116,40 +135,131 @@ extract_body <- function(
   # Ensure that `output` is matched correctly to one option
   output <- rlang::arg_match(output)
```
olivroy, line 136:
```suggestion
  output <- rlang::arg_match(output)
  rlang::check_dots_empty()
```

R/extract.R
```diff
@@ -86,6 +86,23 @@
...
+#' 
+#' @param incl_stub_cols *Should stub columns be included?*
+#' 
+#'   `scalar<logical>` // *default:* `TRUE`
+#' 
+#'   Any stub columns in the **gt** object (which may consist of a grouping
+#'   column and a column for row labels) are included in the extracted data for
+#'   clarity but clearly marked with the names `"::group_id::"` and
+#'   `"::rowname::"`. We can exclude them by setting `incl_stub_cols = FALSE`.
 #'
```
olivroy, line 106:
```suggestion
#'
#'  @inheritParams rlang::args_dots_empty
#'
```

olivroy (approved):

rich-iannone:
Thanks for the review! I'll add a note to NEWS describing the breaking changes here. Also, before finishing this up, I'll add 1-2 examples since there aren't any right now.

PR #1893, opened 2024-09-26
Performance of vec_fmt_*() by avoiding checking of compatibility + coercion to tibble

olivroy:
# Summary

Hopefully, my last one. 

My example is down to about 4 seconds rendering time.

The main idea is to avoid dplyr or coercion to tibble and also avoid to check for the compatibility of the formatter unnecessarily. 

I realized `cols_add()` has some issues, so I fixed up the two corner cases `.before = 0` and `.after = last_col()` + added tests.

Also eliminated some tidyselect deprecation warnings

- Ref: #1888 , #1891

### DIFF
--- a/R/utils.R
+++ b/R/utils.R
@@ -1866,9 +1866,10 @@ num_suffix <- function(
     }
 
     return(
-      dplyr::tibble(
-        scale_by = rep_len(scale_by, length(x)),
-        suffix = rep_len("", length(x))
+      vctrs::data_frame(
+        scale_by = scale_by,
+        suffix = "",
+        .size = length(x)
       )
     )
   }
@@ -1919,7 +1920,7 @@ num_suffix <- function(
 
   # Create and return a tibble with `scale_by`
   # and `suffix` values
-  dplyr::tibble(
+  vctrs::data_frame(
     scale_by = 1 / base^suffix_index,
     suffix = suffix_labels
   )
@@ -1947,9 +1948,10 @@ num_suffix_ind <- function(
     }
 
     return(
-      dplyr::tibble(
-        scale_by = rep_len(scale_by, length(x)),
-        suffix = rep_len("", length(x))
+      vctrs::data_frame(
+        scale_by = scale_by,
+        suffix = "",
+        .size = length(x)
       )
     )
   }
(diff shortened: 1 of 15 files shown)


tests/testthat/_snaps/cols_add.md
```diff
@@ -0,0 +1,25 @@
...
+
+    Code
+      cols_add(tbl, x = 1, .after = 2, .before = 3)
+    Condition
+      Error in `cols_add()`:
+      ! Expressions cannot be given to both `.before` and `.after`.
+    Code
+      cols_add(tbl, x = 1, .after = 15)
+    Condition
+      Error in `vars_select_eval()`:
```
olivroy, line 11: Will be fixed by https://github.com/r-lib/tidyselect/pull/352 when merged

R/utils_render_common.R
```diff
@@ -108,13 +108,17 @@ render_formats <- function(data, context) {
...
-          compat = compat
+        skip_compat_check ||
+        (
+          col %in% colnames(data_tbl) &&
+          is_compatible_formatter(
+            table = data_tbl,
+            column = col,
+            rows = rows,
+            compat = compat
+          )
```
olivroy, line 121: I added this to skip this check for vec_fmt_*

R/utils.R
```diff
@@ -226,9 +226,9 @@ tf_formats <- function() {
 
   dplyr::tribble(
```
olivroy, line 227: Before, it seemed that `tf_formats_icons()` included NAs, not sure what the impact of all this is tbh.

Now, it is:

```
tf_formats_icons()
[1] "c(\"✔\", \"✘\")" "c(\"●\", \"⭘\")" "c(\"■\", \"□\")" "c(\"◆\", \"◇\")" "c(\"↑\", \"↓\")" "c(\"▲\", \"▼\")"
[7] "c(\"▶\", \"◀\")"
```

which seems to make sense?

rich-iannone: Thanks for adding `omit_na()`! Yeah, this return value is correct now but AFAICT nothing in the package uses `tf_formats_icons()` (so I guess no impact).

rich-iannone (approved):
LGTM!

PR #1896, opened 2024-09-30
Improve performance by having boxhead as a data frame.

olivroy:
# Summary

It proved more challenging than I thought initially, because of the way R nests objects.

I added a workaround to unnest such objects to improve performance 

See how the boxhead initalialization takes a full 2 seconds

image

This PR:
image

I have added a few (simpler) tests to make sure this doesn't break (or is easier to debug)

The only remaining thing would be to find a way to avoid resolving columns as this alone takes up a full second of run time. I will think of this.

# Checklist

- [ ] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html).


olivroy:
The snapshot failure was due to the new katex package I didn't yet have installed. To prevent this issue from happening in the future, I am stripping the katex version from the snapshot. I verified manually that only the katex version changed in the snapshot! The rest is the same

rich-iannone (approved):
LGTM!

PR #1927, opened 2024-11-27
Use more human readable snapshots with tests

olivroy:
Will be much easier to inspect the diff now.

The only change is to make `render_as_html()` use `cat()`

My goal is to inspect the tests that are long and see if we could make the tests shorter.

Also, if we change what html output gt is emiting, it will be much easier to see what changed

I am seeing that there are always two new lines at the end of snapshots. maybe we could remove them.


rich-iannone:
This is much better, thanks! Those extra line breaks before the closing `</table>` tag are really more visible now (and it seems like they are present if there aren't any `<tfoot>` tags).

PR #1928, opened 2024-11-29
Avoid tidyselect deprecation warning

olivroy:
For https://github.com/rstudio/pointblank/issues/441

### DIFF
--- a/R/gt_preview.R
+++ b/R/gt_preview.R
@@ -191,7 +191,7 @@ gt_preview <- function(
       tab_style(
         gt_tbl,
         style = cell_fill(color = "#E4E4E4"),
-        locations = cells_body(columns = visible_vars, rows = ellipsis_row)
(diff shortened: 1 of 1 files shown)


olivroy:
Also, maybe revert #1927 ? like `gert::git_reset_hard("HEAD~4");gert::git_push(force = TRUE)`, not sure it is worth it.. It increased the package size by ~30MB...

rich-iannone:
Sorry to leave this for so long. Will review, and with #1927 it's probably fine since the size of the source tarball is still 3.3 MB and we .Rbuildignore a lot. Maybe in the near term, now that we can more easily see the tables, we can cut down on unnecessary rows in a lot of snapshots.

olivroy:
Ok, makes sense! Thanks

rich-iannone (approved):
LGTM!

PR #1930, opened 2024-12-02
Autocoloroptions

xx02al:
I have added light and dark `autocolor_text` color options. Often, tables are embedded in a context and in there "black" is not black, but another very dark color and analogously for white. Being able to design the tables in harmony is central.

The current implementation follows the previous design of {gt}, particularly `data_color()`, but/thus it is unattractive in that if `autocolor = FALSE`, now allready 3 arguments are irrelevant.

- [ x] I understand and agree to the [Code of Conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html).

I did not add tests (as far as now), also because I first want to be sure that the maintaineres support this feature.

Additionally, I fixed a constant in the calculation of relative luminance WCAG (according to [documentation](https://www.w3.org/TR/WCAG21/#dfn-relative-luminance)).

### DIFF
--- a/R/data_color.R
+++ b/R/data_color.R
@@ -57,7 +57,8 @@
 #' text
 #' - text autocoloring: if colorizing the cell background, `data_color()` will
 #' automatically recolor the foreground text to provide the best contrast (can
-#' be deactivated with `autocolor_text = FALSE`)
+#' be deactivated with `autocolor_text = FALSE`; a light and dark color to be
+#' used can be specified with `autocolor_light` and `autocolor_dark`)
 #'
 #' `data_color()` won't fail with the default options used, but
 #' that won't typically provide you the type of colorization you really need.
@@ -233,6 +234,22 @@
 #'   default this is `"apca"` (Accessible Perceptual Contrast Algorithm) and the
 #'   alternative to this is `"wcag"` (Web Content Accessibility Guidelines).
 #'
+#' @param autocolor_light
+#'
+#'   `scalar<character>` // *default:* `"white"`
+#'
+#'   The light color to use when `autocolor_text = TRUE`. By default the color
+#'   `"white"` will be used (`#FFFFFF"`). Alpha channel values will be set to
+#'   1.0 (fully opaque).
+#'
+#' @param autocolor_dark
+#'
+#'   `scalar<character>` // *default:* `"black"`
+#'
+#'   The dark color to use when `autocolor_text = TRUE`. By default the color
+#'   `"black"` will be used (`#000000"`). Alpha channel values will be set to
+#'   1.0 (fully opaque).
+#'
 #' @param colors *[Deprecated] Color mapping function*
 #'
 #'   `function` // *default:* `NULL` (`optional`)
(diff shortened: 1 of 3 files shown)


xx02al:
Hi @rich-iannone and @olivroy since it has been a while since I opened this pull request, and I would still like to be able to use this feature in various situations (i.e., tables I create), I would like to ask what you think about it!?

R/data_color.R
```diff
@@ -1282,6 +1307,9 @@ ideal_fgnd_color <- function(
   # Get the correct `algo` value
   algo <- rlang::arg_match0(algo, values = c("apca", "wcag"))
 
+  light_color <- farver::encode_colour(farver::decode_colour(light))
+  dark_color <- farver::encode_colour(farver::decode_colour(dark))
```
olivroy, line 1311: Why is this necessary?
We don't use this pattern of encoding / decoding colors in the package..

R/utils_color_contrast.R
```diff
@@ -135,7 +135,7 @@ get_relative_luminance_wcag <- function(col) {
 
   coef <- round(c(apca_coeffs$sRco, apca_coeffs$sGco, apca_coeffs$sBco), 4)
 
-  rgb[] <- ifelse(rgb <= 0.03928, rgb / 12.92, ((rgb + 0.055) / 1.055)^2.4)
+  rgb[] <- ifelse(rgb <= 0.04045, rgb / 12.92, ((rgb + 0.055) / 1.055)^2.4)
```
olivroy, line 138: This is great! Great research.

olivroy (commented):
Overall, good start! Sorry for the delay.

I think this is a good idea. Can't garantee that we merge it. I'd suggest adding some before / after (visual) examples of the improvement this PR brings.

We'd like to avoid new imports, so if you find a way to accomplish this without using farver, it would be great.

It is good that you added some tests.

Also, feel free to use gt from your PR `pak::pak("rstudio/gt#1930") and report back if you see that something that is off.

You would need to add a news bullet as well that briefly describe the change https://style.tidyverse.org/news.html

R/data_color.R
```diff
@@ -233,6 +234,22 @@
 #'   default this is `"apca"` (Accessible Perceptual Contrast Algorithm) and the
 #'   alternative to this is `"wcag"` (Web Content Accessibility Guidelines).
 #'
+#' @param autocolor_light
```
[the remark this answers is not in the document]
xx02al: done

R/data_color.R
```diff
@@ -233,6 +234,22 @@
...
+#'
+#'   `scalar<character>` // *default:* `"white"`
+#'
+#'   The light color to use when `autocolor_text = TRUE`. By default the color
+#'   `"white"` will be used (`#FFFFFF"`). Alpha channel values will be set to
+#'   1.0 (fully opaque).
+#'
+#' @param autocolor_dark
```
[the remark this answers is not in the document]
xx02al: done

R/data_color.R
xx02al: See [just above](https://github.com/rstudio/gt/pull/1930#discussion_r1969914825).

xx02al:
Here's an example, not because it is particularly aesthetically pleasing, but I think it serves the purpose of clearly pointing out differences. Run it with and without `autocolor_light = "#FFE8D1"`.

```
exibble |>
  gt() |> 
  data_color(
    columns = c(fctr, datetime, group),
    method = "factor",
    rows = c(2, 4:6, 8),
    palette = "#725F0B",
    na_color = "#B6D7C7",
    autocolor_light = "#FFE8D1"
  ) |> 
  data_color(
    columns = c(date, currency),
    palette = "#6E6684",
    na_color = "#B6D7C7",
    autocolor_light = "#FFE8D1"
  ) |> 
  tab_options(
    table.background.color = "#004249",
    table.font.color.light = "#FFE8D1"
  )
```

rich-iannone (approved):
LGTM!

PR #1952, opened 2025-02-20
Use "t" as default float position instead of "!t"

### LINKED ISSUE #1935: Default latex table position is poor choice in R Markdown
## Prework

* [ x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [ x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

In R Markdown, if the `latex_tbl_pos` option is not modified, the table position defaults to `"!t"`.  This is not a good default, as it forces the table to the top of the page even when that doesn't make sense.  A better default would be `"t"`, which usually puts the table at the top of the page, but might push it to the top of the next page in some circumstances, e.g. on the first page of an article where the article title should be at the top of the first page.

## Reproducible example

Run this in R Markdown:

````
---
title: "Mtcars example"
output: pdf_document
---

```{r setup, include=FALSE}
knitr::opts_chunk$set(echo = TRUE)
library(gt)
```

## Example

Here is some text before the table.

```{r table}
head(mtcars) |> gt() 
sessionInfo()
```
````
This puts the table above the article title:
Image

## Expected result

The table should appear on the next page.  This is what happens If `|> tab_options(latex.tbl.pos = "t")` is added to the line generating the table, because LaTeX knows not to put a table before the title.  That would be a better default than the current one.

olivroy:
# Summary

Finishing up the PR started by @AaronGullickson 

- Ref: #1940

Closes #1935, Closes #1940

# Checklist


rich-iannone (approved):
LGTM, thanks!

Issue #1773, opened 2024-07-11
Footnote marker placement when using `gt::md()` in Quarto in gt v0.11.0


ddsjoberg:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

When a footnote is placed next to text that has been processed with `gt::md()`, the footnote marker is placed in a separate row when rendered with Quarto. I spotted this behaviour in both column labels and footnotes (I haven't checked all location types, e.g. table cells, grouped row headers, summary rows, etc FYI). Note this does not occur when running the code interactively in the RStudio IDE. I think this is new behavior in gt 0.11.0.

## Reproducible example

In the example below, I am using `gt::md()` in the column label and in the actual footnote text. In both cases, the use of `gt::md()` drives the footnote marker so its own row.

````
---
title: "Untitled"
format:
  html:
    embed-resources: true
---

## Quarto

gt footnotes without `gt::md()`: no issues.

```{r}
mtcars[1:2, 1:2] |> 
  gt::gt() |> 
  gt::cols_label(mpg = "**MPG**") |> 
  gt::tab_footnote(
    "_Adding footnote_",
    locations = gt::cells_column_labels(columns = gt::everything())
  )
```

gt footnotes **WITH** `gt::md()`: footnote markers appear on their own line.

```{r}
mtcars[1:2, 1:2] |> 
  gt::gt() |> 
  gt::cols_label(mpg = gt::md("**MPG**")) |> 
  gt::tab_footnote(
    gt::md("_Adding footnote_"),
    locations = gt::cells_column_labels(columns = gt::everything())
  )
```
````

<img width="1215" alt="image" src="https://github.com/user-attachments/assets/f9c230b5-3b45-4e24-a822-28fcc6e122e0">

Thank you again!!! 🕺🏼 

PS This issue could be related to https://github.com/rstudio/gt/issues/1541. But that doesn't mention Quarto?

olivroy:
Thanks for the report!

[...]

Issue #1605, opened 2024-03-11
[quarto] tables with `fmt_markdown` don't render in quarto dashboard


jburos:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

I have a quarto dashboard, converted from an html file, in which some but not all gt tables fail to render. I was able to isolate the behavior only to those tables using `fmt_markdown`. In all my exmaples, the markdown string includes a `<br/>`. I'm not sure if this extends to other contents.

## Reproducible example

* [x] Post a [minimal reproducible example](https://www.tidyverse.org/help/) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
    * [x] **Runnable**: post enough R code and data so any onlooker can create the error on their own computer.
    * [x] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
    * [x] **Readable**: format your code according to the [tidyverse style guide](https://style.tidyverse.org/).

````
---
title: "test-gt-markdown"
format: dashboard
---

# with markdown

## row

```{r}
#| title: with markdown
library(gt)
library(dplyr)
library(glue)
h <- head(sp500, n = 3) |> 
  mutate(test = glue::glue('{LETTERS[row_number()]}<br/>test')) |>
  gt() |>
  fmt_markdown(columns = c(test))
h
```

# session

```{r}
sessioninfo::session_info()
```
````

<img width="1183" alt="Screenshot 2024-03-11 at 3 52 25 PM" src="https://github.com/rstudio/gt/assets/923453/ad469a84-8b68-4e74-ba3d-fb17169673f9">

## Expected result

I expected the table to render within the card. Instead, the rendered document is blank.

## Session info

I'm using quarto version: 1.4.551

```r
sessioninfo::session_info()

─ Session info ───────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.3.2 (2023-10-31)
 os       macOS Sonoma 14.3
 system   aarch64, darwin20
 ui       X11
 language (EN)
 collate  en_US.UTF-8
 ctype    en_US.UTF-8
 tz       America/New_York
 date     2024-03-11
[...]

PR #1958, opened 2025-02-21
Fix issue with markdown and Quarto

### LINKED ISSUE #1957: When character column starts with string suggesting list, and gt::fmt_markdown() enabled, Quarto renders contents twice
## Prework

* [ ] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [ ] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Describe the bug clearly and concisely. 

## Reproducible example

* [ ] Post a [minimal reproducible example](https://www.tidyverse.org/help/) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
    * [ ] **Runnable**: post enough R code and data so any onlooker can create the error on their own computer.
    * [ ] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
    * [ ] **Readable**: format your code according to the [tidyverse style guide](https://style.tidyverse.org/).

## Expected result

What should have happened? Please be as specific as possible.

## Session info

End the reproducible example with a call to `sessionInfo()` in the same session (e.g. `reprex(session_info = TRUE)`) and include the output.

olivroy:
# Summary

Basically this reverts [part](https://github.com/rstudio/gt/pull/1860#discussion_r1966003385) of #1860.

As suggested in https://github.com/quarto-dev/quarto-cli/issues/11932#issuecomment-2609871584, we encode to base64 the unprocessed text.

It seems to do the trick and does not seem to introduce regressions (in particular the example in #1773 renders correctly)

For the review, it would be best to investigate this 60f19adec234093eac4fe39780bb363fec2aec85 instead of the full diff because I added tests in the first commit.

@cderv , do you think this is the right approach? 

Fixes #1957 

Fixes https://github.com/quarto-dev/quarto-cli/issues/11932

Fixes https://github.com/quarto-dev/quarto-cli/issues/11610

# investigation

Here is an issue I found after this change in typst:

* Not getting anything with html-table-processing: none. So, far I haven't seen issues on html output

````qmd
---
title: test
format: typst
execute:
  echo: false
---

```{r}
devtools::load_all()
```

```{r}
#| html-table-processing: none
dplyr::tibble(
  char = LETTERS[1:2],
  num = c(1.2, -33.52),
  text = c("- x", "- y"),
  text2 = "[gog](https://google.com)"
) |>
  gt() |>
  fmt_markdown() %>% 
  gt::tab_source_note(md("[google](https://google.com)"))

```
````

With the following qmd

````qmd

---
title: test
format: html
execute:
  echo: false
---

```{r}
devtools::load_all(
)

```

```{r}
dplyr::tibble(
  char = LETTERS[1:2],
  num = c(1.2, -33.52),
  text = c("- x", "- y"),
  text2 = "[gog](https://google.com)"
) |>
  gt() |>
  fmt_markdown() %>% 
  gt::tab_source_note(md("[google](https://google.com)"))

```

```{r}
exibble %>%
    dplyr::select(num, char, fctr) %>%
    dplyr::mutate(x = "- 1") %>%
    dplyr::slice_head(n  = 5) %>%
    gt() %>%
    fmt_markdown(num) %>%
    tab_footnote(
      md("Problem because num row 1 is `fmt_markdown()` + also the footnote is wrapped in md."),
      locations = cells_body("num", 1)
    ) %>%
    tab_footnote(
      "A problem because fctr is labelled with md",
      locations = cells_column_labels("fctr")
    ) %>%
    tab_footnote(
      "Not a problem",
      locations = cells_column_labels("char")
    ) %>%
    cols_label(fctr = md("Factor")) %>%
    tab_header(md("[gog](https://google.com)")) %>%
    tab_spanner(md("problem"), c(2, 3))
```

````

gt 0.11.1:

Image

This PR:

Image

# Checklist

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,7 +1,6 @@
 # gt (development version)
 
-* Fixed an issue where `fmt_markdown()` could create strange output in Quarto (html and Typst formats) (@olivroy, #1957, [quarto-dev/quarto-cli#11932](https://github.com/quarto-dev/quarto-cli/issues/11932)).
-  This means that using  `#| html-table-processing: none` may no longer work as expected in some cases.
+* Fixed an issue where `fmt_markdown()` could create strange output in Quarto (html and Typst formats) (@olivroy, #1957, [quarto-dev/quarto-cli#11932](https://github.com/quarto-dev/quarto-cli/issues/11932), [quarto-dev/quarto-cli#11610](https://github.com/quarto-dev/quarto-cli/issues/11610)).
  
 * The default table position in LaTeX is now "t" instead of "!t" (@AaronGullickson, #1935).
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 # gt (development version)
 
+* Fixed an issue where `fmt_markdown()` could create strange output in Quarto (html and Typst formats) (@olivroy, #1957, [quarto-dev/quarto-cli#11932](https://github.com/quarto-dev/quarto-cli/issues/11932)).
+  This means that using  `#| html-table-processing: none` may no longer work as expected in some cases.
+ 
 * The default table position in LaTeX is now "t" instead of "!t" (@AaronGullickson, #1935).
 
 * Fixed an issue where cross-references would fail in bookdown::html_document2 (@olivroy, #1948)
--- a/R/quarto.R
+++ b/R/quarto.R
@@ -42,12 +42,15 @@ process_md_quarto <- function(text, md_engine_fn) {
 
   # Remove paragraph
   non_na_text_processed <- gsub("^<p>|</p>\n$", "", non_na_text_processed)
-  # Use base64 encoding to avoid issues with escaping internal double
+
+  # Use base64 encoding on the unprocessed text
+  # https://github.com/quarto-dev/quarto-cli/issues/11932#issuecomment-2609871584
+  # to avoid issues with escaping internal double
   # quotes; used in conjunction with the 'data-qmd-base64' attribute
   # that is recognized by Quarto
   non_na_text <-
     vapply(
-      non_na_text_processed,
+      non_na_text,
       FUN.VALUE = character(1L),
       USE.NAMES = FALSE,
       FUN = function(text) {
--- a/tests/testthat/_snaps/quarto.md
+++ b/tests/testthat/_snaps/quarto.md
@@ -6,7 +6,7 @@
       <table class="gt_table" data-quarto-disable-processing="false" data-quarto-bootstrap="false">
         <thead>
           <tr class="gt_heading">
-            <td colspan="4" class="gt_heading gt_title gt_font_normal gt_bottom_border" style><span data-qmd-base64="PGEgaHJlZj0iaHR0cHM6Ly9nb29nbGUuY29tIj5nb2c8L2E+"><span class='gt_from_md'><a href="https://google.com">gog</a></span></span></td>
+            <td colspan="4" class="gt_heading gt_title gt_font_normal gt_bottom_border" style><span data-qmd-base64="W2dvZ10oaHR0cHM6Ly9nb29nbGUuY29tKQ=="><span class='gt_from_md'><a href="https://google.com">gog</a></span></span></td>
           </tr>
           
           <tr class="gt_col_headings gt_spanner_row">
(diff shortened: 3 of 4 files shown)

### CODE REVIEW

**Comment by olivroy** [comment]
This seems to be correct. I made an unnecessary change in #1860 Half of https://github.com/rstudio/gt/pull/1860/commits/b0a8c76a24c9c8be6092073186f6852e825ce6e7 is wrong but this seems correct now without recreating #1773

**Review by rich-iannone — APPROVED** [review]
LGTM!

**Comment by rich-iannone** [comment]
@olivroy this looks really good to me! We'll hold off on merging until @cderv has a chance to look this over.

File: NEWS.md [hunk 1]
```diff
@@ -1,5 +1,8 @@
 # gt (development version)
 
+* Fixed an issue where `fmt_markdown()` could create strange output in Quarto (html and Typst formats) (@olivroy, #1957, [quarto-dev/quarto-cli#11932](https://github.com/quarto-dev/quarto-cli/issues/11932)).
+  This means that using  `#| html-table-processing: none` may no longer work as expected in some cases.
```
cderv, line 4: This is puzzling to me. I would like to understand why. 

`#| html-table-processing: none` is just a way to tell Quarto to not process the HTML table. It should not have impact on the output. 

If this is a matter of **gt** that needs to know about it, then we can make sure this is communicated correctly to engine. But this is not possible on all cases considering different ways to set it. https://quarto.org/docs/authoring/tables.html#disabling-quarto-table-processing

Anyway, I don't think this should have an impact, but I can see that the HTML will be different in Quarto context. 

**gt** also has an option that is supposed to pass this information to quarto 
https://gt.rstudio.com/reference/tab_options.html?q=tab_options#arg-quarto-use-bootstrap-quarto-disable-processing

It would be good to be sure this still works as expected. And maybe this should be prioritize way to set it for **gt** 🤔 

However, if you are reffering to using `html-table-processing: none` on **gt** table to render to `format: typst`, then no output is expected because

* **gt** produces HTML raw content and not typst raw content
* `html-table-processing: none` means explicit opt-out of Quarto parsing HTML table to render to any format table using Pandoc writers for `Table` element. 

If this why you put the comment in news, then I think it can be removed. This is all working as expected

cderv (commented):
I see in #1860 that `<span>` are used instead of `<div>` ([`b0a8c76` (#1860)](https://github.com/rstudio/gt/pull/1860/commits/b0a8c76a24c9c8be6092073186f6852e825ce6e7)). I suggested this idea (https://github.com/rstudio/gt/issues/1605#issuecomment-2117913671) but I remember a later discussion with @rich-iannone about this being a bad idea because of the content that could be inside the cell of **gt** that could sometime not compatible with span. 🤔 I don't have example in mind though... 

If you think this is all good, then let's merge once the news item is tweaked.

I'll track quarto-cli issue related to this **gt** change to see if this fix others. 

thanks !

cderv:
> I'll track quarto-cli issue related to this gt change to see if this fix others.

This PR will also solve this issue

* https://github.com/quarto-dev/quarto-cli/issues/11610
	Using markdown for newline works with typst format
	````markdown
	---
	title: "Untitled"
	format: typst
	keep-md: true
	---
	
	```{r}
	#| echo: false
	#| message: false
	library(tidyverse)
	# devtools::load_all()
	library(gt)
	
	head(mtcars) %>% 
	  select(1:4) %>% 
	  gt() %>% 
	  tab_source_note(md("*Bla*. hello i'm a footnote thats  
	too long for this table."))
	```
	````
	image

NEWS.md
cderv: We document HTML table parsing in there: https://quarto.org/docs/authoring/tables.html#html-tables

> As a result, you can use HTML table syntax in your documents and it will be converted to Markdown syntax for all formats.

And then we document disabling: https://quarto.org/docs/authoring/tables.html#disabling-quarto-table-processing

> When you disable Quarto’s HTML table processing, tables are not translated to Markdown, will not be rendered to other output formats

How would you make this clearer ? or How would you present it differentlly ?

cderv (approved):

NEWS.md
olivroy: I guess it is implied.  I just guess it would be good to say something specific about `format: typst` in the quarto Disabling Quarto Table Processing. 

Maybe 
````qmd
---
format:
  typst:
    html-table-processing: none
---
````

For the typst format, you will not see tables if you use this? 

But I am okay either way.

olivroy:
Thanks for the explanations!

One last thing: if we revert back to using `div` recreates #1773. So, until we have a clear use case of going back to `div`, I think we should stick to `span` for now.

rich-iannone (approved):

rich-iannone:
I think we ought to keep using `span` because you did solve that important issue in #1773 and we can always explore using a better `div` implementation (if it exists) in the future.

rich-iannone:
@olivroy thanks for doing this important work! I will merge the PR now.

rich-iannone:
@cderv thanks for your review of this! https://github.com/rstudio/gt/pull/1958 is merged now so any downstream issues that are fixed by this could probably be closed. I also intend to submit gt to CRAN very soon.

PR #1964, opened 2025-03-05
Merge pattern regex adjustment

### LINKED ISSUE #1880: When `pattern` manages NAs in `fmt_icon() |> cols_merge()`, `NA` gets repeatedly appended to the value.
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

After passing a table to `fmt_icon()`, if I then `cols_merge()` with a pattern that takes into account NAs, a very large number of `NA` get appended to cells that *don't* have NAs.

## Reproducible example
``` r
library(gt)

data.frame(
  x = c("dice-one", "dice-two", "dice-three"),
  y = c("dice-one", NA, "dice-three")
) |> 
  gt() ->
  base_table

```
### This works
Without using `fmt_icon()`, this works as expected.

```r
base_table |> 
  cols_merge(
    columns = everything(),
    pattern = "({1}<< {2}>>)"
  )
```
blog_-_drafting_-_RStudio

### This does not work
It also takes a *while* to run, I assume because of all the NAs that are getting appended.

```r
base_table |> 
  fmt_icon() |> 
  cols_merge(
    columns = everything(),
    pattern = "({1}<< {2}>>)"
  )
```
blog_-_drafting_-_RStudio

Inspecting the html output for the first row, there were 49,547 `NA`s appended to the cell.

## Expected result

I expected output similar to the working example, but with the icons instead of the names of the icons.

## Session info

```
R version 4.4.0 (2024-04-24)
Platform: aarch64-apple-darwin20
Running under: macOS Sonoma 14.2

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.4-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.0

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

time zone: America/New_York
tzcode source: internal

attached base packages:
[1] stats     graphics  grDevices datasets  utils     methods   base     

other attached packages:
[1] gt_0.11.0

loaded via a namespace (and not attached):
 [1] digest_0.6.37     utf8_1.2.4        R6_2.5.1          fastmap_1.2.0    
 [5] tidyselect_1.2.1  xfun_0.47         magrittr_2.0.3    glue_1.7.0       
 [9] tibble_3.2.1      knitr_1.48        htmltools_0.5.8.1 pkgconfig_2.0.3  
[13] dplyr_1.1.4       generics_0.1.3    lifecycle_1.0.4   xml2_1.3.6       
[17] cli_3.6.3         fansi_1.0.6       vctrs_0.6.5       withr_3.0.1      
[21] renv_1.0.7        compiler_4.4.0    rstudioapi_0.16.0 tools_4.4.0      
[25] pillar_1.9.0      rlang_1.1.4    
```

olivroy:
# Summary

2 tests are failing which I need to investigate, but opening in case you know how to fix.

Basically, the regexp was wrong in this case which caused an infinite loop. So, I added a condition which makes the case where no match occur error early. Just need to figure out why the test is failing and also how I could simplify the test.

Also needs news.
 
- Ref: fixes #1880

# Checklist

### DIFF
--- a/R/utils_render_common.R
+++ b/R/utils_render_common.R
@@ -359,13 +359,24 @@ resolve_secondary_pattern <- function(x) {
 
   while (grepl("<<.*?>>", x)) {
 
-    m <- gregexpr("<<[^<]*?>>", x, perl = TRUE)
+    # stringr::str_extract_all(x, regexp)[1]
+    # (?<!<) means not preceded by <
+    # (?!>) means not followed by >
+    # Which means we take the outer match
+    # safeguarding about potential html elements within
+    # rstudio/gt#1880
+    m <- gregexpr("(?<!<)<<.*>>(?!>)", x, perl = TRUE)
 
     matched <- unlist(regmatches(x, m))[1]
 
     m_start <- as.integer(m[[1]])
     m_length <- attr(m[[1]], "match.length")
 
+    if (m_start == -1) {
+      # Add safeguard instead of going in a very long loop
+      # rstudio/gt#1880
+      cli::cli_abort("Can't resolve pattern.", .internal = TRUE)
+    }
     if (grepl(missing_val_token, matched)) {
 
       # Remove `matched` text from `x`
--- a/R/utils.R
+++ b/R/utils.R
@@ -1162,42 +1162,40 @@ markdown_to_xml <- function(text) {
 
         if (inherits(x, "xml_nodeset")) {
 
-           results <- lapply(x, apply_rules)
+          results <- lapply(x, apply_rules)
           do.call("paste0", c(results, collapse = "\n"))
 
-         } else {
+        } else {
 
           output <- if (xml2::xml_type(x) == "element") {
+          rule <- cmark_rules_xml[[xml2::xml_name(x)]]
 
-            rule <- cmark_rules_xml[[xml2::xml_name(x)]]
+          if (is.null(rule)) {
 
-            if (is.null(rule)) {
+            cli::cli_warn(
+              "Unknown commonmark element encountered: {xml2::xml_name(x)}",
+              .frequency = "once",
+              .frequency_id = "gt_commonmark_unknown_element"
+            )
 
-              rlang::warn(
-                paste0("Unknown commonmark element encountered: ", xml2::xml_name(x)),
-                .frequency = "once",
-                .frequency_id = "gt_commonmark_unknown_element"
-              )
+            apply_rules(xml2::xml_children(x))
 
-              apply_rules(xml2::xml_children(x))
+          } else if (is.function(rule)) {
 
-            } else if (is.function(rule)) {
+            rule(x, apply_rules, ...)
 
-              rule(x, apply_rules, ...)
-
-            }
           }
+        }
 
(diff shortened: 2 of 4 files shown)


rich-iannone:
Oh yes, I discovered problems with `pattern` and nesting like this quite a while back but didn't get to writing up an issue about it. Thank you for taking this one on (`while` loops are dangerous).

olivroy:
Yeah.. I have attempted to fix this before, but I never got it. I am out of idea...

Also, how is `pattern = "<<{1}<<{2}<<{3}>>>>>>"` different from `"<<{1}>><<{2}>><<{3}>>"`. All other cases seem to render fine, but the regex I am using here attemps to exclude too many consecutive `<<` 

If they are the same, maybe we could attempt to normalize <<{1}<<{2}<<{3}>>>>>> to <<{1}>><<{2}>><<{3}>> with a deprecation warning?
 (that seems tedious and not straightforward)

Another idea I had was to change `<` in pattern to `"::opening_secondary::"` and `>` to `"::closing_secondary::"` but I didn't try this..

rich-iannone:
> Also, how is pattern = "<<{1}<<{2}<<{3}>>>>>>" different from "<<{1}>><<{2}>><<{3}>>"

I believe the idea was that if {2} is missing, both {2} and {3} would be removed. Some examples:

`x = 1, y = 2, z = 3`

`"<<{x}<<{y(NA)}<<{z}>>>>>>"` -> `"1"`
`"<<{x}<< [{y}]<<({z(NA)})>>>>>>"` -> `"1 [2]"` 
`"<<{x(NA)}<<{y}<<{z}>>>>>>"` -> `""`

A sort of weird idea to make this work is to substitute an uncommon literal character in a run of `">>>>>>"` at every second position (making `">>|>>|>>"`) and then removing that literal character (probably shouldn't be a `|` though) as the final step.

R/utils_render_common.R
```diff
@@ -359,13 +359,24 @@ resolve_secondary_pattern <- function(x) {
...
     matched <- unlist(regmatches(x, m))[1]
 
     m_start <- as.integer(m[[1]])
     m_length <- attr(m[[1]], "match.length")
 
+    if (m_start == -1) {
+      # Add safeguard instead of going in a very long loop
+      # rstudio/gt#1880
+      cli::cli_abort("Can't resolve pattern.", .internal = TRUE)
+    }
```
olivroy, line 379: This is probably the only part I am satisfied with in this PR. It prevents infinite loops from happening. The rest doesn't work as intended.

Issue #1408, opened 2023-08-14
Ability to add multiple stubhead labels when row_group.as_column = TRUE


bzkrouse:
## Prework

- [x] Read and abide by **gt**'s [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/rstudio/gt/issues) (both open and closed).

## Proposal

For cases where the stub spans multiple columns (i.e. when `tab_options(row_group.as_column=TRUE)` is used), it would be great if there was an option to also provide multiple stub headers. Currently, the user can only pass a single stub header as such:

```
gtcars |>
  dplyr::slice(1:10) |>
  gt(groupname_col = "mfr",
     rowname_col = "model") |>
  tab_options(row_group.as_column = TRUE) |>
  tab_stubhead(label = "My label")  
```

olivroy:
Some code in `opt_interactive()` will need to be tweaked accordingly (will be straightforward). Relevant PR to adding `tab_stubhead()` support to interactive is #1758.
The acceptable length should also be checked. Either 1 or has_rownames + n_group_cols for this to work properly.

Issue #1406, opened 2023-08-14
Multiple rowname cols entry


thebioengineer:
## Prework

- [x] Read and abide by **gt**'s [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
- [x] Search for duplicates among the [existing issues](https://github.com/rstudio/gt/issues) (both open and closed).

## Proposal

Describe the new feature clearly and concisely. If applicable, write a minimal example in R code or pseudo-code to show input, usage, and desired output.

Currently, gt only allows for a single rowname column to be passed. This creates some difficulties if there is a table where you want to construct a table where there may be a few grouping columns you would like to use as rownames. This can be circumvented a little right now by mutating columns together into the final form you would like, but isnt simple and then requires styling downstream.

What I would like to see/be able to do is pass multiple columns to rowname_col, and make them all part of stub. 

```r
gt::gtcars |>
  gt(
    rowname_col = c("mfr","model","year")
  )
```

The order of the columns listed tells you the order of the columns in the stub, and there could be an option to remove duplicates or not:

```r
gt::gtcars |>
  gt(
    rowname_col = c("mfr","model","year")
  ) |>
 tab_option(rowname_col.remove_duplicates = FALSE) # default to true to remove duplicates
```

I can try to create "screen captures" to represent this if you need more details.

Issue #918, opened 2022-03-29
Spanner row


SteveBrouwer:
Hi, 

I'm really enjoying GT Please see my table attached, and my code below. 

I would like to merge rows in the first column. eg both bigeye rows in first column to be a single cell same for yellowfin and skipjack. This seems to be unnecessarily hard for something relatively common, so a spanner column is available but not row?  I do not want to do this as a row group label. 

Two other minor questions
1) How do I continue my line to the right of column 1 into the column label row?
2) in my subscripting function (fn) line 25 any idea why my "None" is coming out as a NA? 

Lastly in Hmisc there is a longtable function that allows a table to carry on over multiple pages, is there something similar in GT without doing it manually?

![tab_rfmo_man_stds](https://user-images.githubusercontent.com/77309577/160507497-19fc04ed-6626-4d44-96e0-4d69b514dd9d.png)

########################################################################
msy<-data.frame(Stock=c("Bigeye", "Bigeye","Skipjack","Skipjack", "Yellowfin", "Yellowfin"), 
                Reference_pt=c(rep(c("TRP/Objective", "LRP"),3)),
                WCPFC=c("1.65*SBMSY", "SBMSY", "2.14*SBMSY", "SBMSY","2.85*SBMSY", "SBMSY"),
                IOTC=c("SBMSY", "0.5*SBMSY", "1.8*SBMSY", "0.9*SBMSY","SBMSY", "0.4*SBMSY"),
                ICCAT=c("SBMSY", "", "SBMSY", "","SBMSY", ""),
                IATCC=c("SBMSY", "0.5*SBMSY", "SBMSY", "0.5*SBMSY","SBMSY", "0.5*SBMSY"))

msy_tab_1<-msy%>% gt() %>%
  tab_header(
    title = md("Tuna RFMO management standards")
  ) %>%
  tab_source_note(md("ISSF. 2021. Status of the world fisheries for tuna. Sept. 2021. ISSF 
                     Technical Report 2021-13. International Seafood Sustainability Foundation, 
                     Washington, D.C., USA.")) 

# to get subscripting
library(stringr)

msy_tab_1<-msy_tab_1%>% 
  text_transform(
    locations = cells_body(
      columns = c(WCPFC,IOTC, ICCAT, IATCC)
    ),
    fn = function(x){
      msya <- str_extract(x, "MSY")
      sb <- str_extract(x, "SB")
      dec<-str_extract(x, "\\d+\\.*\\d")
[...]

Issue #601, opened 2020-06-04
Collapse rows


ratnanil:
A feature that I'm missing in this great package is the ability to "span cells" between rows. This would be a feature similar as `gt` currently behaves when passing it a grouped dataframe. However, instead of adding the grouped column as an additional row, it would be cool if there was a way to merge cells with the same value, similar to the function `kableExtra::collapse_rows` (see way below)

Current behaviour
``` r
library(gt)
library(dplyr)

iris %>%
  group_by(Species) %>%
  sample_n(3) %>%
  gt()
```

![image](https://user-images.githubusercontent.com/12532091/83737714-ac36f280-a653-11ea-98da-9012ad850a56.png)

Desired output as demonstrated with `kableExtra`:

```
library(kableExtra)

iris %>%
  group_by(Species) %>%
  sample_n(3) %>%
  kable() %>%
  collapse_rows(5)
```

![image](https://user-images.githubusercontent.com/12532091/83736976-af7dae80-a652-11ea-8ba6-e95e76e22939.png)

PS: Issue https://github.com/rstudio/gt/issues/108#issuecomment-481441416 contains a link to this very feature from kableExtra, but that might be a mistake

Nova-Scotia:
This is exciting! I'm using the dev version of `gt` and I can get this to work using the example in #2008

```
 library(gt)

df = dplyr::tibble(
    vals = 1:8,
    rowname_left  = c("A",  "A",  "A",  "A",  "B",  "B",  "B",  "B"),
    rowname_mid   = c("AA", "AA", "AZ", "AZ", "BA", "BA", "BZ", "BZ"),
    rowname_right = c("H", "I", "J", "K", "L", "M", "N", "O")
)

gt(
    df,
    rowname_col = c("rowname_left", "rowname_mid", "rowname_right")
) |> tab_stubhead(c("left", "middle", "right"))
```

<img width="230" height="405" alt="Image" src="https://github.com/user-attachments/assets/f2444c8d-7765-4005-b6fd-5d1d376c8e09" />

However, if I just wish to group by one column, it doesn't work since the hierarchical stub layout only happens with multiple values in `rowname_col`. So when I try

```
gt(
    df,
    rowname_col = "rowname_left") |> 
  tab_stubhead("left")
```

I don't get my observations grouped by 'left':

<img width="350" height="400" alt="Image" src="https://github.com/user-attachments/assets/6239e7f7-c110-48aa-8b58-eafd852c4164" />

[...]

PR #2008, opened 2025-07-07
feat: allow arbitrary number of columns in stub

rich-iannone:
This PR adds functionality for having several columns become part of the stub. It will allow for stubs where there are levels of headers for row labels (much in the same way that spanners can be layered over column labels). Here's an example where three columns are part of the stub:

image

Since this feature breaks many assumptions in the table-building logic (i.e., a stub only has one column, but two if `gt(.., row_group_as_column = TRUE)`) there may some instability during the time when this is merged and when the package is updated in CRAN. Things that need to be done (and checked closely) are:

- [x] `tab_stubhead()` works properly with a multi-column stub; documentation should be updated
- [ ] `tab_style()` and `tab_footnote()` works with multi-column stubs
- [x] `cells_stub()` is updated with a `columns=` parameter and works with the above functions
- [ ] `tab_row_group()` works with multi-column stubs
- [x] all existing functionality and tests work when there is no stub or a one-column stub (along with row groups)
- [ ] tables with multi-column stubs render well with all output formats
- [ ] formatting works okay with columns in a multi-column stub
- [ ] summary rows not adversely affected by a multi-column stub

The above considerations may not all be covered by this PR, but addressed in follow-up PRs. And there may be other bugs discovered during that phase.

Fixes: https://github.com/rstudio/gt/issues/601
Fixes: https://github.com/rstudio/gt/issues/918
Fixes: https://github.com/rstudio/gt/issues/1406
Fixes: https://github.com/rstudio/gt/issues/1408


olivroy:
Hi Rich! this looks promising! Looking forward to helping test the new feature!

Just an FYi, as you said this may be a risky move, this PR only has a [43% test coverage](https://app.codecov.io/gh/rstudio/gt/pull/2008). If possible, I think adding a few tests would be worthwhile!

rich-iannone:
@olivroy this is definitely a big feature and I’ve been taking my time with it since so many things can be broken with this (fixed many small issues so far!). 

So far I’ve done a lot of manual/visual testing on my own computer. But yeah, there will be some tests submitted soon! 

Thanks for offering to test this all out! I imagine things will still be rough after this is merged and several follow-up fixes will need to be made before release.

olivroy:
this is great! sorry, accidentally hit the close PR button

rich-iannone:
Styling works! But it's not perfect. You can style components of the multicolumn stub:


```r
library(gt)

df = dplyr::tibble(
    vals = 1:8,
    rowname_left  = c("A",  "A",  "A",  "A",  "B",  "B",  "B",  "B"),
    rowname_mid   = c("AA", "AA", "AZ", "AZ", "BA", "BA", "BZ", "BZ"),
    rowname_right = c("H", "I", "J", "K", "L", "M", "N", "O")
)

gt(
    df,
    rowname_col = c("rowname_left", "rowname_mid", "rowname_right")
) |> 
  tab_stubhead(c("left", "middle", "right")) |>
  tab_style(
    style = list(cell_fill(color="steelblue"), cell_text(color="white")),
    locations = cells_stub(columns = "rowname_mid", rows = c(1, 5))
  )
```

and get this:

image

However, the row indices have to currently match the 'group leaders' (odd numbered rows in the `"rowname_mid"` column). So, using `rows = 2` *won't* apply the style to the 'AA' cell.

Issue #2041, opened 2025-09-29
`latex unicode conversion` leads compiling error for CJK document


leovan:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

`latex unicode conversion` leads compiling error for CJK document. Only gt 1.1.0 has this issue, gt 1.0.0 does not.

## Reproducible example

````
---
title: "quote test"
date: today
lang: zh
engine: knitr
format:
    pdf:
        pdf-engine: xelatex
        keep-tex: true
---

```{r}
library(gt)

df <- data.frame(
    'en' = c('\'test\'', '"test"'),
    'zh' = c('“测试”', '‘测试’')
)

gt(df)
```
````

For gt 1.1.0, it generates the follow latex code:

```
\begin{table}
\fontsize{12.0pt}{14.0pt}\selectfont
\begin{tabular*}{\linewidth}{@{\extracolsep{\fill}}ll}
\toprule
en & zh \\ 
\midrule\addlinespace[2.5pt]
'test' & \textquotedblleft测试\textquotedblright \\ 
"test" & \textquoteleft测试\textquoteright \\ 
\bottomrule
\end{tabular*}
\end{table}
```

There is no spacing between latex code and text characters in `\textquotedblleft测试\textquotedblright`, which make this an `Undefined control sequence`. I think this issue may be related with the follwing code: https://github.com/rstudio/gt/blob/77c3bb9945e37da877c6ee3c76f2edde02379a4e/R/helpers.R#L3130-L3147.

## Expected result

Should be compiled successfully.

## Session info

```
R version 4.5.1 (2025-06-13)
Platform: aarch64-apple-darwin20
Running under: macOS Tahoe 26.0

Matrix products: default
BLAS:   /System/Library/Frameworks/Accelerate.framework/Versions/A/Frameworks/vecLib.framework/Versions/A/libBLAS.dylib 
LAPACK: /Library/Frameworks/R.framework/Versions/4.5-arm64/Resources/lib/libRlapack.dylib;  LAPACK version 3.12.1

locale:
[...]

Issue #2037, opened 2025-09-24
Problem with UTF-8 characters, PDF output, v1.1 regression


cswingle:
## Prework

* [x] Read and agree to the [code of conduct](https://www.contributor-covenant.org/version/2/1/code_of_conduct.html) and [contributing guidelines](https://github.com/rstudio/gt/blob/master/.github/CONTRIBUTING.md).
* [x] If there is [already a relevant issue](https://github.com/rstudio/gt/issues), whether open or closed, comment on the existing thread instead of posting a new issue.

## Description

Using gt within a Quarto document, PDF output, `pdf-engine: xelatex`, I get errors and/or incorrect table formatting with gt 1.1 when the cell content includes accented characters. The new version is changing characters like `’` to `\textquoteright` and isn't adding a space or `{}`, resulting in errors with names. And `é` is converted to `\\'e`, which TeX interprets as a table linefeed instead of an accented e. Part of the reason for using `pdf-engine: xelatex` is to avoid any utf-8 translations.

This is a regression from v1.0, which does not have any of these errors with the same document.

## Reproducible example

* [x] Post a [minimal reproducible example](https://www.tidyverse.org/help/) so the maintainer can troubleshoot the problems you identify. A reproducible example is:
    * [x] **Runnable**: post enough R code and data so any onlooker can create the error on their own computer.
    * [x] **Minimal**: reduce runtime wherever possible and remove complicated details that are irrelevant to the issue at hand.
    * [x] **Readable**: format your code according to the [tidyverse style guide](https://style.tidyverse.org/).

````
---
title: "GT Test"
engine: knitr
execute:
  message: false
  warning: false
  echo: true
  eval: true
format:
  pdf:
    pdf-engine: xelatex
    keep-tex: true
---

```{r}
library(tibble)
library(gt)

sessionInfo()

tribble(
  ~name, ~finish_time,
  "Chloé Laplantine", "4:37:12",
  "Trent O’Shell", "5:15:46",
) |>
  gt()
```
````

## Expected result

I expect the table to render, with the original UTF-8 characters. For example, here's what the TeX looks like for the table when using gt 1.0:

````
\begin{table}
\fontsize{12.0pt}{14.4pt}\selectfont
\begin{tabular*}{\linewidth}{@{\extracolsep{\fill}}lr}
\toprule
name & finish\_time \\ 
\midrule\addlinespace[2.5pt]
Chloé Laplantine & 4:37:12 \\ 
[...]

PR #2042, opened 2025-09-29
Update Latex Unicode conversion

thebioengineer:
# Summary

Thank you for contributing to **gt**! To make this process easier for everyone, please explain the context and purpose of your contribution. Also, list the changes made to the existing code or documentation.

- Ref: #2037 & #2041

# Checklist


rich-iannone:
@thebioengineer before this gets reviewed, could you make sure that the failing tests pass?

snhansen:
```
* Update latex Unicode conversions to be optional (default to TRUE) as well as make them work in-line (@thebioengineer, #2037, #2041)
```

```
\item{unicode_conversion}{\emph{Boolean}

\verb{scalar<boolean>} // \emph{default:} \code{TRUE}

A boolean value indicating whether unicode in text should be turned into
LaTeX.}
```

@thebioengineer: From what I can tell, the default is set to FALSE but that's not what the documentation above says. Am I misunderstanding something or is there an error here?

thebioengineer:
> ```
> * Update latex Unicode conversions to be optional (default to TRUE) as well as make them work in-line (@thebioengineer, #2037, #2041)
> ```
> 
> ```
> \item{unicode_conversion}{\emph{Boolean}
> 
> \verb{scalar<boolean>} // \emph{default:} \code{TRUE}
> 
> A boolean value indicating whether unicode in text should be turned into
> LaTeX.}
> ```
> 
> @thebioengineer: From what I can tell, the default is set to FALSE but that's not what the documentation above says. Am I misunderstanding something or is there an error here?

The default is set in the dt_options.R file, which is TRUE by default right now. I am guessing you looked at the `escape_latex()` function, where I do have `getOption("gt.latex.unicode_convert", default = FALSE)`, but thats just for the `escape_latex()` call. 

I want to discuss with @rich-iannone more about this. I don't like this approach, but in order pass an argument down as opposed to setting an option in the environment so many functions would need to be altered to allow for this single argument, I felt like that was a little unreasonable. But if you want to be more explicit with argument passing down, we could potentially do that?

rich-iannone:
@thebioengineer if you like, we can discuss a bit more through a Zoom call? Just get in touch and we can arrange that! Using withr seems okay for now but keep in mind that's in `Suggests` (and we can't move it into `Imports`) so users will have to know to have the package available you use this Unicode conversion. I think with refactoring done down the line (once the feature is stabilized) there can be the opportunity to use the option setting directly.

I'd recommend setting the default to `FALSE` for a few reasons:

- this feature is a bit experimental: we haven't tested it on a range of LaTeX engines 
- users are running into problems with existing code (having the functionality 'off' by default would allow existing code to keep working)
- opting in is generally better for functionality that only a small cross-section of users would need
- opting in is probably easier than opting out: if the functionality is causing errors users will then have to discover that there is an escape hatch (they might not know that Unicode conversion is causing the problem); on the flip side, users that need that piece of Unicode conversion functionality are more likely to discover it through the docs

thebioengineer:
Agree with those reasonings and arguments. I'll update the code/documentation to reflect it being FALSE by default and then put an if/else in front with a message about needing withr if its true.

rich-iannone (approved):
LGTM, thanks!

PR #2077, opened 2025-11-12
update gh-actions (no need for dev covr)

olivroy:
# Summary

Just a small CI update that avoids dev covr as covr 3.6.5 is released.


.github/workflows/R-CMD-check.yaml
```diff
@@ -8,7 +8,6 @@ on:
   push:
     branches: [main, master]
   pull_request:
-    branches: [main, master]
```
rich-iannone: This PR looks great but are the empty lines after `pull_request:` intended?

olivroy: yeah, it's the new actions. It just enables actions to run on all pull requests (not just the ones to main and master https://github.com/r-lib/actions/commit/3f55ea1e7439a0db933b2d84686c94fbb4e08602

rich-iannone (approved):
LGTM, thanks!

